PR #17, opened 2018-07-06
CDATA parsing implemented, New Parsing Test Format

NikolasKomonen:
Parser now considers CDATA while parsing, Node class modified to handle it as well. Also new Parsing test format that is easier to debug and cleaner.

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/XMLHighlighting.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/XMLHighlighting.java
@@ -18,6 +18,7 @@ import org.eclipse.lsp4j.DocumentHighlight;
 import org.eclipse.lsp4j.DocumentHighlightKind;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.TextDocumentItem;
 import org.eclipse.lsp4xml.internal.parser.BadLocationException;
 import org.eclipse.lsp4xml.internal.parser.Scanner;
 import org.eclipse.lsp4xml.internal.parser.TokenType;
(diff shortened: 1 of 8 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/XMLHighlighting.java
```diff
@@ -73,20 +111,27 @@ private static boolean covers(Range range, Position position) {
...
+
 		TokenType token = scanner.scan();
 		while (token != TokenType.EOS && token != tokenType) {
 			token = scanner.scan();
 		}
 		if (token != TokenType.EOS) {
 			try {
-				return new Range(xmlDocument.positionAt(scanner.getTokenOffset()),
-						xmlDocument.positionAt(scanner.getTokenEnd()));
+				return new Range(fmDocument.positionAt(scanner.getTokenOffset()),
```
angelozerr: why fmDocument? It should be xmlDocument

NikolasKomonen: you're right, I don't know why I even did that.

fbricon: @NikolasKomonen you sjould make sure unit tests catch that

.gitignore
```diff
@@ -25,6 +25,7 @@ target/
 *~
 
 .gradle
+.vscode/
```
fbricon: don't ignore .vscode

NikolasKomonen:
@angelozerr would you be okay with me merging this, or do you have any comments?

angelozerr:
Thanks @NikolasKomonen !

PR #22, opened 2018-07-20
Implemented Logger

NikolasKomonen:


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/XMLLogger.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.logging.FileHandler;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * XMLLogger
+ */
+public class XMLLogger {
```
angelozerr: Why don't use existing Java logger like http://www.vogella.com/tutorials/Logging/article.html or sl4j like suggested in https://github.com/angelozerr/lsp4xml/issues/19 ? Please disscuss about that in https://github.com/angelozerr/lsp4xml/issues/19

NikolasKomonen: I did use the Java logger, but I did it in a different format. I'll follow how they use it in the link you sent.

PR #42, opened 2018-08-07
Scanner considers single quote of attribute value

NikolasKomonen:
eg: `<tag key='value'></tag>`


angelozerr:
Thanks @NikolasKomonen !

PR #52, opened 2018-08-10
LogHelper considers empty initObject

NikolasKomonen:


NikolasKomonen:
@angelozerr I'm not sure what this Travis error is, could you check it.

angelozerr:
Thanks @NikolasKomonen !  The problem is about reposiroty not accesible. We will see that if tomorow we have not the problem?

PR #64, opened 2018-08-17
Minor fixes to CDATA parsing

NikolasKomonen:


NikolasKomonen:
need to fix the git log

PR #65, opened 2018-08-17
Fixes the incorrect tag/content values for cdata

NikolasKomonen:


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/internal/parser/XMLParser.java
```diff
@@ -134,16 +134,17 @@ public XMLDocument parse(TextDocument document, boolean full) {
 				Node cdataNode = new Node(scanner.getTokenOffset(), text.length(), new ArrayList<>(), curr,
 						xmlDocument);// TODO: might need arraylist
 				cdataNode.isCDATA = true;
+				cdataNode.tag = "CDATA";
```
angelozerr: Why do you need that? For me tag was an element, no?

NikolasKomonen: @angelozerr I figured that node.tag would be the beginning text of a tag. And I think it makes more sense for the content inside would go in to node.content.

angelozerr: OK I understand your idea. Thanks for the clarification.

NikolasKomonen: @angelozerr can this be merged?

angelozerr: Just one question. Do you need that only for formatting? If it that I think it should be better to fill CDATA according full flag like https://github.com/angelozerr/lsp4xml/blob/master/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/internal/parser/XMLParser.java#L192 Indeed for instance for Content, we need just for format, so for outline (documents symbols) we don't need it, and other features like completions, etc. So to avoid having this information in memory which is used only for format, I'm using full flag. So please use full if it's only for format.

NikolasKomonen: @angelozerr You're right, I fixed CDATA content will consider if full is true.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/internal/parser/XMLParser.java
```diff
@@ -134,23 +134,26 @@ public XMLDocument parse(TextDocument document, boolean full) {
...
 				curr.children.add(cdataNode);
 				curr = cdataNode;
 				break;
 			}
 
 			case CDATAContent: {
-				if (curr.tag == null) {
-					curr.tag = "";
+				if (curr.content == null) {
```
angelozerr: check full should be done too here, no?

angelozerr:
@NikolasKomonen please update your PR, the full flag is replaced with a bit mask.

NikolasKomonen:
@angelozerr Updated, tests pass except the validation ones

angelozerr:
Thank's @NikolasKomonen !

PR #66, opened 2018-08-17
formatter considers prolog and PI

NikolasKomonen:


angelozerr:
Thanks @NikolasKomonen !

PR #75, opened 2018-08-24
Added lsp4j version property

NikolasKomonen:
Updated pom with property variable and new lsp4j version


angelozerr:
Thanks guys!

PR #83, opened 2018-08-31
Exit immediately when the exit command is issued

fbricon:
...as previous implementation always left the process running for exactly 
1 min after the exit() command was issued by the client.


angelozerr:
I close this PR since it was merged. I don't understand why githib tell me that there are some conflicts.

PR #92, opened 2018-09-04
General Syntax Error Ranges

NikolasKomonen:


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/contentmodel/participants/diagnostics/LSPErrorReporter.java
```diff
@@ -80,12 +82,19 @@ public String reportError(XMLLocator location, String domain, String key, Object
...
 			message = str.toString();
+			
 		}
-
+		
 		// Fill diagnostic
 		diagnostics.add(new Diagnostic(toLSPRange(location, key, arguments, document), message,
 				toLSPSeverity(severity), XML_DIAGNOSTIC_SOURCE, key));
 
+		XMLSyntaxErrorCode code = XMLSyntaxErrorCode.get(key);
```
angelozerr: Please remove this code, you have done that because you cannot have list of ranges, is that? If it that please remove it and disable test. I will do a change to support list of ranges.

angelozerr:
I think it's vbecause of https://github.com/angelozerr/lsp4xml/pull/92#discussion-diff-215595675R92

Please remove this code, IMHO, several ranges for disagnostic should be supported in later, @fbricon what do you think about that?

NikolasKomonen:
@fbricon @angelozerr updated

and in the case of :

```
<foo>
  bar>
  </bar>
</foo>
```

`</bar>` will be underlined in this version of syntax ranges

fbricon:
> and in the case of :
```xml
<foo>
  bar>
  </bar>
</foo>
```
> `</bar>` will be underlined in this version of syntax ranges

This is correct in all cases, since bar> is a valid node content

fbricon:
@angelozerr ok to merge? This is a really nice improvement

angelozerr:
Great job @NikolasKomonen !!!

PR #113, opened 2018-09-10
Travis Dependency Caching

NikolasKomonen:
I'm not sure if this works but this should speed up the builds.

The line 'sudo: false' might be needed before 'cache:' if permissions cause issues


fbricon:
Let's try it :-) It should work though

PR #127, opened 2018-09-14
Fix for validation getting a null when checking a Text Node's tag name

NikolasKomonen:


angelozerr:
@NikolasKomonen thanks for your PR. I would like just to say that I plan to remove the tag field. We should have getNodeName() for Node and Element#getTagName().

NikolasKomonen:
@angelozerr I agree, we should eventually use getters like everything else

PR #129, opened 2018-09-17
Formatted text content has correct newline character placements

NikolasKomonen:


angelozerr:
@NikolasKomonen thanks for your work! I accept your PR because I'm doing a refactoring to remove content filed from Node. My idea is to have a getData (in CharacterData like org.w3c.domCharacterData) to clean XMLDocument, but my other idea is not to store the String content in the node (to avoid doing substring if we not need to read the data content)

PR #136, opened 2018-09-21
Fixes wrong diagnostic range for missing attribute quotations

### LINKED ISSUE #135: Wrong diagnostic range for malformed attribute
Given 
```xml
<foo>
  <bar  one= two="" />
</foo>
```
the error range for `Open quote is expected for attribute "one" associated with an  element type  "bar".`  is underlying `two`, instead of `one`

NikolasKomonen:
Fixes #135

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/contentmodel/participants/XMLSyntaxErrorCode.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/contentmodel/participants/XMLSyntaxErrorCode.java
@@ -146,10 +146,9 @@ public enum XMLSyntaxErrorCode implements IXMLErrorCode {
 		case MarkupEntityMismatch:
 			return XMLPositionUtility.createRange(offset, offset + 1, document);
 		case NameRequiredInReference:
-			// Good as is
-		case OpenQuoteExpected:
-			// Working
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/contentmodel/participants/XMLSyntaxErrorCode.java
```diff
@@ -148,8 +148,7 @@ public static Range toLSPRange(XMLLocator location, XMLSyntaxErrorCode code, Obj
 		case NameRequiredInReference:
```
fbricon: should break here I think

NikolasKomonen:
@fbricon Fixed

angelozerr:
@NikolasKomonen it should be cool if you could add a code action to add those missing quotes lFor that, see existing code action at https://github.com/angelozerr/lsp4xml/tree/master/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/contentmodel/participants/codeactions

fbricon:
@angelozerr please open an enhancement request

angelozerr:
> @angelozerr please open an enhancement request

Done https://github.com/angelozerr/lsp4xml/issues/137

PR #139, opened 2018-09-24
Display server version on startup

fbricon:
helps check which version of the server is running. This is done by enabling 
Maven's resource filtering and lowering the logging threshold to INFO.


angelozerr:
Thanks @fbricon !

fbricon:
We should actually add the git commit id in the version. I'll see if I can do that when I find some time

PR #161, opened 2018-10-11
splitAttributes keeps first attribute on same line

NikolasKomonen:
Fixes #redhat-developer/vscode-xml#59


alessandrodimarco:
@NikolasKomonen thanks for fixing this issue so quickly! I just have one question about the PR. I noticed in the test `testSplitAttributes` that the first `attribute` is on the same line as the `element`. In the issue I submitted https://github.com/redhat-developer/vscode-xml/issues/59 I was only concerned about elements with one attribute not being moved. Would it be possible to keep the existing behavior when there is multiple attributes on a element with the first attribute starting a new line? Thanks!

NikolasKomonen:
@alessandrodimarco sure, I just want sure what people preferred so this will be fine.

angelozerr:
@NikolasKomonen can I merge it?

NikolasKomonen:
@angelozerr yup, ill do it.

PR #170, opened 2018-10-19
Formatting for splitAttributes indentation fixed

### LINKED ISSUE #169: Add preference to align attributes to tag vs simply indent them
Following up on #100, currently when attributes are formatted 1 per line, they're aligned on the tag. Which can be fine in some case, but not to the taste of everybody. There should be a preference that allows to simply indent them instead (default Eclipse behaviour).
So for instance, with `"xml.format.alignAttributesWithTag": true` :
```xml
<xsl:stylesheet
               xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
               xmlns:math="http://www.w3.org/2005/xpath-functions/math"
               xmlns:xd="http://www.oxygenxml.com/ns/doc/xsl"
               xmlns:emp="http://www.semanticalllc.com/ns/employees#"
               xmlns:h="http://www.w3.org/1999/xhtml"
               xmlns:fn="http://www.w3.org/2005/xpath-functions"
               xmlns:j="http://www.w3.org/2005/xpath-functions"
               exclude-result-prefixes="xs math xd h emp"
               version="2.0">

```

and with `"xml.format.alignAttributesWithTag": false`
```xml
<xsl:stylesheet
    xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
    xmlns:math="http://www.w3.org/2005/xpath-functions/math"
    xmlns:xd="http://www.oxygenxml.com/ns/doc/xsl"
    xmlns:emp="http://www.semanticalllc.com/ns/employees#"
    xmlns:h="http://www.w3.org/1999/xhtml"
    xmlns:fn="http://www.w3.org/2005/xpath-functions"
    xmlns:j="http://www.w3.org/2005/xpath-functions"
    exclude-result-prefixes="xs math xd h emp" version="2.0">
```

I now think such preference should be set to false.

NikolasKomonen:
Fixes #169

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/XMLBuilder.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/XMLBuilder.java
@@ -26,6 +26,7 @@ public class XMLBuilder {
 	private final String lineDelimiter;
 	private final StringBuilder xml;
 	private final String whitespacesIndent;
+	private final int splitAttributesIndent = 2;
 
 	public XMLBuilder(XMLFormattingOptions formattingOptions, String whitespacesIndent, String lineDelimiter) {
 		this.whitespacesIndent = whitespacesIndent;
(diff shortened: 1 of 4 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/XMLBuilder.java
```diff
@@ -80,23 +80,21 @@ public XMLBuilder addSingleAttribute(String name, String value) {
...
 		return this;
 	}
 
-	public XMLBuilder addAttribute(String name, String value, int index, int level, String tagName) {
+	public XMLBuilder addAttributes(String name, String value, int index, int level, String tagName) {
 		int spaces = 1; //Accounts for '<' if isSplitAttributes() or space for single attribute 
 		if (isSplitAttributes()) {
 			linefeed();
-			indent(level);
+			indent(level + 2);
```
fbricon: extract 2 as constant

PR #171, opened 2018-10-22
Add default authenticator

snjeza:
Fixes https://github.com/redhat-developer/vscode-xml/issues/60

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLServerLauncher.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLServerLauncher.java
@@ -12,6 +12,8 @@ package org.eclipse.lsp4xml;
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.net.Authenticator;
+import java.net.PasswordAuthentication;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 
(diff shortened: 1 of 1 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLServerLauncher.java
```diff
@@ -27,6 +29,16 @@
 	 * and output streams.
 	 */
 	public static void main(String[] args) {
+		final String username = System.getProperty("http.proxyUser", "");
+		final String password = System.getProperty("http.proxyPassword", "");
+		Authenticator.setDefault(new Authenticator() {
```
angelozerr: Test configuration of Authentificator should be done only if username + password is set, no?

Why not using lambda instead of new Athenticator?

snjeza:
> Test configuration of Authentificator should be done only if username + password is set, no?

Fixed.

> Why not using lambda instead of new Athenticator?

Authenticator is the abstract class.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLServerLauncher.java
snjeza: > Test configuration of Authentificator should be done only if username + password is set, no?

Fixed.

> Why not using lambda instead of new Athenticator?

Authenticator is the abstract class.

angelozerr:
Thanks @snjeza !

PR #187, opened 2018-11-02
Change 'resource downloading' diagnostic severity to Information

fbricon:
Also added some tracing about download performance


fbricon:
Helps you realize downloading  http://www.w3.org/2001/xml.xsd takes 15s!!!!  
Which is really strange considering it *very* fast from a browser. Using curl also results in super very slow queries.

PR #196, opened 2018-11-05
Fixes indentation issue for schema based attributes

### LINKED ISSUE #188: "format.splitAttributes:true" adds excessive indentation
Having "format.splitAttributes:true", given:
```xml
<beans
        xmlns="http://www.springframework.org/schema/beans"
        xmlns:camel="http://camel.apache.org/schema/spring"
        xmlns:util="http://www.springframework.org/schema/util"
        xsi:schemaLocation="http://www.springframework.org/schema/beans 
          http://www.springframework.org/schema/beans/spring-beans-4.2.xsd  
          http://camel.apache.org/schema/spring 
          http://camel.apache.org/schema/spring/camel-spring.xsd
          http://www.springframework.org/schema/util 
          http://www.springframework.org/schema/util/spring-util-4.2.xsd">
    <camelContext xmlns="http://camel.apache.org/schema/spring">
        <!-- Add routes here -->
        <route>
            <proces|// complete here
        </route>
    </camelContext>
</beans>
```
results in weirdly formatted xml, with too many indentations:
splitattributes

NikolasKomonen:
Fixes #188

The issue was that when adding attributes it  needed to know before hand how many attributes
would be added to know which add attribute method should be called.
.isRequired() was messing this up and was moved outside to determine the total variables being added
before the add attribute methods were called.

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/utils/XMLGenerator.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/utils/XMLGenerator.java
@@ -134,23 +134,28 @@ public class XMLGenerator {
 
 	private int generate(Collection<CMAttributeDeclaration> attributes, int level, int snippetIndex, XMLBuilder xml, String tagName) {
 		int attributeIndex = 0;
-		int attributesSize = attributes.size();
-		for (CMAttributeDeclaration attributeDeclaration : attributes) {
-			if (attributeDeclaration.isRequired()) {
-				String value = attributeDeclaration.getDefaultValue();
-				if (canSupportSnippets) {
-					snippetIndex++;
-					value = ("$" + snippetIndex);
-				}
-				if(attributesSize != 1) {
-					xml.addAttributes(attributeDeclaration.getName(), value, attributeIndex, level, tagName);
-				}
-				else {
-					xml.addSingleAttribute(attributeDeclaration.getName(), value);
-				}
-				
-				attributeIndex++;
+		
+		ArrayList<CMAttributeDeclaration> requiredAttributes = new ArrayList<CMAttributeDeclaration>();
+		for (CMAttributeDeclaration att: attributes) {
+			if(att.isRequired()) {
+				requiredAttributes.add(att);
 			}
+		}
+		int attributesSize = requiredAttributes.size();
+		for (CMAttributeDeclaration attributeDeclaration : requiredAttributes) {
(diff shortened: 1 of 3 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/utils/XMLGenerator.java
```diff
@@ -134,23 +134,28 @@ public String generate(Collection<CMAttributeDeclaration> attributes, String tag
...
-				
-				attributeIndex++;
+		
+		ArrayList<CMAttributeDeclaration> requiredAttributes = new ArrayList<CMAttributeDeclaration>();
+		for (CMAttributeDeclaration att: attributes) {
+			if(att.isRequired()) {
+				requiredAttributes.add(att);
+			}
+		}
+		int attributesSize = requiredAttributes.size();
```
fbricon: and if there's no required attributes? no indentation?

NikolasKomonen: Yeah Im not sure why, but thats what angelo had. @angelozerr whats the reason for only using attribute.isRequired() attributes?

angelozerr: If I understand your question, is why we generate only required attributes? If it that it's because when you apply completion, it generates only required attributes and not the whole attributes. Not sure that I have understood the question.

NikolasKomonen: XSD example:
``` 
<xs:attribute name="lang" type="xs:string" use="required"/>
```

See `use="required"` above.

Is that what isRequired() refers to?

And if we had an element that was optional:

```
<xs:attribute name="lang" type="xs:string" use="optional"/>
```

would isRequired() == false ?

Why is isRequired() necessary to be checked to consider the attribute as a completion item, are there too many "optional" attributes?

angelozerr:
> Is that what isRequired() refers to?

Yes

> Why is isRequired() necessary to be checked to consider the attribute as a completion item, are there too many "optional" attributes?

Because when you apply completion item, you want have a valid XML element content (required attribute) and you don't want to fill optional attributes. 

WTP XML Editor works like this, if you don't like this behaviour we could consider to use a new preferences to generate or not optionnal attributes.

fbricon:
no it's fine as it is

PR #199, opened 2018-11-06
Add test for diagnostic updates when DTD changes

fbricon:
Checks for https://github.com/redhat-developer/vscode-xml/issues/79

### DIFF
--- a/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/DTDDiagnosticsTest.java
+++ b/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/DTDDiagnosticsTest.java
@@ -33,9 +33,39 @@ public class DTDDiagnosticsTest {
 				"<web-app>\r\n" + //
 				"	<XXX></XXX>\r\n" + //
 				"</web-app>";
-		testDiagnosticsFor(xml, d(6, 2, 6, 5, DTDErrorCode.MSG_ELEMENT_NOT_DECLARED),
-				d(7, 10, 7, 10, DTDErrorCode.MSG_CONTENT_INVALID));
-		// testCodeActionsFor(xml, d, ca(d, te(3, 2, 3, 15, "")));
+		testDiagnosticsFor(xml, d(6, 2, 5, DTDErrorCode.MSG_ELEMENT_NOT_DECLARED),
+				d(7, 10, 10, DTDErrorCode.MSG_CONTENT_INVALID));
+	}
+
+	@Test
+	public void testDoctypeDiagnosticsRefresh() throws Exception {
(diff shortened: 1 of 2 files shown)


fbricon:
The test fails if applied to the v0.0.1 code base, so it works

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/DTDDiagnosticsTest.java
```diff
@@ -33,9 +33,39 @@ public void MSG_ELEMENT_NOT_DECLARED() throws Exception {
...
 				"</web-app>";
-		testDiagnosticsFor(xml, d(6, 2, 6, 5, DTDErrorCode.MSG_ELEMENT_NOT_DECLARED),
-				d(7, 10, 7, 10, DTDErrorCode.MSG_CONTENT_INVALID));
-		// testCodeActionsFor(xml, d, ca(d, te(3, 2, 3, 15, "")));
+		testDiagnosticsFor(xml, d(6, 2, 5, DTDErrorCode.MSG_ELEMENT_NOT_DECLARED),
+				d(7, 10, 10, DTDErrorCode.MSG_CONTENT_INVALID));
+	}
+
+	@Test
+	public void testDoctypeDiagnosticsRefresh() throws Exception {
```
angelozerr: Fred could you please check that your test failed when you use XMLGrammarCachingConfiguration in https://github.com/angelozerr/lsp4xml/blob/master/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/diagnostics/XMLValidator.java#L49 

It your test fail, it means that your test is correct. Thanks!

fbricon: Yes it fails with XMLGrammarCachingConfiguration:
```
java.lang.AssertionError: Unexpected diagnostics:
[Diagnostic [
  range = Range [
    start = Position [
      line = 8
      character = 3
    ]
    end = Position [
      line = 8
      character = 5
    ]
  ]
  severity = Error
  code = "MSG_ELEMENT_NOT_DECLARED"
  source = "xml"
  message = "Element type "id" must be declared."
  relatedInformation = null
]] expected:<[]> but was:<[Diagnostic [
  range = Range [
    start = Position [
      line = 8
      character = 3
    ]
    end = Position [
      line = 8
      character = 5
    ]
  ]
  severity = null
  code = "MSG_ELEMENT_NOT_DECLARED"
  source = null
  message = null
  relatedInformation = null
]]>
```

angelozerr:
Great! Thanks @fbricon !

PR #200, opened 2018-11-06
DTD losing content on format fixed

### LINKED ISSUE #198: Formatting deletes document's body when there's a DTD declaration
Formatting this:
```
<?xml version="1.0"?>
<!DOCTYPE student [
  <!ELEMENT student (surname,id)>
  <!ELEMENT surname (#PCDATA)>
  <!ELEMENT id (#PCDATA)>
]>
<student>
  <surname>Smith</surname>
  <id>567896</id>
</student>
```

yields
```
<?xml version="1.0"?>
<!DOCTYPE student [
  >
```

NikolasKomonen:
Fixes #198

### DIFF
--- a/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLFormatterTest.java
+++ b/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLFormatterTest.java
@@ -376,20 +376,92 @@ public class XMLFormatterTest {
 	@Test
 	public void testDTDFormatting() throws BadLocationException {
 		String content = 
-			"<!DOCTYPE web-app PUBLIC" + lineSeparator() +
-			"\"-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN\"" + lineSeparator() +
-			"\"http://java.sun.com/dtd/web-app_2_3.dtd\" >" + lineSeparator() + 
-		   	lineSeparator() +
-		    "<web-app>" + lineSeparator() +
- 			"  <display-name>Servlet 2.3 Web Application</display-name>" + lineSeparator() +
+			"<!DOCTYPE web-app PUBLIC\r\n" + 
+			"\"-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN\"\r\n" +
+			"\"http://java.sun.com/dtd/web-app_2_3.dtd\" >\r\n" + 
+		    "<web-app>\r\n" + 
+ 			"  <display-name>Servlet 2.3 Web Application</display-name>\r\n" + 
 		    "</web-app>";
 		String expected = 
-			"<!DOCTYPE web-app PUBLIC" + lineSeparator() +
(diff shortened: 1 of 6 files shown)


org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLFormatterTest.java
```diff
@@ -386,14 +386,87 @@ public void testDTDFormatting() throws BadLocationException {
...
 			"  <display-name>Servlet 2.3 Web Application</display-name>" + lineSeparator() +
 			"</web-app>";
 		XMLFormattingOptions formattingOptions = createDefaultFormattingOptions();
 		format(content, expected, formattingOptions);
 	}
 
+	@Test
+	public void testDTDFormattingInternal() throws BadLocationException {
+		String content = 
+			"<?xml version=\"1.0\"?>" + lineSeparator() +
```
angelozerr: Please don't use lineSeparator() because your tests depends on OS. Please use \r or \ r\n.

If you use Eclipse to develop Java, you can copy a content of XML file and go to the Java Editor code. Type just "" and copy/paste the pasted XML content inside the "". Java Editor JDT will manage \r

angelozerr:
Very cool PR @NikolasKomonen ! IMHO, I think it should really cool if we could use your scanner to manage in XMLDocument, the Entity to display it in the Outline, manage completion, etc

NikolasKomonen:
> I think it should really cool if we could use your scanner to manage in XMLDocument, the Entity to display it in the Outline, manage completion, etc

@angelozerr yes! Thats what I was hoping we could eventually work on. This pr is just temporary to make sure DTD content doesn't disappear.

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLFormatterTest.java
NikolasKomonen: I though \r\n was specific to Windows, do you know why it works for me on linux?

angelozerr:
> @angelozerr yes! Thats what I was hoping we could eventually work on. This pr is just temporary to make sure DTD content doesn't disappear.

Great!

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLFormatterTest.java
angelozerr: Sorry with my bad explanation. 

> I though \r\n was specific to Windows, 

Yes it is.

I meant that we have tests which is not linked to an OS. So in your Linux case, you will havelineSparatator= \r and in my OS Windows case, I will have lineSparatator=\r\n So the offset will be different.

I agree with you, your test doesn't use offset so we will hav eno problem, but it should be more clean to have tests which hard coded the line separator.

Hope you will understand my explanation.

fbricon:
PR fixes the issues. @NikolasKomonen fix the line endings and we can merge it

NikolasKomonen:
@fbricon line endings fixed

PR #224, opened 2018-11-14
Handles setting a space before a self closing tag end bracket.

### LINKED ISSUE #197: Add new preference to insert whitespace before closing empty end-tag
... so that, if enabled: 

`<br/>` -> `<br />`
`<foo attribute="bar"/>` -> `<foo attribute="bar" />`

if disabled:

`<br />` -> `<br/>`
`<foo attribute="bar" />` -> `<foo attribute="bar"/>`

Current behaviour is equivalent to having this option enabled

NikolasKomonen:
Fixes #197

Also fixed an unknown issue where the new settings would replace existing and not
update the current ones. In the case of a setting not being sent from the client,
the default value set on the server side would be lost in XMLFormattingOptions
anymore.

### DIFF
--- a/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/settings/capabilities/XMLCapabilitiesTest.java
+++ b/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/settings/capabilities/XMLCapabilitiesTest.java
@@ -57,7 +57,7 @@ public class XMLCapabilitiesTest {
 
 	@Before
 	public void startup() {
-		XMLFormattingOptions formattingOptions = new XMLFormattingOptions();
+		XMLFormattingOptions formattingOptions = new XMLFormattingOptions(true);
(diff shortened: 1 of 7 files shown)


org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/settings/capabilities/XMLCapabilitiesTest.java
```diff
@@ -57,7 +57,7 @@
 
 	@Before
 	public void startup() {
-		XMLFormattingOptions formattingOptions = new XMLFormattingOptions();
+		XMLFormattingOptions formattingOptions = new XMLFormattingOptions(true);
```
fbricon: please add javadoc to that constructor

fbricon:
OK to merge once the missing javadoc on XMLFormattingOptions(boolean) is added

NikolasKomonen:
@fbricon updated

PR #249, opened 2018-12-04
Added DOMElement properties (more offset values). Also fixed formatting issues.

### LINKED ISSUE #227: Formatter either adds or removes content in malformed xml
In this document, a closing bracket is missing L8:
```xml
<resources something="foo">
  <resource name="res00">
    <property name="propA" value="..."/>
    <property name="propB" value="..."/>
  </resource>
  <resource ddd="ddd" name="">
    <property name="one" value="..."
    <property name="two" value="..."/>
  </resource>
</resources>
```
When invoking the formatter, a closing bracket is added:
formatter-adds-closing-bracket

Here, closing bracket is also missing L8:
```xml
<web-app 
         xmlns="http://xmlns.jcp.org/xml/ns/javaee"
         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee 
		 http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd"
         version="3.1">
         <servlet>
             <servlet-name>sssi</servlet-name>
         </servlet
</web-app>
```
But then, the formatter removes the line completely!
formatter-removing-malformed-element

NikolasKomonen:
Fixes #227

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/DOMElement.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/DOMElement.java
@@ -30,13 +30,14 @@ public class DOMElement extends DOMNode implements org.w3c.dom.Element {
 
 	String tag;
 	boolean selfClosed;
+	
+	//DomElement.start == startTagOpenOffset
+	Integer startTagOpenOffset; // |<root>
+	Integer startTagCloseOffset; // <root |>
 
-	Integer startTagOpenOffset;
-	Integer startTagCloseOffset; // <root|>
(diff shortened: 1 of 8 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/TokenType.java
```diff
@@ -74,5 +74,5 @@
 	DTDEntityKind, 
 	DTDEntityURL,
 	
-	DTDEndTag;
+	DTDEndTag, EndOfIncompleteStartTag, EndOfIncompleteEndTag;
```
fbricon: use new lines

PR #275, opened 2019-01-07
Cache unavailable URI's

### LINKED ISSUE #201: CacheResourcesManager keeps trying to download unavailable resources
When a non existing schema is used, CacheResourcesManager will keep on hitting the bad url on every keystroke.

Failed caching queries should be cached themselves (5 minutes?), so as to avoid unnecessary http requests.

NikolasKomonen:
Fixes #201

Tests aren't complete, need to find a way to handle exceptions.

### DIFF
--- a/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/uriresolver/CacheResourcesManagerTest.java
+++ b/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/uriresolver/CacheResourcesManagerTest.java
@@ -12,10 +12,39 @@ package org.eclipse.lsp4xml.uriresolver;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.fail;
 
+import java.util.concurrent.TimeUnit;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+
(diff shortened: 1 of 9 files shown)


org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/uriresolver/CacheResourcesManagerTest.java
```diff
@@ -32,4 +43,52 @@ private void testCanUseCache(boolean useCacheEnabled) {
...
+				try {
+					assertNull(cacheResourcesManager.getResource("http://bad"));
+				} catch (Exception e) {
+					//TODO: handle exception
+				}
+			}
+		}
+	}
+
+	class MockCacheResourcesManager extends CacheResourcesManager {
```
fbricon: it' unused

angelozerr:
I have just few questions about this PR:

 * why having chosen caffeine? Why not EHCache?
 * please keep in mind I would like to move lsp4xml to Eclipse.org. Having a dependency to caffeine will requires IP Process.
 * It seems that https://mvnrepository.com/artifact/com.github.ben-manes.caffeine/caffeine/2.6.2 takes 644KB, is not a problem for vscode download?

fbricon:
Caffeine is a high performance caching library. Its use has already been approved by the Eclipse Foundation, so IP is not an issue. 
But actually, for our use case, with an extremely small number of urls to cache, we probably can just stick with the default Cache from guava, that we already embed and which has the same API.
Just replace the Caffeine.newBuilder() call with CacheBuilder.newBuilder().

NikolasKomonen:
@fbricon updated with the jetty pr

PR #279, opened 2019-01-14
Adding root element 'xml' to preferences

### LINKED ISSUE #257: Use root element for XMLClientSettings
I'm about to add support for the lsp4xml in theia. In theia, when sending settings to a server, one or more roots can be selected (in my case it's `xml`). Doing so, all the `xml.*` setting will be send to the server wrapped by the root object.

```
"xml": {
     "logs":{"client":true},
     "format":{"enabled":true,"spaceBeforeEmptyCloseTag":true},
     "trace":{"server":"off"}
    }
```

The idea behind that is that the server may later rely on other client settings e.g. `http` to get the `proxy` settings, than you would just add a new `xml` sibling `http`

Would it be possible to load the setting with a `xml` prefix?
Does it make sense?

NikolasKomonen:
Fixes #257, wiki will be updated after merge.

This will require a modification to the wiki. With the addition of the 'xml' key under settings:

**Old:**
```json
  {
        "settings": {
                "trace": {
                    "server": "verbose"
                },
                "catalogs": [],
                "format": {
                    "splitAttributes": true,
                    "joinCDATALines": false,
                    "joinContentLines": false,
                    "joinCommentLines": false,
                    "spaceBeforeEmptyCloseTag": false,
                    "enabled": true
                },
               ...
          }
  }
```
**New:**
```json
  {
        "settings": {
            "xml": {
                "trace": {
                    "server": "verbose"
                },
                "catalogs": [],
                "format": {
                    "splitAttributes": true,
                    "joinCDATALines": false,
                    "joinContentLines": false,
                    "joinCommentLines": false,
                    "spaceBeforeEmptyCloseTag": false,
                    "enabled": true
                },
               ...
            }
        }
  }
```

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLLanguageServer.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLLanguageServer.java
@@ -37,9 +37,10 @@ import org.eclipse.lsp4xml.logs.LogHelper;
 import org.eclipse.lsp4xml.services.IXMLDocumentProvider;
 import org.eclipse.lsp4xml.services.XMLLanguageService;
 import org.eclipse.lsp4xml.services.extensions.CompletionSettings;
+import org.eclipse.lsp4xml.settings.AllXMLSettings;
 import org.eclipse.lsp4xml.settings.InitializationOptionsSettings;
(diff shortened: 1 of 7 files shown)


NikolasKomonen:
@angelozerr Would you be able to update lsp4e to handle this

angelozerr:
I will do it. You Can merge your work

PR #293, opened 2019-01-29
Add WorkspaceFolders support

yanxi0830:
This adds support for workspaceFolders to ServerCapabilities, which addresses https://github.com/eclipse/wildwebdeveloper/issues/124 where XML LS is started multiple times when opening different files from different projects. 

Signed-off-by: Xi Yan <xixiyan@redhat.com>

### DIFF
--- a/extensions/org.eclipse.lsp4xml.extensions.emmet/pom.xml
+++ b/extensions/org.eclipse.lsp4xml.extensions.emmet/pom.xml
@@ -3,7 +3,7 @@
 	<parent>
 		<groupId>org.lsp4xml</groupId>
 		<artifactId>lsp4xml-extensions</artifactId>
-		<version>0.3.0-SNAPSHOT</version>
+		<version>0.4.0-SNAPSHOT</version>
 	</parent>
 	<artifactId>org.eclipse.lsp4xml.extensions.emmet</artifactId>
 	<dependencies>
--- a/extensions/org.eclipse.lsp4xml.extensions.web/pom.xml
+++ b/extensions/org.eclipse.lsp4xml.extensions.web/pom.xml
@@ -3,7 +3,7 @@
 	<parent>
 		<groupId>org.lsp4xml</groupId>
 		<artifactId>lsp4xml-extensions</artifactId>
-		<version>0.3.0-SNAPSHOT</version>
+		<version>0.4.0-SNAPSHOT</version>
 	</parent>
 	<artifactId>org.eclipse.lsp4xml.extensions.web</artifactId>
 	<dependencies>
--- a/extensions/pom.xml
+++ b/extensions/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.lsp4xml</groupId>
 		<artifactId>lsp4xml</artifactId>
-		<version>0.3.0-SNAPSHOT</version>
+		<version>0.4.0-SNAPSHOT</version>
 	</parent>
 	<artifactId>lsp4xml-extensions</artifactId>
 	<packaging>pom</packaging>
--- a/org.eclipse.lsp4xml/pom.xml
+++ b/org.eclipse.lsp4xml/pom.xml
@@ -3,7 +3,7 @@
 	<parent>
 		<groupId>org.lsp4xml</groupId>
 		<artifactId>lsp4xml</artifactId>
-		<version>0.3.0-SNAPSHOT</version>
+		<version>0.4.0-SNAPSHOT</version>
 	</parent>
 	<artifactId>org.eclipse.lsp4xml</artifactId>
 	<properties>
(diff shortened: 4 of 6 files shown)


mickaelistria:
We indeed need a way for a LS to declare its multiplicity as the protocol doesn't assume a given server can handle multiple projects/files by default.
In LSP4E, we assume that the workspaceFolders capability declares ability for a given LS to cover multiple files, but if anyone can recommend something better and more reliable, we're also open to change LSP4E. It's just what seems to work best at the moment to our knowledge.

angelozerr:
For me it seems Good. Fred what do you think ?

fbricon:
In its current state, this PR doesn't really add workspaceFolders support, it merely lets the server declares it does support it (but really doesn't). The fact it seems to work in lsp4e seems like a fluke.
Only the root uri is used at the moment, Workspace folders sent to the server are ignored.
I haven't analyzed the requirements of multi-root support in the context of lsp4xml, but it's certainly more complex than that.

mickaelistria:
If you look at https://github.com/Microsoft/language-server-protocol/issues/281 and https://github.com/Microsoft/language-server-protocol/issues/289 , it seems like the multi-root is the only way for a client to know whether to spin a new LS or not according to the file URI.
I think the root issue is in the protocol that doesn't explicitly support the case of LS that aren't attached to any folder, and I'll open a bug to the LSP, but in the meantime, just declaring support for workspaceFolders seems to be a fair thing to do in LSP4XML since ignoring them is the best way to support them. Workspace Folders currently ignored => Workspace Folders are supported.

mickaelistria:
How do you deal with it in VSCode? You just start the LS and handle this singleton specifically in the extension, or is there a more "declarative" way to describe multiplicity in VSCode usage of Language Servers?

fbricon:
vscode seems to work all right with the stock lsp4xml server, in a multi-root setup, but we haven't tested it extensively, and we don't claim multi-root support anyway. Only one server instance is launched.

mickaelistria:
What should be done in XML-LS to safely claim it supports workspace folders? Since it doesn't seem to care about rootUri or workspace folders to work well?

NikolasKomonen:
The only issue I can find as of now is to do with catalog and file association paths. 

If paths from the non-main workspace folder are relative then there could be issues as they will get concatenated the the main workspace folder's uri.

I think as long as those issues are indicated everything should work.

NikolasKomonen:
@sandy081 based off of you work [here](https://github.com/Microsoft/vscode/commit/df073373e18f47d083911400e18592e888d6a808#diff-f9a5df3b8f5be28a559ff81a266c2bfb), I hope you might know the answer to the following:

The `resource` scope for a configuration property says it will handle:
    
    User, Workspace, and Folder settings. 

Im assuming that Folder settings will override all previous preferences in the case of multi-root workspaces, but when I've tested it with our XML LS and the Python LS it seems that folder settings are ignored. I was able to see that only Workspace overrode User settings, while Folder settings did not. I verified this during the output of the extensions' trace server which showed no change for Folder settings.

sandy081:
@NikolasKomonen Yes a resource scoped setting should override user and workspace if it is defined in folder in a MR workspace.


Can you please give me a sample code to reproduce.

NikolasKomonen:
@sandy081 

1) Open an HTML file in a VSCode Workspace.
2) Go to Preferences and set `"html.trace.server"` to verbose
3) Also in Preferences for each setting scope (User, Workspace, Folder)
set a different setting for [`"html.format.WrapAttributes"`](https://github.com/Microsoft/vscode/blob/master/extensions/html-language-features/package.json#L116)
3) View tab -> Output -> HTML Language Server from dropdown menu
4) ctrl+f and search for "wrapAttributes" and you should see the Workspace scoped preference and not the Folder one.

sandy081:
@NikolasKomonen I am looking for sample code that I can try and repro. It is easy to figure out the issue and fix it.

NikolasKomonen:
@sandy081 Here is a sample of HTML code.

```html
<a att="value1" att2="value2" att3="value3">

</a>
```

sandy081:
@NikolasKomonen Sorry if I am not clear. Can you please write a sample extension that uses the configuration API and reproduces the bug you are seeing? I would like to use that sample extension to reproduce and investigate.

NikolasKomonen:
@sandy081 Heres the repo to a stripped down version of the yaml server: https://github.com/NikolasKomonen/vscode-yaml/tree/yamlLite

> uses the configuration API

If [this](https://github.com/NikolasKomonen/vscode-yaml/blob/yamlLite/src/extension.ts#L39) was not what you mean let meant, let me know.

sandy081:
@NikolasKomonen Sorry it is not helpful. 

- Create an extension using yo code generator
- Write the code that will demonstrate reproducing the issue you are seeing
- Provide me that code.

Reason I am asking is that, it is hard for me to debug your real extension.

fbricon:
@sandy081 the bug can be reproduced with VSCode's HTML extension, it's literally part of vscode

sandy081:
@fbricon Ah ok. Can you please file an issue in VS Code with necessary details?

mickaelistria:
What's the status of this issue? I don't really get how last comments are related.

mickaelistria:
This was added by https://github.com/eclipse/lemminx/commit/97ea5070d99c9a7e4017862333ff692878fc98c6 . This PR is not necessary any more. Thanks!

PR #298, opened 2019-02-12
Potential leak of handles and memory

snjeza:
Fixes https://github.com/redhat-developer/vscode-xml/issues/108
Related PR - https://github.com/eclipse/eclipse.jdt.ls/pull/931

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/ParentProcessWatcher.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/ParentProcessWatcher.java
@@ -83,14 +83,34 @@ public final class ParentProcessWatcher implements Runnable, Function<MessageCon
 		} else {
 			command = "ps -p " + pid;
 		}
+		Process process = null;
+		boolean finished = false;
 		try {
-			Process process = Runtime.getRuntime().exec(command);
-			int processResult = process.waitFor();
-			return processResult == 0;
+			process = Runtime.getRuntime().exec(command);
+			finished = process.waitFor(POLL_DELAY_SECS, TimeUnit.SECONDS);
+			if (!finished) {
+				process.destroy();
+				finished = process.waitFor(POLL_DELAY_SECS, TimeUnit.SECONDS); // wait for the process to stop
+			}
+			return !finished || process.exitValue() == 0;
 		} catch (IOException | InterruptedException e) {
-			//JavaLanguageServerPlugin.logException(e.getMessage(), e);
+			// JavaLanguageServerPlugin.logException(e.getMessage(), e);
 			return true;
+		} finally {
+			if (process != null) {
+				if (!finished) {
+					process.destroyForcibly();
+				}
+				// Terminating or destroying the Process doesn't close the process handle on Windows. 
+				// It is only closed when the Process object is garbage collected (in its finalize() method).
+				// On Windows, when the Java LS is idle, we need to explicitly request a GC, 
+				// to prevent an accumulation of zombie processes, as finalize() will be called.
+				if (isWindows) {
+					System.gc();
+				}
+			}
 		}
+
 	}
 
 	@Override


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/ParentProcessWatcher.java
```diff
@@ -83,14 +83,28 @@ private boolean parentProcessStillRunning() {
...
+		boolean result = false;
 		try {
-			Process process = Runtime.getRuntime().exec(command);
-			int processResult = process.waitFor();
-			return processResult == 0;
+			process = Runtime.getRuntime().exec(command);
+			result = process.waitFor(POLL_DELAY_SECS, TimeUnit.SECONDS);
+			if (!result) {
+				process.destroy();
+				result = process.waitFor(POLL_DELAY_SECS, TimeUnit.SECONDS);
```
fbricon: why do you wait twice?

fbricon: On SO, this [answer](https://stackoverflow.com/a/5483976/753170) mentions having to read the streams. Does it apply here?

fbricon:
I'll paste @snjeza's comment for the jdt.ls fix here too:

```
Java doesn't call the finalize method when terminating a process. Because of that, when Java LS isn't used, it creates a lot of zombie processes on Windows.
See https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/ParentProcessWatcher.java#L64

Steps to reproduce:

* start VS Code and Java LS, but don't use it
* download https://download.sysinternals.com/files/Handle.zip and unpack it
* call the following command several times:

```
handle64 -s -p <Java_LS_pid> | grep Process
```
You will see that the number of processes is constantly increasing.

The solution is to call System.gc(). We can also call it on Linux and Mac because it runs when Java LS isn't active.
```

fbricon:
@snjeza did you mean `destroy` instead of `finalize`?

fbricon:
Ok so IIUC, we need to call the GC explicitly so that it calls finalize() on the process, which in turn releases the handles, right?

snjeza:
> Ok so IIUC, we need to call the GC explicitly so that it calls finalize() on the process, which in turn releases the handles, right?

Right.

snjeza:
@fbricon I have updated the PR.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/ParentProcessWatcher.java
```diff
@@ -83,14 +83,32 @@ private boolean parentProcessStillRunning() {
...
 			command = "ps -p " + pid;
 		}
+		Process process = null;
+		boolean result = false;
 		try {
-			Process process = Runtime.getRuntime().exec(command);
-			int processResult = process.waitFor();
-			return processResult == 0;
+			process = Runtime.getRuntime().exec(command);
+			result = process.waitFor(POLL_DELAY_SECS, TimeUnit.SECONDS);
```
fbricon: s/result/finished/g

snjeza: changed.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/ParentProcessWatcher.java
snjeza: No, it doesn't.

snjeza: > why do you wait twice?

We wait for the process to stop.

snjeza:
> @snjeza did you mean destroy instead of finalize?

finalize.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/ParentProcessWatcher.java
```diff
@@ -83,14 +83,32 @@ private boolean parentProcessStillRunning() {
...
 		} catch (IOException | InterruptedException e) {
-			//JavaLanguageServerPlugin.logException(e.getMessage(), e);
+			// JavaLanguageServerPlugin.logException(e.getMessage(), e);
 			return true;
+		} finally {
+			if (process != null) {
+				if (!finished) {
+					process.destroyForcibly();
+				}
+				// Java doesn't call the finalize method when terminating a process.
```
fbricon: So let me rephrase:

> Terminating or destroying the Process doesn't close the process handle. It is only closed when the Process object is garbage collected (in its finalize() method).
> On Windows, when the Java LS is idle, we need to explicitly request a GC, to prevent an accumulation of zombie processes, as finalize() will be called.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/ParentProcessWatcher.java
```diff
@@ -83,14 +83,32 @@ private boolean parentProcessStillRunning() {
...
 			return true;
+		} finally {
+			if (process != null) {
+				if (!finished) {
+					process.destroyForcibly();
+				}
+				// Java doesn't call the finalize method when terminating a process. 
+				// Because of that, when Java LS isn't used, it creates a lot of zombie processes on Windows.
+				// GC calls finalize
+				// We can also call it on Linux and Mac because it runs when Java LS isn't active.
```
fbricon: While calling the GC on an idle server is not a big deal on Linux and Mac, it will still cause unnecessary CPU usage at this point. So we might as well do it only for Windows

snjeza: I have updated the PR.

fbricon:
@snjeza so it took me a while to test on windows because my Win 10 VM is excessively slow, but after the fix, I can see that running `handle64 -s -p <Java_LS_pid> | findstr Process` consistently returns 1, which is great, but I see that `handle64 -s -p <Java_LS_pid> | findstr File` shows file handles keep growing

fbricon:
The total amount of handles reported by handle64 is consistent with the handles reported in the task manager. And it keeps growing with the patch anyway

snjeza:
@fbricon checking...

NikolasKomonen:
Looks fine to me, caps at 26 file handlers

fbricon:
https://youtu.be/Z0fw1SmQKhI?t=52 tested on https://developer.microsoft.com/en-us/windows/downloads/virtual-machines

snjeza:
> https://youtu.be/Z0fw1SmQKhI?t=52 tested on https://developer.microsoft.com/en-us/windows/downloads/virtual-machines

@fbricon could you test on real Windows?

fbricon:
No I don't have the machines for that.
@yaohaizh could you check please (this PR or jdt.ls's)? I'm sure Microsoft has some Windows machines for you guys.

yaohaizh:
@fbricon  Tested on physical Windows system, both the file & process handles numbers ( ~133 & 1 on my machine) don't increase with the patch. Looks good to me.

fbricon:
Ok, let's merge it and see if more people can test it before we release an update.

NikolasKomonen:
@yaohaizh Heres the link to get the latest VSIX, if you have time could you check if it works on your machine.

http://download.jboss.org/jbosstools/vscode-xml/staging/?C=M;O=D

PR #308, opened 2019-03-04
Return Markdown-formatted hovers

### LINKED ISSUE #245: Return hover as Markdown
Currently, hover returns documentation as text, meaning if the content is actual html, then the client doesn't render it in a user-friendly way. See a hover over the `<packaging>` element of a pom.xml:

screen shot 2018-12-02 at 1 13 02 pm

We could probably transform the html doc into markdown, using the [remark](https://github.com/kotcrab/remark-java/) library, like in [jdt.ls](https://github.com/eclipse/eclipse.jdt.ls/blob/fc6b40e58808d23a324628a1e79e50f117c534de/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/javadoc/JavaDoc2MarkdownConverter.java). That'd also add jsoup as a dependency of the project.

fbricon:
Fixes #245

### DIFF
--- a/org.eclipse.lsp4xml/pom.xml
+++ b/org.eclipse.lsp4xml/pom.xml
@@ -107,6 +107,16 @@
 			<version>1.4.01</version>
 		</dependency>
 		<dependency>
+			<groupId>com.kotcrab.remark</groupId>
+			<artifactId>remark</artifactId>
+			<version>1.0.0</version>
+		</dependency>
+		<dependency>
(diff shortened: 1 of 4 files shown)


fbricon:
Still need to fix the case in https://github.com/angelozerr/lsp4xml/issues/245#issuecomment-457214267

And add unit tests obviously.

@NikolasKomonen knock yourself out :-)

PR #313, opened 2019-03-06
Saving an XSD should trigger validation on files that use it.

### LINKED ISSUE #213: Changing the content of XML Schema should trigger validation
Changing the content of XMl Schema should trigger validation for opened XML document.

NikolasKomonen:
Fixes #213

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLTextDocumentService.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLTextDocumentService.java
@@ -279,6 +279,9 @@ public class XMLTextDocumentService implements TextDocumentService {
 	}
 
 	@Override
+	/**
+	 * This method is triggered when the user types on an XML document.
+	 */
 	public void didChange(DidChangeTextDocumentParams params) {
 		documents.onDidChangeTextDocument(params);
 		triggerValidation(params.getTextDocument().getUri(), params.getTextDocument().getVersion());
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/SchemaLocation.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/SchemaLocation.java
@@ -10,9 +10,14 @@
  */
 package org.eclipse.lsp4xml.dom;
 
+import java.net.URI;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.StringTokenizer;
+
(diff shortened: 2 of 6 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/SchemaLocation.java
```diff
@@ -43,4 +45,22 @@ public String getLocationHint(String namespaceURI) {
...
+	/**
+	 * Given a schema URI, this will return true if the
+	 */
+	public boolean usesSchema(Path rootPath, Path schemaURI) {
+		if (schemaURI == null) {
+			return false;
+		}
+
+		for (String value : schemaLocationValuePairs.values()) {
+			Path currentSchemaURI = Paths.get(value);
```
fbricon: not sure this is a valid use case, but if schemaLocation uses an absolute path it might not be considered in use

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLTextDocumentService.java
```diff
@@ -279,6 +279,9 @@ public void didOpen(DidOpenTextDocumentParams params) {
 	}
 
 	@Override
+	/**
+	 * Triggers if the user types.
```
fbricon: Triggered?

NikolasKomonen: "The method triggers if the user types"

NikolasKomonen:
@fbricon updated

fbricon:
If I open /Users/fbricon/Downloads/sample-resources/foo.xml, with:
```xml
<root
        xsi:noNamespaceSchemaLocation='file:///Users/fbricon/Downloads/sample-resources/foo.xsd'
        xsi:nil='false'
        xmlns=''
        xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'>
    <e1>xxx</e1>
</root>
```

and then open and change foo.xsd in to make it fail validation:
```xml
<xsd:schema xmlns:xsd='http://www.w3.org/2001/XMLSchema'>
    <xsd:element
            name='root'
            nillable='true'>
        <xsd:complexType>
            <xsd:sequence>
                <xsd:element
                        name='e1'
                        type='xsd:integer'/>
            </xsd:sequence>
        </xsd:complexType>
    </xsd:element>
</xsd:schema>
```

Then saving the schema doesn't trigger the revalidation. Using xsi:noNamespaceSchemaLocation='/Users/fbricon/Downloads/sample-resources/foo.xsd' doesn't work either.

NikolasKomonen:
@fbricon updated, found out the Paths library didn't handle 'file://'

Issue #239, opened 2018-11-29
Autoclose self-closing tags


fbricon:
As an improvement to tag autoclosing, if we type `<foo /`, the closing `>` should be added automatically (controlled by a new preference, `autoCloseTags.selfClosingTags.enabled`)

angelozerr:
It should be cool if this option could manage too the same feature than WTP:

![foo](https://user-images.githubusercontent.com/1932211/49278475-30b81c00-f485-11e8-8418-4bb22cb24569.gif)

fbricon:
Yes that's pretty cool. Need to make sure it only works if tag is empty

PR #322, opened 2019-03-08
Stops autoClose of end tag if it already exists

### LINKED ISSUE #314: Dont autoclose tag if the closing tag already exists.
When I start with:

```xml
<a  </a>
```

then type '>'

the result is:

```xml
<a></a> </a>
```

Fix this so nothing happens in this case.

NikolasKomonen:
Fixes #314, #239


fbricon:
Problem is it prevents completion of tags if they're nested:
```xml
<abc>
    <abc|  <!-- closing with > doesn't autoclose the tag -->
</abc>
```

You need to check if tags are balanced

NikolasKomonen:
@fbricon I fixed the case you mentioned

fbricon:
No it's not working. And a simple "<a>" is not autocompleted either

fbricon:
Excellent! It brings tears of joy to my eyes ;-)

Issue #208, opened 2018-11-08
Add completion for xmlns attribute by default


fbricon:
So we can easily add a schema to the top level node of an xml document
```
<from xmlns="...">

</from>
```

angelozerr:
You mean completion on attribute value for xmlns?

If yes, we could use catalog for that?

fbricon:
to automatically add `xmlns="$0"`

PR #329, opened 2019-03-15
Completion for xmlns:xsi and xmlns

### LINKED ISSUE #326: Have value completion for 'xmlns:xsi'
Have a completion item for the following value:  
```xml
```

NikolasKomonen:
Fixes #326, #208


fbricon:
Mmmm
Screen Shot 2019-03-18 at 4 05 33 PM

NikolasKomonen:
@fbricon updated the `null:..` case

NikolasKomonen:
@fbricon fixed the case of `null:...`

PR #330, opened 2019-03-16
Make ParentProcessWatcher optional

### LINKED ISSUE #328: Make ParentProcessWatcher optional
The `ParentProcessWatcher` functionality is extremely inefficient in some scenarios. For example, when we're running on Windows under VS Code as launched by [Red Hat's VS Code XML](https://github.com/redhat-developer/vscode-xml) extension a separate instance of the lsp4xml language server will be launched per tab where the extension is loaded. Each of these tabs will independently start checking if the parent process which launched the LSP (`Code.exe`) has exited after the idle period. Note the idle period is simply calculated as >30 seconds since a message was last received from the client, so this tends to be reached quickly.

Each check for if the parent process has exited will typically spawn four unique short-lived processes: `cmd.exe`, `conhost.exe`, `tasklist.exe`, and `findstr.exe`. This is multiplied by the number of instances of the LSP which are running. Somewhat ironically, the entire check is actually unnecessary in this scenario as the LSP is launched by VS Code as part of a Windows Job object which guarantees that if it exits the child processes it spawns will be terminated with it.

I propose a flag which can be passed to the LSP on startup to disable the parent process watcher when it's being run in an environment that will handle its termination. There's separately scope for improving the underlying `ParentProcessWatcher` implementation for when it is required but that should be addressed in a separate issue.

snjeza:
Fixes #328 

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLServerLauncher.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLServerLauncher.java
@@ -16,8 +16,10 @@ import java.net.Authenticator;
 import java.net.PasswordAuthentication;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.function.Function;
 
 import org.eclipse.lsp4j.jsonrpc.Launcher;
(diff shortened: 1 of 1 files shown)


fbricon:
@snjeza could you manage to get zombie lsp4xml processes after killing its parent process (vscode specifically)

PR #331, opened 2019-03-18
Fix processing instruction with attributes in formatting

NikolasKomonen:
Fixes https://github.com/redhat-developer/vscode-xml/issues/115

Also fixes an issue with the Prolog removing unknown attributes on format.

Signed-off-by: Nikolas <nikolaskomonen@gmail.com>

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/DOMAttr.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/DOMAttr.java
@@ -108,9 +108,8 @@ public class DOMAttr extends DOMNode implements org.w3c.dom.Attr {
 	}
 
 	/*
-	 * (non-Javadoc)
 	 * 
-	 * @see org.w3c.dom.Attr#getValue()
+	 * Returns the attribute's value without quotes.
 	 */
 	@Override
 	public String getValue() {
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/Constants.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/Constants.java
@@ -60,17 +60,22 @@ public class Constants {
 
 	public static final Pattern ELEMENT_NAME_REGEX = Pattern.compile("^[_:\\w][_:\\w-.\\d]*");
 
-	public static final Pattern ATTRIBUTE_NAME_REGEX = Pattern.compile("^[^\\s\"'<>/=\\x00-\\x0F\\x7F\\x80-\\x9F]*");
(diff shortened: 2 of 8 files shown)


fbricon:
why do you need to treat xml-stylesheet differently than any other PI?

NikolasKomonen:
@fbricon 

Based off [the wiki](https://en.wikipedia.org/wiki/Processing_Instruction)  it looks like `xml-stylesheet` is a common processing instruction that is defined to have attributes. Other processing instructions have text content in them (<?mvn  aaa bbb ccc ?>. 

It helps in the parsing since we know what to look for after a certain PI

fbricon:
In http://www.sagehill.net/docbookxsl/ProcessingInstructions.html (linked from the wiki), you can see 

> Processing instructions
> A processing instruction is a means of passing hints to the XSLT processor about how to handle something. XML permits the placement of processing instructions just about anywhere in a document. A processing instruction (sometimes referred to as a PI) uses a special syntax, <?name content ?>, to indicate that it is not an element, comment, or text. For example:
> 
> `<?dbfo bgcolor="#EEEEEE" ?>`
> The first word in a PI is the name. In this example, the name of the processing instruction is dbfo and the content is bgcolor="#EEEEEE". This style of PI content could be called a “pseudo attribute”, because it provides a name-value pair similar to a real attribute (which only elements can have). You can put several such pseudo attributes in one PI, or you can put each of them in their own PI.
> 
> 

So xml-stylesheet should not be a special case

NikolasKomonen:
@fbricon I know this is the case, but the regex in the xml specification considers the PI content [as text](https://www.w3.org/TR/REC-xml/#sec-pi).

What pr does is allow common PI declarations to be defined to be parsed with their attributes considered. If not then it defaults to saving all the content as a single string.

fbricon:
can't you split content by whitespace and then check if you have key/value patterns in each group?

NikolasKomonen:
@fbricon something like this wouldn't work in that case:

```xml
<?aaa b =   "123" c  ="234 ?>
```

fbricon:
Mar-20-2019 02-07-07

Should be the same behavior for all PIs

NikolasKomonen:
@fbricon What if there exists a PI that looks like
```xml
<?whatever foo   bar  = hello ?>
<?whatever this    should not be   formatted ?>
<?whatever This thing = no formatting ?>
```
at what point should it decide to format the content inside?

fbricon:
any = sign you find indicates this is the middle of a key/value pair

fbricon:
I still think all PIs should be treated like xml-stylesheet. If somehow we got a complaint, we can revisit, but that seems the most sensible, logical behavior.

NikolasKomonen:
@fbricon Updated, all PI's are formatted the same (not touched)

PR #332, opened 2019-03-18
Fixed EOF Error

### LINKED ISSUE #157: 'textDocument/publishDiagnostics' failed with message: Illegal argument: line must be non-negative
When debugging in vscode, when pasting `<?xml version="1.0" encoding="UTF-8"?>` into a new empty document, the host vscode instance displays an error:

> Notification handler 'textDocument/publishDiagnostics' failed with message: Illegal argument: line must be non-negative

The server sent :
```json
Params: {
    "uri": "file:///Users/fbricon/Dev/xml/foo.xml",
    "diagnostics": [
        {
            "range": {
                "start": {
                    "line": -2,
                    "character": -2
                },
                "end": {
                    "line": -2,
                    "character": -2
                }
            },
            "severity": 1,
            "code": "PrematureEOF",
            "source": "xml",
            "message": "Premature end of file."
        }
    ]
}

```

NikolasKomonen:
Fixes #157

The most common ocurrences of this error are
when the document is empty or there is just the
xml prolog (`<?xml ... ?>`). All errors that give negative values
will be ignored for now.

Signed-off-by: Nikolas <nikolaskomonen@gmail.com>

### DIFF
--- a/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
+++ b/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
@@ -28,6 +28,18 @@ import org.junit.Test;
 public class XMLSchemaDiagnosticsTest {
 
 	@Test
+	public void prematureEOFNoErrorReported() throws Exception {
+		String xml = " ";
+		testDiagnosticsFor(xml);
+	}
+
+	@Test
+	public void prematureEOFWithPrologNoErrorReported() throws Exception {
+		String xml = "<?xml version=\"1.0\" encoding=\"UTF-8\" ?> ";
+		testDiagnosticsFor(xml);
+	}
+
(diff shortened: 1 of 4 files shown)


fbricon:
unit test?

fbricon:
I'd expect an error to be raised to the user. When only the xml prolog is present in a file, Intellij shows: "Valid XML document should have a root tag".

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
```diff
@@ -27,6 +27,18 @@
  */
 public class XMLSchemaDiagnosticsTest {
```
fbricon: so right now PrematureEOF doesn't generate an error for empty docs, which is debatable, but let's say we keep that behavior for now. At least the test methods should be changed to *NoError, since no diagnostic is expected

NikolasKomonen:
@fbricon fixed the unit test names

PR #333, opened 2019-03-19
CodeActions for missing attribute quotations

### LINKED ISSUE #137: Code action to close missing quotes for attributes

NikolasKomonen:
Fixes #137


fbricon:
When the code action is called, it ignores the expected value.

Mar-20-2019 01-37-03

Also, there's no code action if the quotation is unbalanced, like `<a b="value></a>`
No code action either when the cursor is on the quote-less value

NikolasKomonen:
@fbricon I was thinking it might not be safe in some cases.
What do you think about doing it if there is no white space between the '=' and the value?

NikolasKomonen:
@fbricon updated to surround next value with quotes

PR #348, opened 2019-04-08
Fixed file path issue on Windows

### LINKED ISSUE #125: Duplicate suggestions
image

Noteworthy is that one `Name` suggestion shows docs, the other doesn't. Looks like one is inferred from the existing `Name` tag.
Expected would be 0 suggestions for `Name`, because `Name` must only appear once.

Schema is here: https://github.com/felixfbecker/PowerShellXSD/blob/master/Types.xsd

Relevant section:
```xml
<xs:complexType name="Type">
    <xs:sequence>
      <xs:element name="Name" type="xs:string">
        <xs:annotation>
          <xs:documentation>A pair of `&lt;Name&gt;` tags that enclose the name of the affected .NET Framework type.</xs:documentation>
        </xs:annotation>
      </xs:element>
      <xs:element minOccurs="0" name="Members" type="Members" />
      <xs:element minOccurs="0" name="TypeConverter" type="TypeConverter" />
      <xs:element minOccurs="0" name="TypeAdapter" type="TypeAdapter" />
    </xs:sequence>
  </xs:complexType>
```

NikolasKomonen:
Fixes https://github.com/redhat-developer/vscode-xml/issues/125

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
@@ -10,6 +10,7 @@
  */
 package org.eclipse.lsp4xml.utils;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Writer;
@@ -62,10 +63,17 @@ public class FilesUtils {
 	/**
 	 * Given a file path as a string, will normalize it
 	 * and return the normalized string if valid, or null if not.
+	 * 
+	 * The '~' home symbol will be converted into the actual home path.
+	 * Slashes will be corrected depending on the OS.
 	 */
 	public static String normalizePath(String pathString) {
-		if(pathString != null && !pathString.isEmpty()) {
-			pathString = pathString.replaceFirst("^~", System.getProperty("user.home"));
+		if (pathString != null && !pathString.isEmpty()) {
+			if (pathString.indexOf("~") == 0) {
+				pathString = System.getProperty("user.home") + (pathString.length() > 1? pathString.substring(1):"");
+			}
+			pathString = pathString.replace("/", File.separator);
+			pathString = pathString.replace("\\", File.separator);
 			Path p = Paths.get(pathString);
 			pathString = p.normalize().toString();
 			return pathString;
--- a/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/utils/FilesUtilsTest.java
+++ b/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/utils/FilesUtilsTest.java
@@ -17,6 +17,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 
 import org.junit.Test;
+import static java.io.File.separator;
 
 /**
  * FilesUtilsTest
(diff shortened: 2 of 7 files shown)


org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/utils/ProjectUtils.java
```diff
@@ -24,6 +24,7 @@
 	 */
 	public static Path getProjectDirectory() {
 		String currPath = ProjectUtils.class.getClassLoader().getResource("").getFile();
+		currPath = FilesUtils.normalizePath(currPath);
```
fbricon: why is this necessary?

NikolasKomonen: On Windows the path looking like "\C:\Users\..." was being used and it broke, "\C:\Users\..." the leading '\' broke it because it is not valid in Windows. normalizePath() checks for that.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -62,15 +66,26 @@ public static void resetDeployPath() {
...
-		if(pathString != null && !pathString.isEmpty()) {
-			pathString = pathString.replaceFirst("^~", System.getProperty("user.home"));
-			Path p = Paths.get(pathString);
-			pathString = p.normalize().toString();
-			return pathString;
-		}
-		return null;
+		
+		if (pathString != null && !pathString.isEmpty()) {
+            if (pathString.indexOf("~") == 0) {
```
fbricon: indentation is hosed

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/utils/ProjectUtils.java
```diff
@@ -24,6 +24,7 @@
 	 */
 	public static Path getProjectDirectory() {
 		String currPath = ProjectUtils.class.getClassLoader().getResource("").getFile();
```
fbricon:
```java
String currPath = new File(ProjectUtils.class.getClassLoader().getResource("").getPath()).toString();
```
still ugly, but should remove the leading `/`

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/utils/FilesUtilsTest.java
```diff
@@ -27,18 +28,19 @@
...
 	}
 
 	@Test
 	public void normalizePathTest() {
-		assertEquals(System.getProperty("user.home") + "/Test/Folder", FilesUtils.normalizePath("~/Test/Folder"));
-		assertEquals("/Test/~/Folder", FilesUtils.normalizePath("/Test/~/Folder"));
-		assertEquals("~/Test/Folder", FilesUtils.normalizePath("./~/Test/Folder"));
-		assertEquals("/Folder", FilesUtils.normalizePath("/Test/../Folder"));
+		assertEquals(Paths.get(System.getProperty("user.home"), "Test", "Folder").toString(), FilesUtils.normalizePath("~/Test/Folder"));
+		assertEquals(Paths.get(separator, "Test", "~", "Folder").toString(), FilesUtils.normalizePath("/Test/~/Folder"));
```
NikolasKomonen: @fbricon This test was failing for me on Windows, an extra '\' was inserted at the beginning of the output of Paths.get(...) could you check if you have the same issue

NikolasKomonen:
@fbricon updated

PR #353, opened 2019-04-15
Forward slash in bad places caused autoclose to break

### LINKED ISSUE #126: Language server cannot find schema after a while
After I leave a file open in the background for a while, then come back to it, I often see that the root element is not recognised anymore. Restarting solves it, so it cannot be a problem with the schema.

NikolasKomonen:
Fixes https://github.com/redhat-developer/vscode-xml/issues/126

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/XMLScanner.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/XMLScanner.java
@@ -263,7 +263,7 @@ public class XMLScanner implements Scanner {
 			}
 			return finishToken(offset, TokenType.Unknown);
 
-		case WithinTag:
+		case WithinTag: {
 			if (stream.skipWhitespace()) {
 				return finishToken(offset, TokenType.Whitespace);
 			}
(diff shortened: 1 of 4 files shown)


fbricon:
Got a weird completion on the forward slash here:
Apr-15-2019 15-43-43

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/XMLScanner.java
```diff
@@ -272,18 +272,26 @@ TokenType internalScan() {
...
-				state = ScannerState.WithinContent;
+				state = ScannerState.WithinTag;
 				if(stream.advanceIfChar(_RAN)) { // >
+					state = ScannerState.WithinContent;
 					return finishToken(offset, TokenType.StartTagSelfClose);
 				}
+				return finishToken(offset, TokenType.Unknown);
+			}
+		
+			if (stream.peekChar() == _DQO || stream.peekChar() == _SIQ) { // " || '
```
fbricon: maybe store the stream.peekChar() value to avoid all the underlying computations calling peekChar() everytime

NikolasKomonen:
@fbricon updated, the weird completion on the forward slash was to do with it thinking it was at the end of a self closed tag, instead of being only after the '/'

fbricon:
Ok it works much better. Now there's something I find annoying:
if you have a non empty tag, mistakenly type `/` right before the closing bracket, then autoclose will 
return `"snippet": ">$0"` so the cursor jumps right after the closing bracket. Now if you hit delete, you remove `>` instead of that `/`. I don't think we should return anything if there's nothing to autoclose.
Apr-16-2019 18-50-35

NikolasKomonen:
@fbricon updated not to do the snippet on self closing with content.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/XMLCompletions.java
```diff
@@ -368,10 +370,21 @@ public AutoCloseTagResponse doTagComplete(DOMDocument xmlDocument, Position posi
...
-					Integer closeBracket = element1.isNextChar('>', offset); // After the slash is a close bracket
+					List<DOMAttr> attrList = element1.getAttributeNodes();
+					if(attrList != null) {
+						DOMAttr lastAttr = attrList.get(attrList.size() - 1);
+						if(slashOffset < lastAttr.getEnd()) { //slash in attribute value
+							return null;
+						}
+					}
+					String text = xmlDocument.getText();
+					Boolean closeBracketAfterSlash = offset < text.length() ? text.charAt(offset) == '>' : false; // After the slash is a close bracket
```
fbricon: boolean closeBracketAfterSlash

fbricon:
You need to change the commit message, as it doesn't close #126 but https://github.com/redhat-developer/vscode-xml/issues/126

NikolasKomonen:
@fbricon updated

fbricon:
- forward slash in closing tag should do nothing
- tags containing empty space should not autoclose

NikolasKomonen:
>* tags containing empty space should not autoclose

This requires incremental parsing to be working since it needs to check the previous state before it can decide if there was an empty space, we can merge now and open new issue

PR #359, opened 2019-04-23
Finalizes incremental settings

### LINKED ISSUE #133: Experimental Incremental support
In large XML file, vscode like Eclipse freezes as soon as you type something. One of problem is because the didChange command send a big JSON content every time.

One idea that I have is to se the Incremental support on the server side, and didChange update the string of stored TextDocument.

I have started to implement this idea, but I think we should have a preferences like "experimentalIncrementalSupport" because I fear that there are some bugs.

@fbricon @NikolasKomonen what do you think about this experimentalIncrementalSupport preferences?

NikolasKomonen:
Fixes #133

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocument.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocument.java
@@ -26,12 +26,13 @@ import org.eclipse.lsp4j.TextDocumentItem;
  */
 public class TextDocument extends TextDocumentItem {
 
+	private final Object lock = new Object();
+
 	private static String DEFAULT_DELIMTER = System.lineSeparator();
 
-	private ListLineTracker lineTracker;
+	private final ListLineTracker lineTracker;
 
-	// Buffer of the text document used only in incremental mode.
-	private StringBuilder buffer;
+	private boolean incremental;
 
 	public TextDocument(TextDocumentItem document) {
 		this(document.getText(), document.getUri());
@@ -40,22 +41,23 @@ public class TextDocument extends TextDocumentItem {
 	}
 
 	public TextDocument(String text, String uri) {
+		this.lineTracker = new ListLineTracker();
 		super.setUri(uri);
-		super.setText(text);
+		this.setText(text);
 	}
 
 	public void setIncremental(boolean incremental) {
-		if (incremental) {
-			buffer = new StringBuilder(getText());
-		} else {
-			buffer = null;
-		}
+		this.incremental = incremental;
+	}
+
+	public boolean isIncremental() {
+		return incremental;
 	}
 
 	@Override
 	public void setText(String text) {
 		super.setText(text);
-		lineTracker = null;
+		lineTracker.set(text);
 	}
 
 	public Position positionAt(int position) throws BadLocationException {
@@ -117,10 +119,6 @@ public class TextDocument extends TextDocumentItem {
 	}
 
 	private ListLineTracker getLineTracker() {
-		if (lineTracker == null) {
-			lineTracker = new ListLineTracker();
-			lineTracker.set(super.getText());
-		}
 		return lineTracker;
 	}
 
@@ -137,8 +135,17 @@ public class TextDocument extends TextDocumentItem {
 		}
 		if (isIncremental()) {
 			try {
-				synchronized (buffer) {
-					for (TextDocumentContentChangeEvent changeEvent : changes) {
+				// Initialize buffer and line tracker from the current text document
+				String initialText = getText();
+				StringBuilder buffer = new StringBuilder(getText());
+				ListLineTracker lt = new ListLineTracker();
+				lt.set(initialText);
+				synchronized (lock) {
+					for (int i = 0; i < changes.size(); i++) {
+						if (i > 0) {
+							lt.set(buffer.toString());
+						}
+						TextDocumentContentChangeEvent changeEvent = changes.get(i);
 
 						Range range = changeEvent.getRange();
 						int length = 0;
@@ -147,17 +154,18 @@ public class TextDocument extends TextDocumentItem {
 							length = changeEvent.getRangeLength().intValue();
 						} else {
 							// range is optional and if not given, the whole file content is replaced
-							length = getText().length();
-							range = new Range(positionAt(0), positionAt(length));
+							length = buffer.length();
+							range = new Range(lt.getPositionAt(0), lt.getPositionAt(length));
 						}
 						String text = changeEvent.getText();
-						int startOffset = offsetAt(range.getStart());
+						int startOffset = lt.getOffsetAt(range.getStart());
 						buffer.replace(startOffset, startOffset + length, text);
 					}
+					// Update the new text content from the updated buffer
 					setText(buffer.toString());
 				}
 			} catch (BadLocationException e) {
-				// Should never occurs.
+				// Should never occur.
 			}
 		} else {
 			// like vscode does, get the last changes
@@ -168,10 +176,6 @@ public class TextDocument extends TextDocumentItem {
 				setText(last.getText());
 			}
 		}
-	}
-
-	public boolean isIncremental() {
-		return buffer != null;
 	}
 
 }
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocuments.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocuments.java
@@ -25,9 +25,9 @@ import org.eclipse.lsp4j.TextDocumentItem;
  */
 public class TextDocuments implements ITextDocumentFactory {
 
-	private boolean incremental;
-
 	private final Map<String, TextDocument> documents;
+
+	private boolean incremental = true; //default on
 
 	public TextDocuments() {
 		documents = new HashMap<>();
@@ -72,7 +72,7 @@ public class TextDocuments implements ITextDocumentFactory {
 	@Override
 	public TextDocument createDocument(TextDocumentItem document) {
 		TextDocument doc = new TextDocument(document);
-		doc.setIncremental(incremental);
+		doc.setIncremental(isIncremental());
 		return doc;
 	}
 
@@ -121,4 +121,5 @@ public class TextDocuments implements ITextDocumentFactory {
 			return documents.values();
 		}
 	}
+
 }
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/settings/XMLIncrementalSupportSettings.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/settings/XMLIncrementalSupportSettings.java
@@ -14,11 +14,14 @@ package org.eclipse.lsp4xml.settings;
  * XML experimental incremental support capabilities.
  *
  */
-public class XMLIncrementalSupportCapabilities {
+public class XMLIncrementalSupportSettings {
 
 	private Boolean enabled;
 
 	public Boolean getEnabled() {
+		if(enabled == null) {
+			enabled = true; // default on
+		}
 		return enabled;
 	}
 
--- a/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/IncrementalParsingTest.java
+++ b/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/IncrementalParsingTest.java
@@ -1,0 +1,250 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+
+package org.eclipse.lsp4xml.services;
+
+import java.util.ArrayList;
+
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.TextDocumentContentChangeEvent;
+import org.eclipse.lsp4xml.commons.TextDocument;
+import org.junit.Test;
+import static org.junit.Assert.*;
+
+/**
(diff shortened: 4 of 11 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocuments.java
```diff
@@ -26,21 +26,13 @@
  */
 public class TextDocuments implements ITextDocumentFactory {
 
-	private boolean incremental;
+	private final Boolean incremental;
```
fbricon: why do you need a Boolean?

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/IncrementalParsingTest.java
```diff
@@ -0,0 +1,253 @@
+/*******************************************************************************
+* Copyright (c) 2019 Z and others.
```
fbricon: who is Z?

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocuments.java
NikolasKomonen: was used as a safety safeguard for a previous implementation, removed now.

NikolasKomonen:
@fbricon updated

NikolasKomonen:
Current issues, a large file will break incremental parsing. If many characters are typed quickly the document on the server side ends up misrepresenting the actual document.

Difficult to debug this.

angelozerr:
@NikolasKomonen please see https://github.com/angelozerr/lsp4xml/pull/369 which is a PR just with an idea which could fix perhaps bug with large file. Not sure it works well, but this PR is just to open discussion

angelozerr:
Nick i seen that you integrated my work . could you give me feedback please if it fix your problem which is difficult to debug. Thanks

NikolasKomonen:
@angelozerr I've been testing it on a file with 140,000 lines and it seems to be working properly. There are no more synchronization issues.

The file, nasa.xml, in src/test/resources/xml will crash the extension. I'm not sure if that is on VSCode or lsp4xml.

Otherwise, from a regular use case it looks good to me for now.

angelozerr:
Thanks for your feedback.

Could you explain more the use cases which crashes vscode please. Imho we need to improve performance and or memory problem because i see more and more issues about that

NikolasKomonen:
For the nasa.xml file, while waiting for initialization of the server to finish VS and return to the client side I get an error from VSCode "Extension host terminated unexpectedly"

Error:
```
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [/usr/share/code/libnode.so]
 2: 0x7fe6994ea867 [/usr/share/code/libnode.so]
 3: 0x7fe698c2ae5e [/usr/share/code/libnode.so]
 4: 0x7fe698c2ade9 [/usr/share/code/libnode.so]
 5: v8::internal::Factory::NewFillerObject(int, bool, v8::internal::AllocationSpace) [/usr/share/code/libnode.so]
 6: 0x7fe699123670 [/usr/share/code/libnode.so]
 7: 0x22976a842dd
```

From other issues I've looked at it could be to do with node and
throwing the flag`--max_old_space_size=8192` in the node build cli command could fix it.
I'm just not sure where that would be. @fbricon any ideas?

fbricon:
Looks like vscode doesn't have enough memory to open the document, if you can reproduce that after disabling all the vscode extensions, then you should open a ticket in https://github.com/Microsoft/vscode/issues.

NikolasKomonen:
Per @fbricon 's recommendation, I tested the nasa.xml file with the HTML language server and that had the same error with the HTMLLS crashing too.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocument.java
```diff
@@ -147,17 +149,17 @@ public void update(List<TextDocumentContentChangeEvent> changes) {
 							length = changeEvent.getRangeLength().intValue();
 						} else {
 							// range is optional and if not given, the whole file content is replaced
-							length = getText().length();
-							range = new Range(positionAt(0), positionAt(length));
+							length = buffer.toString().length();
```
angelozerr: Change with length = buffer.length(); to avoid creating an instance of String

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocuments.java
```diff
@@ -26,21 +26,13 @@
...
-	}
-
-	/**
-	 * Set the incremental support.
-	 * 
-	 * @param incremental
-	 */
-	public void setIncremental(boolean incremental) {
-		this.incremental = incremental;
+		incremental = true; // Quick fix to set incremental
```
angelozerr: Please remove this hard coded incremental support and let's the user to set incremental support as it's a major changed. Once we will have good feedback from users, we will switch to incremental support every time.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/settings/XMLIncrementalSupportSettings.java
```diff
@@ -14,11 +14,14 @@
...
  *
  */
-public class XMLIncrementalSupportCapabilities {
+public class XMLIncrementalSupportSettings {
 
 	private Boolean enabled;
 
 	public Boolean getEnabled() {
+		if(enabled == null) {
+			enabled = true; // default on
```
angelozerr: Please remove this defaut value as it's a major change.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocument.java
```diff
@@ -139,6 +139,8 @@ public void update(List<TextDocumentContentChangeEvent> changes) {
 			try {
 				synchronized (buffer) {
```
angelozerr: Please remove buffer field because the TextDocument stores 2 xml content in this case (buffer and text field). It's not good for memory (2 XML content stored). So the idea is to create StringBuilder instance from the current text in the update method to benefit with StringBuilder#replace  method. and synchronized to a lock field instance (not to the buffer field which must be removed)

...java
private final Object lock = new Object();

...
public void setIncremental(boolean incremental) {
		this.incremental = incremental;
	}

	public boolean isIncremental() {
		return incremental;
	}
...
public void update(List<TextDocumentContentChangeEvent> changes) {
		if (isIncremental()) {
			try {
				StringBuilder buffer = new StringBuilder(getText());
				synchronized (lock) {
...

An another change is to set lineTracker as final because the call of getLineTracker can be created several time when there are several threads which call it in the same time, we could use synchronized but we need not that because lineTracker.set exists as methods.

```java
private final ListLineTracker lineTracker;

	public TextDocument(String text, String uri) {
		this.lineTracker = new ListLineTracker();
		super.setUri(uri);
		setText(text);
	}
       ...

	@Override
	public void setText(String text) {
		super.setText(text);
		lineTracker.set(text);
	}
```

Here the final TextDocument with those 2 changes:

```java
/**
 *  Copyright (c) 2018 Angelo ZERR.
 *  All rights reserved. This program and the accompanying materials
 *  are made available under the terms of the Eclipse Public License v2.0
 *  which accompanies this distribution, and is available at
 *  http://www.eclipse.org/legal/epl-v20.html
 *
 *  Contributors:
 *  Angelo Zerr <angelo.zerr@gmail.com> - initial API and implementation
 */
package org.eclipse.lsp4xml.commons;

import java.util.List;
import java.util.regex.Matcher;
import java.util.regex.Pattern;

import org.eclipse.lsp4j.Position;
import org.eclipse.lsp4j.Range;
import org.eclipse.lsp4j.TextDocumentContentChangeEvent;
import org.eclipse.lsp4j.TextDocumentItem;

/**
 * Text document extends LSP4j {@link TextDocumentItem} to provide methods to
 * retrieve position.
 *
 */
public class TextDocument extends TextDocumentItem {

	private static String DEFAULT_DELIMTER = System.lineSeparator();

	private final ListLineTracker lineTracker;

	private boolean incremental;

	private final Object lock = new Object();

	public TextDocument(TextDocumentItem document) {
		this(document.getText(), document.getUri());
		super.setVersion(document.getVersion());
		super.setLanguageId(document.getLanguageId());
	}

	public TextDocument(String text, String uri) {
		this.lineTracker = new ListLineTracker();
		super.setUri(uri);
		setText(text);
	}

	public void setIncremental(boolean incremental) {
		this.incremental = incremental;
	}

	public boolean isIncremental() {
		return incremental;
	}

	@Override
	public void setText(String text) {
		super.setText(text);
		lineTracker.set(text);
	}

	public Position positionAt(int position) throws BadLocationException {
		ListLineTracker lineTracker = getLineTracker();
		return lineTracker.getPositionAt(position);
	}

	public int offsetAt(Position position) throws BadLocationException {
		ListLineTracker lineTracker = getLineTracker();
		return lineTracker.getOffsetAt(position);
	}

	public String lineText(int lineNumber) throws BadLocationException {
		ListLineTracker lineTracker = getLineTracker();
		Line line = lineTracker.getLineInformation(lineNumber);
		String text = super.getText();
		return text.substring(line.offset, line.offset + line.length);
	}

	public String lineDelimiter(int lineNumber) throws BadLocationException {
		ListLineTracker lineTracker = getLineTracker();
		String lineDelimiter = lineTracker.getLineDelimiter(lineNumber);
		if (lineDelimiter == null) {
			if (lineTracker.getNumberOfLines() > 0) {
				lineDelimiter = lineTracker.getLineInformation(0).delimiter;
			}
		}
		if (lineDelimiter == null) {
			lineDelimiter = DEFAULT_DELIMTER;
		}
		return lineDelimiter;
	}

	public Range getWordRangeAt(int textOffset, Pattern wordDefinition) {
		try {
			Position pos = positionAt(textOffset);
			ListLineTracker lineTracker = getLineTracker();
			Line line = lineTracker.getLineInformation(pos.getLine());
			String text = super.getText();
			String lineText = text.substring(line.offset, textOffset);
			int position = lineText.length();
			Matcher m = wordDefinition.matcher(lineText);
			int currentPosition = 0;
			while (currentPosition != position) {
				if (m.find()) {
					currentPosition = m.end();
					if (currentPosition == position) {
						return new Range(new Position(pos.getLine(), m.start()), pos);
					}
				} else {
					currentPosition++;
				}
				m.region(currentPosition, position);
			}
			return new Range(pos, pos);
		} catch (BadLocationException e) {
			return null;
		}
	}

	private ListLineTracker getLineTracker() {
		return lineTracker;
	}

	/**
	 * Update text of the document by using the changes and according the
	 * incremental support.
	 * 
	 * @param changes the text document changes.
	 */
	public void update(List<TextDocumentContentChangeEvent> changes) {
		if (changes.size() < 1) {
			// no changes, ignore it.
			return;
		}
		if (isIncremental()) {
			try {
				StringBuilder buffer = new StringBuilder(getText());
				synchronized (lock) {
					for (TextDocumentContentChangeEvent changeEvent : changes) {
						Range range = changeEvent.getRange();
						int length = 0;

						if (range != null) {
							length = changeEvent.getRangeLength().intValue();
						} else {
							// range is optional and if not given, the whole file content is replaced
							length = buffer.length();
							range = new Range(positionAt(0), positionAt(length));
						}
						String text = changeEvent.getText();
						int startOffset = offsetAt(range.getStart());
						buffer.replace(startOffset, startOffset + length, text);
					}
					setText(buffer.toString());
				}
			} catch (BadLocationException e) {
				// Should never occurs.
			}
		} else {
			// like vscode does, get the last changes
			// see
			// https://github.com/Microsoft/vscode-languageserver-node/blob/master/server/src/main.ts
			TextDocumentContentChangeEvent last = changes.size() > 0 ? changes.get(changes.size() - 1) : null;
			if (last != null) {
				setText(last.getText());
			}
		}
	}

}

```

angelozerr: @NikolasKomonen are you sure that you have done the update of TextDocument like I suggest below?

NikolasKomonen: @angelozerr I'm getting offset issues with the changes. Did you try running tests with your implementation?

angelozerr: ?Yes it should work. What is your test (and traces) which failed ?

NikolasKomonen: My failures are that it "Couldnt get the position at Offset", and BadLocation errors.

I pushed the changes to this PR, I'm sorry if I missed anything from your changes.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocument.java
```diff
@@ -40,22 +41,23 @@ public TextDocument(TextDocumentItem document) {
...
+
+	public boolean isIncremental() {
+		return incremental;
 	}
 
 	@Override
 	public void setText(String text) {
 		super.setText(text);
-		lineTracker = null;
+		lineTracker.set(text);;
```
angelozerr: remove ';'

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/TextDocument.java
```diff
@@ -40,22 +41,23 @@ public TextDocument(TextDocumentItem document) {
 	}
 
 	public TextDocument(String text, String uri) {
+		this.lineTracker = new ListLineTracker();
 		super.setUri(uri);
 		super.setText(text);
```
angelozerr: change super to this (this.setText(text)) which should fix tests falures (sorry with that!)

angelozerr: change super to this (this.setText(text)) which should fix tests failures (sorry with that!)

PR #369, opened 2019-05-08
Try to fix incremental with large file

angelozerr:
This PR is just a POC (please don't merge it) to try to fix incrememntal support with big file (not tested).

The basic idea is to be sure that ListLineTracker which is used to retreive offset is synchronized with the current buffer. 

It's just an idea, not sure it's a clean way


NikolasKomonen:
PR https://github.com/angelozerr/lsp4xml/pull/359 adds onto this.

PR #391, opened 2019-05-24
Fix https://github.com/angelozerr/lsp4xml/issues/386

angelozerr:
This PR fixes https://github.com/angelozerr/lsp4xml/issues/386 The problem comes from of populate of DTD element children when it is loaded. As elements can be declared in many area, we need to check that children element name was not added.

This PR is just to see how to fix problem. Please ignore it and creates a new PR with tests.


fbricon:
@angelozerr can you please use a self-descriptive commit message. Same for the PR

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/dtd/contentmodel/CMDTDDocument.java
```diff
@@ -101,7 +101,9 @@ public void startContentModel(String elementName, Augmentations augs) throws XNI
 
 	@Override
 	public void element(String elementName, Augmentations augs) throws XNIException {
-		hierachies.add(elementName);
+		if (!hierachies.contains(elementName)) {
```
fbricon: why not use a Set instead of a List? LinkedHashSet if order is important

fbricon:
superseded by #396

Issue #389, opened 2019-05-24
Memory usage improvement


angelozerr:
When XML large file is managed by the XML language server like https://github.com/angelozerr/lsp4xml/blob/master/org.eclipse.lsp4xml/src/test/resources/xml/nasa.xml (23.9MB), the memory grows up a lot (it can take 5GB!)

This issue is to improve memory problem. The first improvement that we can do is to improve memory with parse of DOM Document. It seems scanner is OK, but creating of DOM Document can be improved:

 * use int instead of Integer (ex: in DOMElement). Using Integer provides a problem with memory. Indeed autoboxing create a lot of Integer which are cached. Wen nasa.xml is parsed every time (without changes) we can see that there are a lot of memory used by Integer. To fix this problem, int should be used.

 * Memory Leak problem with DOMNode. DOMNode stores owner document, owner element, list of attributes with parent DOM node link. To help GC, we should have a dispose method in DOMNode to set parent, owner element, owncer document to null. It seems that with this fix, the memory keeps in 2GB after 10 minutes of parse instead of 5GB.

After thoses 2 fixes, we must continue to study problems (ex: execute symbol provider every time, etc) to try to display nasa.xml in vscode because today we cannot display it.

angelozerr:
The XML scanner seems OK. If we parse the big file nasa.xml with scanner every time (see https://github.com/angelozerr/lsp4xml/blob/memory-improvement/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/performance/XMLScannerPerformance.java) we have this benchmark:

![image](https://user-images.githubusercontent.com/1932211/58340513-59c1d400-7e4c-11e9-83f7-ad12757e51a3.png)

It takes around 100MB

angelozerr:
The DOM parser has problem with memory. If we parse the big file nasa.xml with DOM parser every time (see https://github.com/angelozerr/lsp4xml/blob/memory-improvement/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/performance/DOMParserPerformance.java) we have this benchmark:

![image](https://user-images.githubusercontent.com/1932211/58344519-30a64100-7e56-11e9-8432-3f4e81de9561.png)

It takes around 5-6GB!

angelozerr:
We can see that there are a lot of Integer instances which takes a lot memories:

[...]

PR #393, opened 2019-05-27
Don't store the owner document instance in each nodes.

angelozerr:
Don't store the owner document instance in each nodes (see #389). This PR improves the memory with big file (there is around 1GB gain with nasa.xml file)


angelozerr:
Here the result when nasa.xml is parsed with DOM Parser with existing code:

image

Here the result when nasa.xml is parsed with DOM Parser with this PR (owner document is not stored in each nodes instances):

image

There is around 1GB of gain.

PR #395, opened 2019-05-27
Fix error range for ETagRequired

### LINKED ISSUE #387: Fix error range for ETagRequired
When multiple child elements are missing their end tags, the error range refers to the parent element's start tag, rather than the child element's.

This error does not happen when only one child is missing their end tag.

Example XML file:
```
<picture>
    <p1></p1>
    <p2>    <!-- missing end tag -->
    <p3>    <!-- missing end tag -->
</picture>
```

Result:
image

image

Error message:
`The element type "p3" must be terminated by the matching end-tag "</p3>".`

xorye:
Fixes #387 

image

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/XMLPositionUtility.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/XMLPositionUtility.java
@@ -177,20 +177,32 @@ public class XMLPositionUtility {
 		if (parent == null || !parent.isElement() || ((DOMElement) parent).getTagName() == null) {
 			return null;
 		}
-		if (parent != null) {
-			DOMNode child = findChildNode(childTag, parent.getChildren());
-			if (child != null) {
+
+		DOMNode curr = parent;
+		DOMNode child;
+		while (curr != null) {
+			child = findUnclosedChildNode(childTag, curr.getChildren());
+			if (child == null) {
+				curr = findUnclosedChildNode(curr.getChildren());
+			} else {
 				return createRange(child.getStart() + 1, child.getStart() + 1 + childTag.length(), document);
 			}
-			if(parent.isElement()) {
-				String parentName = ((DOMElement) parent).getTagName();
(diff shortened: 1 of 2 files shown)


fbricon:
Please use self-descriptive commit messages, like `Fix error range for ETagRequired`

xorye:
Ok, I have just changed the commit message.

fbricon:
LGTM but I'll let @NikolasKomonen or @angelozerr check if that's actually correct.

NikolasKomonen:
@xorye Looks good to me

One thing to change, and I know you didn't do this. But could you update the method name of 'findChildNode' https://github.com/angelozerr/lsp4xml/blob/ccc55276ec3505f93a3d654235560f5e3504a03b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/XMLPositionUtility.java#L205

to something like 'findUnclosedChildNodeWithName'

Other than than it works great.

fbricon:
> One thing to change, and I know you didn't do this. But could you update the method name of 'findChildNode'  to something like 'findUnclosedChildNodeWithName'

findUnclosedChildNode will be fine

xorye:
Cool, changing it now.

xorye:
Actually, this function finds a closed child node. Not an unclosed one.

I'll change it to findClosedChildNode

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/XMLPositionUtility.java
```diff
@@ -177,20 +177,32 @@ public static Range selectChildEndTag(String childTag, int offset, DOMDocument d
...
 			}
 		}
 		return null;
 	}
 
-	static DOMNode findChildNode(String childTag, List<DOMNode> children) {
+	static DOMNode findClosedChildNode(String childTag, List<DOMNode> children) {
 		for (DOMNode child : children) {
 			if (child.isElement() && childTag != null && childTag.equals(((DOMElement) child).getTagName())
 					&& !child.isClosed()) {
```
fbricon: this child is unclosed

PR #396, opened 2019-05-27
Fix duplicate completions for svg DTD

### LINKED ISSUE #386: Completion in svg file proposes many duplicates
Editing an svg file like this one causes completion to return many duplicate items:
```xml
<?xml version="1.0" standalone="no" ?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.0//EN" "http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd">
<svg xmlns="http://www.w3.org/2000/svg">
    <animate attributeName="foo">
        <
    </animate>
</svg>

```

Screen Shot 2019-05-22 at 6 41 07 PM

Referenced from: https://github.com/redhat-developer/vscode-xml/issues/141

NikolasKomonen:
Fixes #386

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/dtd/contentmodel/CMDTDDocument.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/dtd/contentmodel/CMDTDDocument.java
@@ -15,8 +15,10 @@ import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.xerces.impl.dtd.DTDGrammar;
 import org.apache.xerces.impl.dtd.XMLDTDLoader;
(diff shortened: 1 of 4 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/dtd/contentmodel/CMDTDDocument.java
```diff
@@ -94,20 +96,22 @@ public void startContentModel(String elementName, Augmentations augs) throws XNI
...
-		hierachiesMap.put(elementName, hierachies);
+		hierarchies = new LinkedHashSet<String>();
+		hierachiesMap.put(elementName, hierarchies);
 		super.startContentModel(elementName, augs);
 	}
 
 	@Override
 	public void element(String elementName, Augmentations augs) throws XNIException {
-		hierachies.add(elementName);
+		if(!hierarchies.contains(elementName)) {
```
fbricon: check is unnecessary, now that you're using a set

NikolasKomonen:
@fbricon updated, will add issue about startup time problem

Issue #397, opened 2019-05-27
Startup time to do completion using SVG DTD file is slow


NikolasKomonen:
After server starts up, trying to do completion takes a while at first even with the cached DTD file.

```xml
<?xml version="1.0" standalone="no" ?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.0//EN" "http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd">
<svg xmlns="http://www.w3.org/2000/svg">
    <animate attributeName="foo">
        <
    </animate>
</svg>

```

<img width="929" alt="Screen Shot 2019-05-22 at 6 41 07 PM" src="https://user-images.githubusercontent.com/148698/58213751-48988c00-7cc1-11e9-8112-c3539621bcdd.png">

There is probably a performance issue somewhere.

PR #398, opened 2019-05-28
Use DTD/XML Schema cache manager for completion

angelozerr:
This PR uses like validation the cache manager to load the DTD/XML
Schema content model used for completion based on grammar, hover. It
should fix the performance problem of #397


angelozerr:
> Please add some tests

@fbricon updated

NikolasKomonen:
Works great, the wait time is gone.

@fbricon Ready to merge if you have no issues.

PR #402, opened 2019-05-29
Dispose DOM document to improve memory (see #389)

angelozerr:
This PR adds a dispose method for DOM node to set as null the array list fields and referenced fields (like owner element, owner attribute) to help GC.

Here the result without call of dispose when nasa.xml file is parsed:

image

Here the result with call of dispose when nasa.xml file is parsed:

image

We win around 1GB of used heap.

Please note this PR doesn't consume dispose in the XML language server. It should used as soon as DOM Document must be reparsed (the old document must be disposed). I'm waiting for accept this PR to create a new PR which consumes the dispose and which will fix several parsing of DOM Document when several request as done for the same document version.


fbricon:
> Please note this PR doesn't consume dispose in the XML language server. It should used as soon as DOM Document must be reparsed (the old document must be disposed). I'm waiting for accept this PR to create a new PR which consumes the dispose and which will fix several parsing of DOM Document when several request as done for the same document version.

I don't see why we'd need to do that in a separate PR. I'd rather see it done in this one. It'd be easier to rollback if needed.

angelozerr:
> I don't see why we'd need to do that in a separate PR. I'd rather see it done in this one. It'd be easier to rollback if needed.

It was to facilitate the review and not mix problem with several parsing and dispose in the same PR. I will add update the PR as soon as I will have something which will work.

angelozerr:
Completion problem should be fixed. Let me time to fix problem with several parsing for several request but for the same document version.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/Disposable.java
```diff
@@ -0,0 +1,15 @@
+package org.eclipse.lsp4xml.commons;
```
fbricon: missing copyright header

PR #404, opened 2019-05-29
Indicate source of element for Completion.

### LINKED ISSUE #210: Mark element coming from XML Schema/DTD for completion
While studying bug with duplicate elements https://github.com/redhat-developer/vscode-xml/issues/78#issuecomment-437049479 I find it should be good to mark the completion elements origin (from XML Schema, from DTD, copy sibling)

A quick idea is to add -  XML Schema:

image

@fbricon @NikolasKomonen @ralfhandl what do you think about that?

NikolasKomonen:
Fixes #210


fbricon:
https://github.com/angelozerr/lsp4xml/pull/394 was already merged

PR #410, opened 2019-05-31
File completion in attribute value

### LINKED ISSUE #345: Provide file/folder completion for certain xml attributes
When an attribute is named:
- src
- path
- directory
- file
(list could be defined in some preferences), then lsp4xml should propose file completion.
Also, that could be triggered by any attribute value starting with the file:// scheme.

eg (html support in vscode):
Apr-04-2019 16-05-12

NikolasKomonen:
Provides file completion when the user starts typing a path in an attribute value

Fixes #345

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/ParentProcessWatcher.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/ParentProcessWatcher.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package org.eclipse.lsp4xml.commons;
 
+import static org.eclipse.lsp4xml.utils.OSUtils.isWindows;
+
 import java.io.IOException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -19,16 +21,17 @@ import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import com.google.common.io.Closeables;
+
 import org.eclipse.lsp4j.jsonrpc.MessageConsumer;
 import org.eclipse.lsp4j.services.LanguageServer;
 
-import com.google.common.io.Closeables;
-
 /**
  * Watches the parent process PID and invokes exit if it is no longer available.
- * This implementation waits for periods of inactivity to start querying the PIDs.
+ * This implementation waits for periods of inactivity to start querying the
+ * PIDs.
  */
-public final class ParentProcessWatcher implements Runnable, Function<MessageConsumer, MessageConsumer>{
+public final class ParentProcessWatcher implements Runnable, Function<MessageConsumer, MessageConsumer> {
 
 	private static final Logger LOGGER = Logger.getLogger(ParentProcessWatcher.class.getName());
 	private static final boolean isJava1x = System.getProperty("java.version").startsWith("1.");
@@ -38,9 +41,7 @@ public final class ParentProcessWatcher implements Runnable, Function<MessageCon
 	 */
 	private static final int FORCED_EXIT_CODE = 1;
 
-	private static final boolean isWindows = System.getProperty("os.name").toLowerCase().indexOf("win") >= 0;
-	
-	private static final long INACTIVITY_DELAY_SECS = 30 *1000;
+	private static final long INACTIVITY_DELAY_SECS = 30 * 1000;
 	private static final int POLL_DELAY_SECS = 10;
 	private volatile long lastActivityTime;
 	private final ProcessLanguageServer server;
@@ -50,14 +51,14 @@ public final class ParentProcessWatcher implements Runnable, Function<MessageCon
 	public interface ProcessLanguageServer extends LanguageServer {
 
 		long getParentProcessId();
-		
+
 		void exit(int exitCode);
 	}
-	
-	public ParentProcessWatcher(ProcessLanguageServer server ) {
+
+	public ParentProcessWatcher(ProcessLanguageServer server) {
 		this.server = server;
 		service = Executors.newScheduledThreadPool(1);
-		task =  service.scheduleWithFixedDelay(this, POLL_DELAY_SECS, POLL_DELAY_SECS, TimeUnit.SECONDS);
+		task = service.scheduleWithFixedDelay(this, POLL_DELAY_SECS, POLL_DELAY_SECS, TimeUnit.SECONDS);
 	}
 
 	@Override
@@ -70,8 +71,8 @@ public final class ParentProcessWatcher implements Runnable, Function<MessageCon
 	}
 
 	/**
-	 * Checks whether the parent process is still running.
-	 * If not, then we assume it has crashed, and we have to terminate the Java Language Server.
+	 * Checks whether the parent process is still running. If not, then we assume it
+	 * has crashed, and we have to terminate the Java Language Server.
 	 *
 	 * @return true if the parent process is still running
 	 */
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/Constants.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/Constants.java
@@ -70,7 +70,7 @@ public class Constants {
 
 	public static final Pattern PI_TAG_NAME = Pattern.compile("^[a-zA-Z0-9]+");
 
-	//Add comming processing instructions that are defined to have attributes as content
+	//Add coming processing instructions that are defined to have attributes as content
 	public static final Pattern PI_WITH_VARIABLES = Pattern.compile("^(xml-stylesheet)[\\s<>?]?");
 
 	public static final Pattern DOCTYPE_KIND_OPTIONS = Pattern.compile("^(PUBLIC|SYSTEM)([\\s<>\"'])");
@@ -91,4 +91,5 @@ public class Constants {
 
 	public static final Pattern DOCTYPE_NAME =
 	Pattern.compile("^[_:\\w][_:\\w-.\\d]*");
+
 }
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
@@ -1,0 +1,241 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+
+package org.eclipse.lsp4xml.extensions.general.completion;
+
+import static org.eclipse.lsp4xml.utils.FilesUtils.convertToWindowsPath;
+import static org.eclipse.lsp4xml.utils.FilesUtils.getFilePathSlash;
+import static org.eclipse.lsp4xml.utils.FilesUtils.getNormalizedPath;
+import static org.eclipse.lsp4xml.utils.OSUtils.isWindows;
+import static org.eclipse.lsp4xml.utils.StringUtils.isEmpty;
+
+import java.io.File;
+import java.io.FilenameFilter;
+import java.net.URI;
+import java.nio.file.Path;
+
+import org.eclipse.lsp4j.CompletionItem;
+import org.eclipse.lsp4j.CompletionItemKind;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.TextEdit;
+import org.eclipse.lsp4xml.commons.BadLocationException;
+import org.eclipse.lsp4xml.dom.DOMDocument;
+import org.eclipse.lsp4xml.services.extensions.CompletionParticipantAdapter;
+import org.eclipse.lsp4xml.services.extensions.ICompletionRequest;
+import org.eclipse.lsp4xml.services.extensions.ICompletionResponse;
+import org.eclipse.lsp4xml.settings.SharedSettings;
+import org.eclipse.lsp4xml.utils.CompletionSortTextHelper;
+import org.eclipse.lsp4xml.utils.FilesUtils;
+import org.eclipse.lsp4xml.utils.StringUtils;
+
+/**
+ * FilePathCompletionParticipant
+ */
+public class FilePathCompletionParticipant extends CompletionParticipantAdapter {
+
+	public static final String FILE_SCHEME = "file";
+
+	@Override
+	public void onAttributeValue(String valuePrefix, Range fullRange, boolean addQuotes, ICompletionRequest request,
+			ICompletionResponse response, SharedSettings settings) throws Exception {
+
+		DOMDocument xmlDocument = request.getXMLDocument();
+		String text = xmlDocument.getText();
+
+		// Get full attribute value range
+		// + 1 since it includes the quotations
+		int documentStartOffset = xmlDocument.offsetAt(fullRange.getStart()) + 1;
+		
+		String fullAttributeValue = valuePrefix;
+		if (isEmpty(fullAttributeValue)) {
+			return;
+		}
+
+		// Get value and range from fullAttributeValue
+		int completionOffset = request.getOffset(); // offset after the typed character
+		int parsedAttributeStartOffset = StringUtils.getOffsetAfterWhitespace(fullAttributeValue, completionOffset - documentStartOffset) + documentStartOffset; // first character of URI
+		String attributePath = text.substring(parsedAttributeStartOffset, completionOffset);
+	
+		Position startValue = xmlDocument.positionAt(parsedAttributeStartOffset);
+		Position endValue = xmlDocument.positionAt(completionOffset);
+		fullRange = new Range(startValue, endValue);
+
+		// Try to get the URI string from the attribute value in case it has a file scheme
+		// header (eg: "file://")
+		String osSpecificAttributePath = attributePath;
+		boolean hasFileScheme = false;
+
+		hasFileScheme = attributePath.startsWith(FilesUtils.FILE_SCHEME);
+		if (hasFileScheme) {
+			osSpecificAttributePath = attributePath.substring(FilesUtils.FILE_SCHEME.length());
+		}
+
+		String slashInAttribute = getFilePathSlash(attributePath);
+
+		if (hasFileScheme) {
+			if (!osSpecificAttributePath.startsWith("/")) {
+				return; // use of 'file://' and the path was not absolute
+			}
+			if (isWindows && osSpecificAttributePath.length() == 1) { // only '/', so list Windows Drives
+				
+				Range replaceRange = adjustReplaceRange(xmlDocument, fullRange, attributePath, "/");
+
+				File[] drives = File.listRoots();
+				for (File drive : drives) {
+					createFilePathCompletionItem(drive, replaceRange, response, "/");
+				}
+				return;
+			}
+		}
+		
+		if(isWindows) {
+			osSpecificAttributePath = convertToWindowsPath(osSpecificAttributePath);
+		}
+		else if("\\".equals(slashInAttribute)) { // Backslash used in Unix
+			osSpecificAttributePath = osSpecificAttributePath.replace("\\", "/");
+		}
+		
+		// Get the normalized URI string from the parent directory file if necessary
+		String workingDirectory = null; // The OS specific path for a working directory
+
+		if (!hasFileScheme) { //The path from the attribute value is not a uri, so we might need to reference the working directory path
+			String uriString = xmlDocument.getTextDocument().getUri();
+			URI uri = new URI(uriString);
+
+			if(!FILE_SCHEME.equals(uri.getScheme())) {
+				return;
+			}
+
+			String uriPathString = uri.getPath();
+			if(!uriPathString.startsWith("/")) {
+				return; //file uri is incorrect
+			}
+			int lastSlash = uriPathString.lastIndexOf("/");
+			if(lastSlash > -1) {
+				workingDirectory = uriPathString.substring(0, lastSlash);
+
+				if(isWindows) {
+					// Necessary, so that this path is readable in Windows
+					workingDirectory = convertToWindowsPath(workingDirectory);
+				}
+			}
+		}
+		
+		//Try to get a correctly formatted path from the given values
+		Path validAttributeValuePath = getNormalizedPath(workingDirectory, osSpecificAttributePath); 
+		
+		if(validAttributeValuePath == null) {
+			return;
+		}
+
+		//Get adjusted range for the completion item (insert at end, or overwrite some existing text in the path)
+		Range replaceRange = adjustReplaceRange(xmlDocument, fullRange, attributePath, slashInAttribute);
+
+		createNextValidCompletionPaths(validAttributeValuePath, slashInAttribute, replaceRange, response, null);
+	}
+
+	/**
+	 * Returns a Range that covers trailing content after a slash, or 
+	 * if it already ends with a slash then a Range right after it.
+	 * @param xmlDocument
+	 * @param fullRange
+	 * @param attributeValue
+	 * @param slash
+	 * @return
+	 */
+	private Range adjustReplaceRange(DOMDocument xmlDocument, Range fullRange, String attributeValue, String slash) {
+		//In the case the currently typed file/directory needs to be overwritten
+		Position replaceStart = null;
+		Position currentEnd = fullRange.getEnd();
+		
+		int startOffset;
+		try {
+			startOffset = xmlDocument.offsetAt(fullRange.getStart());
+		} catch (BadLocationException e) {
+			return null;
+		}
+		int lastSlashIndex = attributeValue.lastIndexOf(slash);
+		if(lastSlashIndex > -1) {
+			try {
+				replaceStart = xmlDocument.positionAt(startOffset + lastSlashIndex);
+			} catch (BadLocationException e) {
+				return null;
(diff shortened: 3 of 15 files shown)


fbricon:
Can we get it to work for "xsi:schemaLocation='http://invoice  file:///Users/username/path/to/invoice.xsd"?

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,189 @@
...
+public class FilePathCompletionParticipant implements ICompletionParticipant {
+
+	private static final FilenameFilter xsdFilter = new FilenameFilter() {
+
+		@Override
+		public boolean accept(File dir, String name) {
+			if(name.matches(".+\\.[^\\s]+$") == false) {
+				return true;
+			}
+			return name.toLowerCase().endsWith(".xsd");
```
fbricon: file filter should be for specific attributes, e.g: ` <?xml-stylesheet href="./my-style.css" encoding="UTF-8" type="text/css"?>`

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,189 @@
...
+			return f.listFiles(filter);
+		}
+		return null;
+	}
+
+	private void createFilePathCompletionItem(Path pathOfDirectory, String attributePath, File f, Range replaceRange, ICompletionResponse response, String slash) {
+		CompletionItem item = new CompletionItem();
+		String fName = f.getName();
+		String insertText;
+		// if(f.isDirectory()) {
```
NikolasKomonen: Remove Comments

NikolasKomonen:
@fbricon If I am in an empty attribute value between quotes, and I do ctrl+space should completion be done for all files/folders adjacent to the xml file or only when the user starts off with a `./` or `/`?

fbricon:
for now let's keep . and / (file:// too) as triggers.
We can iterate later on a better behavior. (Automatically enable for attributes names like path, file, folder, href...make that list configurable).
We also need to be able to disable file/path completion through preferences

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/ParentProcessWatcher.java
```diff
@@ -38,7 +38,7 @@
 	 */
 	private static final int FORCED_EXIT_CODE = 1;
 
-	private static final boolean isWindows = System.getProperty("os.name").toLowerCase().indexOf("win") >= 0;
+	public static final boolean isWindows = System.getProperty("os.name").toLowerCase().indexOf("win") >= 0;
```
angelozerr: Should be part in a new  OSUtils class

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,235 @@
...
+import org.eclipse.lsp4xml.services.extensions.ICompletionRequest;
+import org.eclipse.lsp4xml.services.extensions.ICompletionResponse;
+import org.eclipse.lsp4xml.settings.SharedSettings;
+import org.eclipse.lsp4xml.utils.CompletionSortTextHelper;
+import org.eclipse.lsp4xml.utils.StringUtils;
+
+/**
+ * FilePathCompletionParticipant
+ */
+public class FilePathCompletionParticipant implements ICompletionParticipant {
```
angelozerr: I suggest you that you extends CompletionParticipantAdapter to avoid having complation problem in the future when we will add new method on ICompletionParticipant

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,235 @@
...
+
+	public static final String FILE_SCHEME = "file";
+
+	private static final FilenameFilter xsdFilter = new FilenameFilter() {
+
+		@Override
+		public boolean accept(File dir, String name) {
+			if(name.matches(".+\\.[^\\s]+$") == false) {
```
angelozerr: It should be more performant to compile the pattern in Pattern class (like XMLScanner)

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,235 @@
...
+				return true;
+			}
+			return name.toLowerCase().endsWith(".xsd");
```
angelozerr: Use DOMUtils#isXsd

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,235 @@
...
+			ICompletionResponse response, SharedSettings settings) throws Exception {
+
+		DOMDocument xmlDocument = request.getXMLDocument();
+		String text = xmlDocument.getText();
+
+		//Get full attribute value range
+		// +- 1 since it includes the quotations
+		int documentStartOffset = xmlDocument.offsetAt(fullRange.getStart()) + 1;
+		int documentEndOffset = xmlDocument.offsetAt(fullRange.getEnd()) - 1;
+		String fullAttributeValue = text.substring(documentStartOffset, documentEndOffset);
```
angelozerr: Why not using directly valuePrefix ?

angelozerr:
@NikolasKomonen several tests failed on Windows OS. I will investigate that tomorrow and give you feedback. But I think we should simplify code to avoid get Path from String and after convert it to URI and after a file, etc. I think we should use Path only and their API like normalize, abolutize, etc. It's just my first impression, perhaps I'm wrong.

NikolasKomonen:
@angelozerr I've fixed all the issues you've mentioned. When you have time could you test the pr.

angelozerr:
yes sure! I will test tomorrow

angelozerr:
@NikolasKomonen tests are passing now on Windows OS

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,265 @@
...
+		DOMDocument xmlDocument = request.getXMLDocument();
+		String text = xmlDocument.getText();
+
+		// Get full attribute value range
+		// + 1 since it includes the quotations
+		int documentStartOffset = xmlDocument.offsetAt(fullRange.getStart()) + 1;
+		
+		String fullAttributeValue = valuePrefix;
+		if (isEmpty(fullAttributeValue)) {
+			return;
```
angelozerr: Coudn't consider that if value is empty, it's the same thing than '.' ?

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,265 @@
...
+				File[] drives = File.listRoots();
+				for (File drive : drives) {
+					createFilePathCompletionItem(Paths.get("/"), drive, replaceRange, response, "/");
+				}
+				return;
+			}
+		}
+		else {
+			if(isWindows && "/".equals(slash)) { // non-URI using forward slash in Windows
+				return;
```
angelozerr: The '/' using must work with Windows OS too because we can use it in noSchemaLocation. In other words filr path completion must have the same behaviour for any OS.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,265 @@
...
+		
+		//Try to get an correctly formatted path from the given values
+		Path validAttributeValuePath = getNormalizedPath(directoryOfFile, pathInAttribute); 
+		
+		if(validAttributeValuePath == null) {
+			return;
+		}
+
+		//Get adjusted range for the completion item (insert at end, or overwrite some existing text in the path)
+		Range replaceRange = adjustReplaceRange(xmlDocument, fullRange, subAttributeValue, slash);
```
angelozerr: I think replace range should be fullRange every time to update the whole attribute value (like completion based on XML Schema).

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
NikolasKomonen: Do we want to have path completion for every attribute all the time?

I thought it would be better to trigger when they type '.'

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
NikolasKomonen: I think it makes it harder, if you have 
```xml
<a k="AAA BBB ./fileA/| CCC"> </a>
```
If you try to replace the whole attribute then you have to parse it anyways to find out what values you don't touch ("AAA BBB ", " CCC")

angelozerr: I understand but is it the commons case? If i take noschemanamzspace location attribute i prefer it replaces the attribute value. What do you think Fred?

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
angelozerr: No but it works like this today no?

angelozerr: @NikolasKomonen here a commons case that it doesn't work with XML Schema " `xsi:noNamespaceSchemaLocation` (see `|` where completion is triggered:

If I use `.` in the path:

```xml
<data xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="./|">
```
I have completion for xsd files.

image

If I have an empty path:

```xml
<data xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="|">
```
it doesn't work. Generally we write xsd file references without `./`

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/Constants.java
```diff
@@ -91,4 +95,15 @@
 
 	public static final Pattern DOCTYPE_NAME =
 	Pattern.compile("^[_:\\w][_:\\w-.\\d]*");
+
+	private static final FilenameFilter xsdFilter = new FilenameFilter() {
```
angelozerr: Please remove this code which not used for the moment.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,260 @@
...
+
+/**
+ * FilePathCompletionParticipant
+ */
+public class FilePathCompletionParticipant extends CompletionParticipantAdapter {
+
+	public static final String FILE_SCHEME = "file";
+
+
+	@Override
```
angelozerr: Please remove this method which does nothing

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,260 @@
...
+	public void onTagOpen(ICompletionRequest completionRequest, ICompletionResponse completionResponse) throws Exception {
```
angelozerr: Please remove this method which does nothing

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,260 @@
...
+
+	}
+
+	@Override
+	public void onXMLContent(ICompletionRequest request, ICompletionResponse response) throws Exception {
```
angelozerr: Please remove this method which does nothing

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,260 @@
...
+
+	}
+
+	@Override
+	public void onAttributeName(boolean generateValue, Range fullRange, ICompletionRequest request,
```
angelozerr: Please remove this method which does nothing

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -145,4 +152,153 @@ static String toString(InputStream is) {
...
+				return p;
+			}
+			return null;
+		} catch (Exception e) {
+			return null;
+		}
+
+	}
+
+	public static String getFilePathSlash(String text) {
```
angelozerr: Please add javadoc and tests for that

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -145,4 +152,153 @@ static String toString(InputStream is) {
...
+		if (text.contains("\\")) {
+			return "\\";
+		} else if (text.contains("/")) {
+			return "/";
+		} else {
+			return SLASH;
```
angelozerr: I think it should be better to return every time '/' even in Windows OS because for noSchemaNamespaceLocation we use '/' and not '\'. The method could be simplified with 

```java
public static String getFilePathSlash(String text) {
	if (text.contains("\\")) {
		return "\\";
	}
	return "/";
}
```

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -145,4 +152,153 @@ static String toString(InputStream is) {
...
+		}
+	}
+
+	/**
+	 * Ensures there is no slash before a drive letter, and
+	 * forces use of '\'
+	 * @param pathString
+	 * @return
+	 */
+	public static String cleanPathForWindows(String pathString) {
```
angelozerr: Please add test in FilesUtilstest

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -145,4 +152,153 @@ static String toString(InputStream is) {
...
+		}
+
+		int uriLength = uri.length();
+		if (offsetAfterScheme > -1 && uriLength > offsetAfterScheme) {
+			return uri.substring(offsetAfterScheme, uriLength);
+		}
+		return null;
+	}
+
+	public static boolean hasScheme(String uri) {
```
angelozerr: Please remove this not used code

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -145,4 +152,153 @@ static String toString(InputStream is) {
...
+		return getOffsetAfterScheme(uri) > -1;
+	}
+
+	public static boolean hasFileScheme(String uri) {
```
angelozerr: Please remove this unused method

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -145,4 +152,153 @@ static String toString(InputStream is) {
...
+		int offsetAfterScheme = getOffsetAfterScheme(uri);
+
+		if (offsetAfterScheme > -1) {
+			return FILE_SCHEME.equals(uri.substring(0, offsetAfterScheme));
+		}
+
+		return false;
+	}
+
+	public static int getOffsetAfterScheme(String uri) {
```
angelozerr: Please remove this unused method

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -145,4 +152,153 @@ static String toString(InputStream is) {
 			return s.hasNext() ? s.next() : "";
 		}
 	}
+
+	public static String getPathFromURI(String uri) {
```
angelozerr: Please remove this unused method

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -29,18 +33,21 @@
  */
 public class FilesUtils {
 
+	public static final String FILE_SCHEME = "file://";
 	public static final String LSP4XML_WORKDIR_KEY = "lsp4xml.workdir";
 	private static String cachePathSetting = null;
 
+	private static Pattern uriSchemePattern = Pattern.compile("^([a-zA-Z\\-]+:\\/\\/).*");
```
angelozerr: Please remove this unused method

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -29,18 +33,21 @@
  */
 public class FilesUtils {
 
+	public static final String FILE_SCHEME = "file://";
 	public static final String LSP4XML_WORKDIR_KEY = "lsp4xml.workdir";
 	private static String cachePathSetting = null;
 
+	private static Pattern uriSchemePattern = Pattern.compile("^([a-zA-Z\\-]+:\\/\\/).*");
+	private static Pattern endFilePattern = Pattern.compile(".*[\\\\\\/]\\.[\\S]+");
```
angelozerr: Please remove this unused field

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -145,4 +152,153 @@ static String toString(InputStream is) {
...
+					pathString = pathString.substring(1);
+				}
+			}
+			
+		}
+		pathString = pathString.replace("/", "\\");
+		return pathString;
+	}
+
+	public static boolean pathEndsWithFile(String pathString) {
```
angelozerr: Please remove this unused method

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/StringUtils.java
```diff
@@ -216,17 +217,66 @@ public static int getNumberOfNewLines(String text, boolean isWhitespace, String
...
+	 * The offset of the character after the whitespace is returned.
+	 * 
+	 * (text = "abcd efg|h", endOffset = 8) -> 5
+	 * 
+	 * 
+	 * @param text
+	 * @param endOffset non-inclusive
+	 * @return Start offset directly after the first whitespace.
+	 */
+	public static int getOffsetAfterWhitespace(String text, int endOffset) {
```
angelozerr: Please add test in StringUtilsTest

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -145,4 +152,153 @@ static String toString(InputStream is) {
...
+			if(pathString.length() > 3) {
+				char letter = pathString.charAt(1);
+				char colon = pathString.charAt(2);
+				if(Character.isLetter(letter) && ':' == colon) {
+					pathString = pathString.substring(1);
+				}
+			}
+			
+		}
+		pathString = pathString.replace("/", "\\");
```
angelozerr: return pathString.replace("/", "\\");

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/CompletionSortTextHelper.java
```diff
@@ -0,0 +1,56 @@
...
+package org.eclipse.lsp4xml.utils;
+
+import org.eclipse.lsp4j.CompletionItemKind;
+
+/**
+ * CompletionSortText class to get sortText for CompletionItem's
+ */
+public class CompletionSortTextHelper {
+
+	private int i;
```
angelozerr: Please remove all code which is not used

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/CompletionSortTextHelper.java
```diff
@@ -0,0 +1,56 @@
...
+		return base + Integer.toString(i);
+	}
+
+	public String next(CompletionItemKind kind) {
+		String tempBase = getSortText(kind);
+		i++;
+		return tempBase + Integer.toString(i);
+	}
+
+	public static String getSortText(CompletionItemKind kind) {
```
angelozerr: Please add test

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/FilesUtils.java
```diff
@@ -145,4 +152,153 @@ static String toString(InputStream is) {
...
+		combinedPath = parentDirectory + SLASH + givenPathCleaned;
+		p = getPathIfExists(combinedPath);
+		if (p != null) {
+			return p;
+		}
+
+		return null;
+	}
+
+	private static Path getPathIfExists(String path) {
```
angelozerr: Simplify code like this:

private static Path getPathIfExists(String path) {
		try {
			Path p = Paths.get(path).normalize();
			return p.toFile().exists() ? p : null;
		} catch (Exception e) {
			return null;
		}
	}

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -0,0 +1,260 @@
...
+		}
+	}
+
+	/**
+	 * Returns a list of File objects that are in the given directory
+	 * @param pathOfDirectory
+	 * @param filter
+	 * @return
+	 */
+	private File[] gatherFiles(Path pathOfDirectory, FilenameFilter filter) {
```
angelozerr: Simplify code like this:

```java
private File[] gatherFiles(Path pathOfDirectory, FilenameFilter filter) {
		File f = pathOfDirectory.toFile();
		return f.isDirectory() ? f.listFiles(filter) : null; 
	}
```

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
angelozerr: @NikolasKomonen I understand more your strategy now. We could see that in an another issue.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/general/completion/FilePathCompletionParticipant.java
angelozerr: Forget my comments

angelozerr:
@NikolasKomonen I test your PR and it works like a charm in Windows OS. I add some comments just to clean some pieces of code.

The only thing that I find it's annoying is that it doesn't work if you don't write `.` or `/` in the attribute.

IMHO I think it should work, see my comment at https://github.com/angelozerr/lsp4xml/pull/410/#issuecomment-498399361

But for @fbricon it seems not important https://github.com/angelozerr/lsp4xml/pull/410/

Once you will clean code, I think we could merge it this PR.

angelozerr:
For me PR can be merged. I let @fbricon to test it again and merge it.

fbricon:
When invoking code completion in the middle of the value, a relative path is computed even though the path starts with `/`:
Jun-10-2019 16-10-47

PR #412, opened 2019-06-04
Cancel support.

### LINKED ISSUE #409: Cancel support
I think it's an hard task but to improve performance we should try to implement https://microsoft.github.io/language-server-protocol/specification#cancelRequest

The basic idea is to cancel:

 * LSP requests which are executed on each change of text editor like symbol provider, highligtings, hover, closeTag, etc
 * parse of DOM Document

This cancel support will improve performance and memory too. i'm working on this issue.

angelozerr:
The basic idea is to cancel:

 - LSP requests which are executed on each change of text editor like
symbol provider, highligtings, hover, closeTag, etc
 - parse of DOM Document

This cancel support will improve performance and memory too.

Fixes #409

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/LanguageModelCache.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/LanguageModelCache.java
@@ -13,27 +13,29 @@ package org.eclipse.lsp4xml.commons;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
-import java.util.function.Function;
+import java.util.concurrent.CancellationException;
+import java.util.function.BiFunction;
 
 import org.eclipse.lsp4j.TextDocumentItem;
+import org.eclipse.lsp4j.jsonrpc.CancelChecker;
 
 /**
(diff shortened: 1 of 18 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/commons/LanguageModelCache.java
```diff
@@ -43,10 +45,35 @@ public LanguageModeInfo(T languageModel, int version, String languageId, long cT
...
+		 * @param languageId the language id to check
+		 * @return true if the given version and language id checks the version and
+		 *         language id from the last cached model and false otherwise.
+		 */
+		public boolean check(int version, String languageId) {
+			if (this.version > version) {
+				// Stop the process because the given version is more old than the cached model
+				// version
+				// This case comes from when there are an old request which comes after a new
+				// request.
```
fbricon: I'm trying with nasa.xml, but never seem to be able to reach that part.

angelozerr:
Close in favor to https://github.com/angelozerr/lsp4xml/pull/418

@fbricon I fixed in this new PR your comments

PR #418, opened 2019-06-06
Cancel support to fix #409

### LINKED ISSUE #409: Cancel support

angelozerr:
The basic idea is to cancel:

LSP requests which are executed on each change of text editor like
symbol provider, highligtings, hover, closeTag, etc
parse of DOM Document
This cancel support will improve performance and memory too.

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLTextDocumentService.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLTextDocumentService.java
@@ -21,7 +21,6 @@ import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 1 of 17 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLTextDocumentService.java
```diff
@@ -395,39 +377,30 @@ private void triggerValidationFor(Collection<TextDocument> documents) {
...
-			getXMLLanguageService().publishDiagnostics(xmlDocument,
-					params -> xmlLanguageServer.getLanguageClient().publishDiagnostics(params),
-					(u, v) -> triggerValidation(u, v), monitor, sharedSettings.validationSettings);
-		}
+	private void validate(TextDocument document, CancelChecker cancelChecker) throws CancellationException {
+		cancelChecker.checkCanceled();
+		DOMDocument xmlDocument = getXMLDocument(document);
+		getXMLLanguageService().publishDiagnostics(xmlDocument,
+				params -> xmlLanguageServer.getLanguageClient().publishDiagnostics(params),
+				(doc) -> triggerValidationFor(doc), sharedSettings.validationSettings, cancelChecker);
```
fbricon: here, the client cancel checker is not used by the TextDocumentVersionChecker in triggerValidationFor, it should be wrapped in theory, but since validation is not something controlled by the client today, it's probably fine for now. Still, we should fix it in the next iteration

PR #434, opened 2019-06-12
Element/Attribute names must be case sensitive (see #433)

### LINKED ISSUE #433: Element/Attribute names must be case sensitive
XML Element/Attribute names must be case sensitive. For instance if we have this xml

`<a></A>`

For example if we set the cursor in `<a>`, document highlight should not highlight `</A>`

angelozerr:
This PR fix #433 and avoid using String#toLowerCase() which creates a
lot of instances of String.

### DIFF
--- a/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLHighlightingTest.java
+++ b/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLHighlightingTest.java
@@ -73,9 +73,9 @@ public class XMLHighlightingTest {
 	}
 
 	@Test
-	public void caseInsensivity() throws BadLocationException {
-		assertHighlights("<HTML><diV><Div></dIV></dI|v></html>", new int[] { 7, 24 }, "div");
-		assertHighlights("<HTML><diV|><Div></dIV></dIv></html>", new int[] { 7, 24 }, "div");
+	public void caseSensitive() throws BadLocationException {
+		assertHighlights("<HTML><diV><Div></dIV></dI|v></html>", new int[] { 24 }, "div");
+		assertHighlights("<HTML><diV|><Div></dIV></dIv></html>", new int[] { 7 }, "div");
 	}
 	
 	@Test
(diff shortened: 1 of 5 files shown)


fbricon:
@angelozerr any noticeable performance impact?

angelozerr:
> @angelozerr any noticeable performance impact?

not really -( I just notice that we gain some ms (very few) but editor continue to be slow. But this PR follow the XML spec for case sensitive tag

angelozerr:
@NikolasKomonen @xorye  is there any chance that you review this simple PR please.

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLHighlightingTest.java
```diff
@@ -74,8 +74,8 @@ public void selfclosed() throws BadLocationException {
 
 	@Test
 	public void caseInsensivity() throws BadLocationException {
```
xorye: I think a better name for this test case would be `caseSensitivity` or `caseNoInsensivity`

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLRenameTest.java
```diff
@@ -79,8 +79,8 @@ public void selfclosed() throws BadLocationException {
 
 	@Test
 	public void caseInsensivity() throws BadLocationException {
```
xorye: I think a better name for this test case would be caseSensitivity  or `caseNoInsensivity`

fbricon:
Tests should be renamed *caseSensitive

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLHighlightingTest.java
angelozerr: Good catch! I rename my test with `caseSensitive` according @fbricon suggestion

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/XMLRenameTest.java
angelozerr: Good catch! I rename my test with `caseSensitive` according @fbricon suggestion

angelozerr:
Thanks guys for your review. I merge the PR even if travis failed with https://travis-ci.org/angelozerr/lsp4xml/builds/545792737#L930 (I don't understand why it failed?)

PR #435, opened 2019-06-12
Fix error range for cos-all-limited.2

xorye:
image

In `XSDErrorCode.java`, the arguments that xerces gives back is "unbounded" and "testEle".

Therefore:
`arguments[0] = "unbounded"`, which is the value for the attribute `maxOccurs`
`arguments[1] = "testEle"`, which is the value for attribute `name`
Because of this, I hardcoded `"maxOccurs"` and `"name"` in my code.

My solution works under the assumption that the `name` attribute exists and the cos-all-limited.2 error only occurs with the `maxOccurs` attribute.

This seems to be the case because `maxOccurs` is mentioned from the error message from xerces, and is not included in the arguments, neither is it provided as a parameter for the error message formatter.

I may be wrong and cos-all-limited.2 may occur with attributes other than "maxOccurs". I could not easily find any documentation on this error.

Lastly, xerces provides the location of the parent node of the element with the "unbounded" value. Given the parent, I found the child node with the correct name and got the range for the "unbounded" value for that child.

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/DOMNode.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/DOMNode.java
@@ -15,6 +15,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.function.Function;
 
+import java.util.Objects;
+
 import org.w3c.dom.DOMException;
 import org.w3c.dom.NamedNodeMap;
 import org.w3c.dom.Node;
@@ -376,6 +378,26 @@ public abstract class DOMNode implements Node {
 
 	public List<DOMAttr> getAttributeNodes() {
 		return attributeNodes;
+	}
+
+	/**
+	 * Returns a list of children, each having an attribute called name, with a value
+	 * of value
+	 * @param name name of attribute
+	 * @param value value of attribute
+	 * @return  list of children, each having a specified attribute name and value
+	 */
+	public List<DOMNode> getChildrenWithAttributeValue(String name, String value) {
+		List<DOMNode> result = new ArrayList<>();
+		for (DOMNode child: getChildren()) {
+			if (child.hasAttribute(name)) {
+				String attrValue = child.getAttribute(name);
+				if (Objects.equals(attrValue, value)) {
+					result.add(child);
+				}
+			}
+		}
+		return result;
 	}
 
 	/**
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/participants/XSDErrorCode.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/participants/XSDErrorCode.java
@@ -11,11 +11,13 @@
 package org.eclipse.lsp4xml.extensions.xsd.participants;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.apache.xerces.xni.XMLLocator;
 import org.eclipse.lsp4j.Range;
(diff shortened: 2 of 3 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/DOMNode.java
```diff
@@ -378,6 +378,26 @@ public void setAttributeNode(DOMAttr attr) {
...
+	 * @param name name of attribute
+	 * @param value value of attribute
+	 * @return  list of children, each having a specified attribute name and value
+	 */
+	public List<DOMNode> getChildrenWithAttributeValue(String name, String value) {
+		List<DOMNode> result = new ArrayList<>();
+		for (DOMNode child: getChildren()) {
+			if (child.hasAttribute(name)) {
+				String attrValue = child.getAttribute(name);
+				if (attrValue.equals(value)) {
```
angelozerr: It should be better to do `Objects.equals(attrValue, value)` to prevent from NPE

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/participants/XSDErrorCode.java
```diff
@@ -82,7 +84,18 @@ public static Range toLSPRange(XMLLocator location, XSDErrorCode code, Object[]
 		int offset = location.getCharacterOffset() - 1;
 		// adjust positions
 		switch (code) {
-		case cos_all_limited_2:
+		case cos_all_limited_2: {
+			String nameValue = (String) arguments[1];
+			DOMNode parent = document.findNodeAt(offset);
+			List<DOMNode> children = parent.getChildrenWithAttributeValue("name", nameValue);
+
+			if (children.size() == 0) {
```
angelozerr: children.isEmpty()

angelozerr:
>Therefore:
>arguments[0] = "unbounded", which is the value for the attribute maxOccurs
>arguments[1] = "testEle", which is the value for attribute name

Please note this arguments is used to create xerces message. As this arguments is very helpful to adjust range, the xerces reporter was overrided to retrieve this arguments information.

> Because of this, I hardcoded "maxOccurs" and "name" in my code.

> My solution works under the assumption that the name attribute exists and the cos-all-limited.2 error > only occurs with the maxOccurs attribute.

Very good!

>This seems to be the case because maxOccurs is mentioned from the error message from xerces, and is > not included in the arguments, neither is it provided as a parameter for the error message formatter.

Indeed see xerces sources https://github.com/apache/xerces2-j/blob/trunk/src/org/apache/xerces/impl/xs/traversers/XSDAbstractTraverser.java#L925
I may be wrong and cos-all-limited.2 may occur with attributes other than "maxOccurs". I could not easily find any documentation on this error.

> Lastly, xerces provides the location of the parent node of the element with the "unbounded" value. >Given the parent, I found the child node with the correct name and got the range for the "unbounded" >value for that child.

Once you will update the 2 little changes, I think we can merge it.

xorye:
I have made the changes

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/DOMNode.java
```diff
@@ -15,6 +15,8 @@
 import java.util.List;
 import java.util.function.Function;
 
+import com.google.common.base.Objects;
```
angelozerr: please use java.util.Objects

angelozerr: @xorye please use java.util.Objects

angelozerr:
Thank's @xorye !

PR #440, opened 2019-06-14
Fix error range for p-props-correct.2.1

### LINKED ISSUE #436: Error range for p-props-correct.2.1
if minOccurs > maxOccurs in xs:element, xerces report a `p-props-correct.2.1` and the range is not managed:

image

Here the XSD sample:

```xml
<xs:schema 
	xmlns:xs="http://www.w3.org/2001/XMLSchema">
	<xs:complexType name="testType">     
		<xs:all>
			<xs:element name="testEle" minOccurs="1" maxOccurs="0" type="xs:string"/>
		</xs:all>
	</xs:complexType>
</xs:schema>
```

xorye:
Fixes #436 

image

There were three arguments provided by xerces:
`arguments[0] = "element"` (from `<xs:element`)
`arguments[1] = XInt(1)` (value of `minOccurs`)
`arguments[2] = XInt(0)` (value of `maxOccurs`)

The location given by xerces is the location of the element with the error.


angelozerr:
I love this kind of PR which is easy to test, the explanation is very good and commit comment is perfect. Thank's @xorye !

PR #441, opened 2019-06-14
Fix src-element.3 error range

### LINKED ISSUE #420: Fix Error Range for `src-element.3`
Given:
```xml
<a xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
     xsi:noNamespaceSchemaLocation="intType.xsd">
	<b></b>
</a>
```

and the XSD:
```xml
  <xs:element name="a" type="xs:integer">
    <xs:complexType>
      <xs:sequence>
        <xs:element name="b"></xs:element>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
</xs:schema>
```

You get errors on both documents:

The XML:
Selection_048

The XSD:
Selection_047

xorye:
Fixes #420 

Upon looking at how this error is handled in IntelliJ and Eclipse, this is what I found:

IntelliJ
For xsd, there is no error
For xml, there is an error on the `b` tag name

Eclipse
For xsd, the error is on the whole line that contains `<xs:element name="a" type="xs:integer">`
For xml, there is no error.

This fix displays an error for both xsd and xml.

The error ranges look like the following:
xml:
image

xsd:
image

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/XMLSchemaErrorCode.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/XMLSchemaErrorCode.java
@@ -67,7 +67,8 @@ public enum XMLSchemaErrorCode implements IXMLErrorCode {
 	cvc_minInclusive_valid("cvc-minInclusive-valid"), // https://wiki.xmldation.com/Support/validator/cvc-mininclusive-valid
 	TargetNamespace_2("TargetNamespace.2"), 
 	SchemaLocation("SchemaLocation"),
-	schema_reference_4("schema_reference.4"); //
+	schema_reference_4("schema_reference.4"), //
+	src_element_3("src-element.3");
 
 	private final String code;
 
@@ -128,6 +129,7 @@ public enum XMLSchemaErrorCode implements IXMLErrorCode {
 		case cvc_complex_type_2_4_f:
 		case cvc_elt_1_a:
 		case cvc_complex_type_4:
+		case src_element_3:
 		case TargetNamespace_2:
 			return XMLPositionUtility.selectStartTag(offset, document);
 		case cvc_complex_type_3_2_2: {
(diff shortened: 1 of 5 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/XMLSyntaxErrorCode.java
```diff
@@ -47,7 +47,7 @@
 	EqRequiredInAttribute, the_element_type_lmsg("the-element-type-lmsg"), EqRequiredInXMLDecl, IllegalQName,
 	InvalidCommentStart, LessthanInAttValue, MarkupEntityMismatch, MarkupNotRecognizedInContent,
 	NameRequiredInReference, OpenQuoteExpected, PITargetRequired, PseudoAttrNameExpected, QuoteRequiredInXMLDecl,
-	SDDeclInvalid, SpaceRequiredBeforeEncodingInXMLDecl, SpaceRequiredBeforeStandalone, SpaceRequiredInPI,
+	SDDeclInvalid, src_element_3("src-element.3"), SpaceRequiredBeforeEncodingInXMLDecl, SpaceRequiredBeforeStandalone, SpaceRequiredInPI,
```
angelozerr: For me it's not a syntax error but an error coming from XML Schema. In other words please remove src_element_3 in XMLSchemaErrorCode (error in XML file coming from a XML Schema)

xorye: Did you mean remove src_element_3 in XMLSyntaxErrorCode instead?

angelozerr: yes sorry and move to XMLSchemaErrorCode

xorye: Awesome. I also moved my test case from XMLSyntaxDiagnosticsTest to XMLSyntaxDiagnosticsTest

angelozerr:
Thanks @xorye for your fix, it's perfect.

PR #443, opened 2019-06-15
Improve performance and memory with large file.

angelozerr:
This 'draft' PR (please don't merge it and some tests which must be fixed) is the mix of the 3 issues:

 - incremental support
 - tree line tracker
 - load DOM document in completable future

 to see in action the improve of memory and performance with large file.
The performance is not improved when you type one character but when you
type several character.


angelozerr:
I close this draft PR since all my work was merged.

PR #445, opened 2019-06-17
Improve XMLScanner performance

### LINKED ISSUE #444: Improve XML Scanner performance
In XML Scanner, it seems that consumed CPU time is spent in 3 regex to extract element name, attribute name and attribute value. This issue is to change just thoses regexp into a Java code to improve performance. 

After testing that, a large file like nasa.xml is parsed 2-3 times faster

angelozerr:
Fix #444

This PR improve XMLScanner performance by replacing regex with java code
for the 3 regexp which are the most used (element name, attribute name,
attribute value).

After testing that, a large file like nasa.xml is parsed 2-3 times
faster. You can see this time when you start XMLScannerPerformance and
DOMParserPerformance.

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/Constants.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/parser/Constants.java
@@ -16,6 +16,9 @@ public class Constants {
 
 	public final static int _EXL = "!".codePointAt(0);
 	public final static int _MIN = "-".codePointAt(0);
+	public final static int _UDS = "_".codePointAt(0);
+	public final static int _DDT = ":".codePointAt(0);
+	public final static int _DOT = ".".codePointAt(0);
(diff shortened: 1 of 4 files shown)


fbricon:
Can you check if using switch statements make any difference?

angelozerr:
> Can you check if using switch statements make any difference?

I will check that ASAP, but if I remember while is more performant than if/else. In our case we have just if (with or). is it the same thing?

angelozerr:
> Can you check if using switch statements make any difference?

I'm sorry @fbricon but I'm not sure that I can use while since I'm using constants for int (code point). Have you some suggestion of code please?

fbricon:
Yeah I saw the issue about the compiler complaining about the constants not being constants (!?!?!?).  You could either use actual int values instead of computing them or just ignore switch statements. In any case it shouldn't make a significant difference

angelozerr:
@fbricon perhaps we could see that in an another issue?

angelozerr:
Thanks @fbricon for your review

PR #446, opened 2019-06-17
XSD 1.1 working, but fails tests cases

NikolasKomonen:


angelozerr:
I close this PR which is replaced with  https://github.com/angelozerr/lsp4xml/pull/515

PR #447, opened 2019-06-17
Returns markdown formatted hovers.

### LINKED ISSUE #245: Return hover as Markdown

NikolasKomonen:
Fixed issue with confusion between html and already markdown documentation. Wrote unit tests

Fixes #245

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/MarkdownConverter.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/MarkdownConverter.java
@@ -1,0 +1,80 @@
+/*******************************************************************************
+ * Copyright (c) 2016-2017 Red Hat Inc. and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.lsp4xml.utils;
+
+import static org.apache.commons.lang3.StringEscapeUtils.unescapeJava;
+import static org.apache.commons.lang3.StringEscapeUtils.unescapeXml;
+
+import java.lang.reflect.Field;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+
+import com.overzealous.remark.Options;
+import com.overzealous.remark.Options.FencedCodeBlocks;
+import com.overzealous.remark.Options.Tables;
+import com.overzealous.remark.Remark;
+
+import org.jsoup.safety.Cleaner;
+import org.jsoup.safety.Whitelist;
+
+/**
+ * Converts HTML content into Markdown equivalent.
+ *
+ * @author Fred Bricon
+ */
+public class MarkdownConverter {
+
+	private static final Logger LOGGER = Logger.getLogger(MarkdownConverter.class.getName());
+
+	private static Remark remark;
+
+	//Pattern looking for any form of tag eg: <head>
+	private static final Pattern markdownPattern = Pattern.compile("`[^`]*<[a-z][\\s\\S]*>[^`]*`");
+
+	private MarkdownConverter(){
+		//no public instanciation
+	}
+
+	static {
+		Options options = new Options();
+		options.tables = Tables.CONVERT_TO_CODE_BLOCK;
+		options.hardwraps = true;
+		options.inlineLinks = true;
+		options.autoLinks = true;
+		options.reverseHtmlSmartPunctuation = true;
+		options.fencedCodeBlocks = FencedCodeBlocks.ENABLED_BACKTICK;
+		remark = new Remark(options);
+		//Stop remark from stripping file protocol in an href
+		try {
+			Field cleanerField = Remark.class.getDeclaredField("cleaner");
+			cleanerField.setAccessible(true);
+
+			Cleaner c = (Cleaner) cleanerField.get(remark);
+
+			Field whitelistField = Cleaner.class.getDeclaredField("whitelist");
+			whitelistField.setAccessible(true);
+
+			Whitelist w = (Whitelist) whitelistField.get(c);
+
+			w.addProtocols("a", "href", "file");
+		} catch (NoSuchFieldException | SecurityException | IllegalArgumentException | IllegalAccessException e) {
+			LOGGER.severe("Unable to modify jsoup to include file protocols "+ e.getMessage());
+		}
+	}
+
+	public static String convert(String html) {
+		if(!StringUtils.isTagOutsideOfBackticks(html)) {
+			return unescapeXml(html); // is not html so it can be returned as is (aside from unescaping)
+		}
+		return unescapeJava(remark.convert(html));
+	}
+
+}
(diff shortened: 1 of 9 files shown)


org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/utils/MarkdownConverterTest.java
```diff
@@ -0,0 +1,36 @@
...
+package org.eclipse.lsp4xml.utils;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+/**
+ * MarkdownConverterTest
+ */
+public class MarkdownConverterTest {
```
angelozerr: Could you write more tests with HTML that you wish to convert because i would like to investigate to use xmlscanner to generate markdown. Thanks!

NikolasKomonen: From what I have seen for XSD's, this seems to cover all the cases I could find.

fbricon: see https://github.com/angelozerr/lsp4xml/issues/245#issuecomment-457214267 and https://github.com/felixfbecker/PowerShellXSD/blob/master/Test.Types.ps1xml

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/MarkdownConverter.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.commons.lang3.StringEscapeUtils;
+import org.jsoup.safety.Cleaner;
+import org.jsoup.safety.Whitelist;
+
+/**
+ * Converts HTML content into Markdown equivalent.
+ *
+ * @author Fred Bricon
+ */
+public class MarkdownConverter {
```
angelozerr: Here a version of markdown converter (written in few minutes, so I think we can improve it again to support other HTML elements) which uses the XMLScanner. All tests are passed with this converter.

```java
/*******************************************************************************
 * Copyright (c) 2016-2017 Red Hat Inc. and others.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/legal/epl-v10.html
 *
 * Contributors:
 *     Red Hat, Inc. - initial API and implementation
 *******************************************************************************/
package org.eclipse.lsp4xml.utils;

import java.util.logging.Logger;
import java.util.regex.Pattern;

import org.apache.commons.lang3.StringEscapeUtils;
import org.eclipse.lsp4xml.dom.parser.Scanner;
import org.eclipse.lsp4xml.dom.parser.TokenType;
import org.eclipse.lsp4xml.dom.parser.XMLScanner;

/**
 * Converts HTML content into Markdown equivalent.
 *
 * @author Fred Bricon
 */
public class MarkdownConverter {

	private static final Logger LOGGER = Logger.getLogger(MarkdownConverter.class.getName());

	// Pattern looking for any form of tag eg: <head>
	private static final Pattern htmlPattern = Pattern.compile("[^`]*<[a-z][\\s\\S]*>[^`]*");

	private MarkdownConverter() {
		// no public instanciation
	}

	public static String convert(String html) {
		if (!htmlPattern.matcher(html).matches()) {
			return StringEscapeUtils.unescapeXml(html); // is not html so it can be returned
		}
		return doConvert(html);
	}

	private static String doConvert(String html) {
		StringBuilder markdown = new StringBuilder();
		Scanner scanner = XMLScanner.createScanner(html);
		TokenType token = scanner.scan();
		while (token != TokenType.EOS) {
			switch (token) {
			case Content:
				markdown.append(scanner.getTokenText().replaceAll("&lt;", "<").replaceAll("&gt;", ">"));
				break;
			case StartTag: {
				String tag = scanner.getTokenText();
				if ("code".equals(tag)) {
					markdown.append("`");
				} else if ("br".equals(tag)) {
					markdown.append("\n");
				} else if ("b".equals(tag) || "strong".equals(tag)) {
					markdown.append("**");
				}
				break;
			}
			case EndTag: {
				String tag = scanner.getTokenText();
				if ("code".equals(tag)) {
					markdown.append("`");
				} else if ("b".equals(tag) || "strong".equals(tag)) {
					markdown.append("**");
				}
				break;
			}
			}
			token = scanner.scan();
		}

		return markdown.toString();
	}

}

```

fbricon:
PR doesn't seem to work if you have a some html tags in the documentation (eg. `<b>bold</b>`

NikolasKomonen:
@fbricon 

Should something like this work?
```xml
<xs:documentation>
    <b>bold</b>
</xs:documentation>
```

I see maven.xsd using something like:
```xml
<xs:documentation>
    &lt;b&gt;bold&lt;/b&gt;
</xs:documentation>
```

fbricon:
can contain "Any well-formed XML content" https://www.w3schools.com/xml/el_documentation.asp

fbricon:
Using `<b>bold</b>` is correctly rendered as bold in Intellij IDEA

NikolasKomonen:
@fbricon This looks like a xerces issue

There is no documentation returned from the call here: https://github.com/angelozerr/lsp4xml/blob/master/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/contentmodel/XSDAnnotationModel.java#L83

Maybe @angelozerr would know if there is a fix?

angelozerr:
> @fbricon This looks like a xerces issue

There is no documentation returned from the call here: https://github.com/angelozerr/lsp4xml/blob/master/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/contentmodel/XSDAnnotationModel.java#L83

Indeed, so it's an issue from lsp4xml.

> Maybe @angelozerr would know if there is a fix?

The XSDAnnotationModel load the <xs;annotation with SAX to ignore XML element like br because the plain text must not display br. Now we need this br to convert it to markdown. I think we could skip XSDAnnotationModel  and use some substring to extract xs:documentation. If you need help, please tell me otherwise I will try to do it.

NikolasKomonen:
@angelozerr I will try to implement the substring solution.

fbricon:
If the documentation contains html and is multiline:
```xml
 <xs:documentation>multiline documentation 
      returns <code>null</code> </xs:documentation>
```
=> returns null

and if the documentation mixes escaped and regular tags, like
```xml
 <xs:documentation>mixing escaped `&lt;tags&gt;` and regular <code>html</code></xs:documentation>
```
Then it returns 
Screen Shot 2019-06-19 at 4 34 43 PM

angelozerr:
Thanks @NikolasKomonen !

PR #449, opened 2019-06-19
Use CompletableFuture to load DOMDocument

### LINKED ISSUE #439: Use CompletableFuture to load DOMDocument
To improve memory and performance with large file we need :

 * the incremental support 
 * load DOM document with CompletableFuture

This issue is to implement the load of DOMDocument with CompletableFuture

angelozerr:
Fix #439

This PR uses  CompletableFuture to load DOMDocument. The basic idea is when a service is called (ex: foldingRanges) it load DOM document if it need and then execute foldingRanges compute.

It's more cleaner than using a synchronized keyword. This PR remove the language model cache (which stored the DOMDocument in a map) and now the ModelTextDocument (which extends TextDocument) provides a getModel which returns the completable future which load the DOM document.

This PR should improve the stop of parse of DOM document which is done when requets is canceled (by client) or when text document version changed.


angelozerr:
> I'm not completely sure about the validity code wise. But the PR works well for me, no issues found from a user level.

Thank's @NikolasKomonen for your feedback. @fbricon I merge this PR, if you don't like it, please remove my PR. Thanks!

PR #456, opened 2019-06-20
Fix error range for duplicate attribute

xorye:
image

The error range for both `AttributeNSNotUnique` and `AttributeNotUnique` is now on the first occurrence of the duplicate attribute. If needed, it could be on the last occurrence of the attribute, but we would need a method like `DOMNode.getAttributeNode(String prefix, String suffix)`, except it iterates through the list of attributes, backwards.

For the above screenshot, the arguments were:
```
arguments[0] = "xs:schema"
arguments[1] = "tns"
arguments[2] = "http://www.w3.org/2000/xmlns"
```


angelozerr:
It's perfect for me @xorye . Thanks!

PR #459, opened 2019-06-21
Fix error range for src-ct.1

### LINKED ISSUE #453: Error range for src-ct.1
With this XML file:

```xml
<?xml version="1.1" ?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified" attributeFormDefault="unqualified">
	 	
	<xs:complexType name="fullpersoninfo">
		<xs:complexContent>
			<xs:extension base="xs:string">
			</xs:extension>
		</xs:complexContent>
	</xs:complexType>
	
</xs:schema>
```

The `xs:string` attribute value should be highligted as an error.

xorye:
Fixes #453 

image

The arguments provided by xerces were:
```
arguments[0] = "fullpersoninfo"
arguments[1] = "string"
```

The "base" string was hard coded in my solution because this error happens when xerces checks the base type.


angelozerr:
Thank's @xorye !

PR #468, opened 2019-06-25
Fix error range for s4s-elt-must-match.2

### LINKED ISSUE #458: Error range for s4s-elt-must-match.2
With the XSD file:

```xml
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" version="1.0">
	
	<xs:simpleType name="X" ></xs:simpleType>
	
</xs:schema>
```

`xs:simpleType `should be highlighted with the `s4s-elt-must-match.2` error

xorye:
Fixes #458 

image

For this error range, I simply used the start tag name for the given offset.

The arguments were:
```
arguments[0] = "simpleType"
arguments[1] = "(annotation?, (restriction | list | union))"
```


angelozerr:
Thanks @xorye !

PR #469, opened 2019-06-25
Fix error range for ct-props-correct.3 when attribute value has a namespace

### LINKED ISSUE #467: Error range for ct-props-correct.3
This XSD file:

```xml
<?xml version="1.1" ?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" targetNamespace="http://www.w3.org/2001/XMLSchema">
	<xs:complexType name="aComplexType">
		<xs:complexContent>
			<xs:extension base="xs:aComplexType"></xs:extension>
		</xs:complexContent>
	</xs:complexType>
</xs:schema>
```

should highlight `xs:aComplexType`

xorye:
Fixes #467 

image

Previously, the error range was incorrect when the attribute value had a namespace.
This PR fixes the range, and considers both the case where the value has a namespace, and doesn't.


angelozerr:
Thank's @xorye !

PR #471, opened 2019-06-26
Add support for `textDocument/codeLens` for XML Schema types

### LINKED ISSUE #55: Add support for `textDocument/codelens` for XML Schema types
Add support for codelens https://microsoft.github.io/language-server-protocol/specification#textDocument_codeLens for extension.

Here a screenshot with XSD types references

image

angelozerr:
Fix #55

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLLanguageServer.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLLanguageServer.java
@@ -46,6 +46,7 @@ import org.eclipse.lsp4xml.settings.InitializationOptionsSettings;
 import org.eclipse.lsp4xml.settings.LogsSettings;
 import org.eclipse.lsp4xml.settings.ServerSettings;
 import org.eclipse.lsp4xml.settings.SharedSettings;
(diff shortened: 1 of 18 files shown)


xorye:
Very cool!

I have a few things to bring up:
gif
(Open the gif in a new tab to view more clearly)

In XMLSchema.xsd, for "simpleType", codelens indicates that there are 3 references, but upon clicking "Find All References", I can only see one. Similarly, for "facet", codelens says 24 references, but I can only see 15. Am I testing this correctly? Are there other references located separate files?

Also, one minor detail, I think that the word "reference" should be plural if there are 0 or more than 1 reference, like the following in the java codelens: 

image

And singular if there is only 1:
image

angelozerr:
> In XMLSchema.xsd, for "simpleType", codelens indicates that there are 3 references, but upon clicking "Find All References", I can only see one. Similarly, for "facet", codelens says 24 references, but I can only see 15. Am I testing this correctly? Are there other references located separate files?

Good catch, it was a bug, it should be fixed.

> Also, one minor detail, I think that the word "reference" should be plural if there are 0 or more than 1 reference, like the following in the java codelens:

You are right, it should be fixed.

PR #474, opened 2019-06-27
Fix error range for sch-props-correct.2

### LINKED ISSUE #462: Error range for sch-props-correct.2
When two XML xs:element is declared with the same name:

```xml
<?xml version="1.1" ?>
	<xs:element name="elt1" />
	<xs:element name="elt1" />
</xs:schema> 
```

elt1 attribute value should be highlighted.

xorye:
Fixes #462

image

The only argument provided by xerces was:
`arguments[0] = ",elt1"`

From the xerces source code, it looks like if a namespace exists, it gets placed right before the attribute value:
https://github.com/apache/xerces2-j/blob/cf0c517a41b31b0242b96ab1af9627a3ab07fcd2/src/org/apache/xerces/impl/xs/traversers/XSDHandler.java#L3384-L3389

I'm not sure if I interpreted this correctly, but I tried to take into account the situation where a namespace does exist:
https://github.com/xorye/lsp4xml/blob/be47ec532e8b566b1c779efe50253fc18e41216c/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/participants/XSDErrorCode.java#L137

I did not add a test case for that specific situation, because I'm not very sure how to add a namespace and create a `sch-props-correct.2` error at the same time.

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/participants/XSDErrorCode.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/participants/XSDErrorCode.java
@@ -40,6 +40,7 @@ public enum XSDErrorCode implements IXMLErrorCode {
 	s4s_att_not_allowed("s4s-att-not-allowed"), //
 	s4s_att_invalid_value("s4s-att-invalid-value"), //
 	s4s_elt_character("s4s-elt-character"), //
+	sch_props_correct_2("sch-props-correct.2"),
 	src_ct_1("src-ct.1"),
 	src_element_3("src-element.3"),
 	src_resolve_4_2("src-resolve.4.2"), //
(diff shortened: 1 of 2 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/participants/XSDErrorCode.java
```diff
@@ -128,7 +129,17 @@ public static Range toLSPRange(XMLLocator location, XSDErrorCode code, Object[]
 		}
 		case s4s_elt_character:
 			return XMLPositionUtility.selectContent(offset, document);
-		case src_ct_1: 
+		case sch_props_correct_2: {
+			String attrName = (String) arguments[0];
+			if (attrName.charAt(0) == ',') {
+				attrName = attrName.substring(attrName.indexOf(",") + 1);
+			} else {
+				attrName = attrName.replace(',', ':');
```
angelozerr: I have never seen case with prefix in xs:element/@name (but perhaps I'm wrong). So please remove this case. If we find the case with prefix in the future, we will fix it and add a dedicated tests.

angelozerr: Please ignore my previous comment, I found a case with : in element name. We should add a test with that.

angelozerr:
Thanks @xorye !

PR #475, opened 2019-06-27
Fix error range for EmptyTargetNamespace

### LINKED ISSUE #472: Error range for EmptyTargetNamespace
With this XSD file:

```xml
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
           elementFormDefault="qualified" xml:lang="EN"
           targetNamespace=""
           version="1.0">
</xs:schema>
```

Error should highlight the attribute value of targetNamespace

xorye:
Fixes #472 

image


angelozerr:
Thanks @xorye !

PR #478, opened 2019-06-28
Discrepancy in completion between prefixed and default namespaces

### LINKED ISSUE #311: Discrepancy in completion between prefixed and default namespaces
https://github.com/redhat-developer/vscode-xml/issues/87

angelozerr:
Fix #311

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/model/CMDocument.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/model/CMDocument.java
@@ -20,8 +20,23 @@ import org.eclipse.lsp4xml.dom.DOMElement;
  */
 public interface CMDocument {
 
+	/**
+	 * Returns true if the model document defines the given namespace and false
+	 * otherwise.
+	 * 
+	 * @param namespaceURI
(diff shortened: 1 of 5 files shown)


angelozerr:
To review this PR, 

 * download the attached project in https://github.com/redhat-developer/vscode-xml/issues/87#issue-380298676 (vscode-xml-catalog-test.zip)

 * open mini-catalog.xml set cursor after comment 

`  
`

open the completion, edmx:DataServices and edmx:Reference, #region, #endregion should only displayed (not edm:Annotation)

xorye:
> * open mini-catalog.xml set cursor after comment
@angelozerr Did you mean mini-catalog-prefix.xml instead? Because the original issue [here](https://github.com/redhat-developer/vscode-xml/issues/87) said that mini-catalog.xml  was good

angelozerr:
> Did you mean mini-catalog-prefix.xml instead? Because the original issue here said that mini-catalog.xml was good

Ooops sorry, yes it was that, but the 2 XML file mini-catalog-prefix.xml  and mini-catalog.xml should do the same completion result.

xorye:
@angelozerr  Works great! The completion also matches up  with the comments in `/examples/mini-catalog-prefix`, not just for `<!-- should suggest edmx:DataServices and edmx:Reference here -->`.

angelozerr:
Thanks @xorye for your feedback!

PR #486, opened 2019-06-28
Fix error range for cvc-pattern-valid

### LINKED ISSUE #477: Error range for cvc-pattern-valid
Take the `pattern.xsd` XML Schema file:

```xml
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
	elementFormDefault="qualified" attributeFormDefault="unqualified">

	<xs:element name="Annotation">
		<xs:complexType>
			<xs:attribute name="Term" type="TQualifiedName"
				use="required" />
		</xs:complexType>
	</xs:element>


	<xs:simpleType name="TQualifiedName">
		<xs:restriction base="xs:NCName">
			<xs:pattern
				value="[\p{L}\p{Nl}_][\p{L}\p{Nl}\p{Nd}\p{Mn}\p{Mc}\p{Pc}\p{Cf}]{0,}(\.[\p{L}\p{Nl}_][\p{L}\p{Nl}\p{Nd}\p{Mn}\p{Mc}\p{Pc}\p{Cf}]{0,}){1,}" />
		</xs:restriction>
	</xs:simpleType>

</xs:schema>
```

The XML file bound to this XML pattern.xsd:

```xml
<Annotation
	xsi:noNamespaceSchemaLocation="pattern.xsd"
	Term="X"
	>
</Annotation>
```

should highlight `"X"` as error.

xorye:
Fixes #477 

image

I was thinking that this is an XMLSchema error, please correct me if I am wrong.


angelozerr:
It's perfect @xorye !

PR #493, opened 2019-07-01
Add telemetry/event support

### LINKED ISSUE #430: Add telemetry/event support
collect at least the startup event with 
- Server Version
- JVM Version
- memory

angelozerr:
This PR fix #430

The OSUtils was renamed to Platform which is more generic (eclipse like
name) which contains information about OS, JVM (name, version, memory).

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/Platform.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/Platform.java
@@ -1,0 +1,188 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+
+package org.eclipse.lsp4xml;
+
+import org.eclipse.lsp4xml.utils.StringUtils;
+
+/**
+ * PLatform information about OS and JVM.
+ */
+public class Platform {
+
+	private static final String UNKNOWN_VALUE = "unknown";
+
+	private static final OS os = new OS();
+
+	public static final boolean isWindows = getOS().isWindows();
+	public static String SLASH = isWindows ? "\\" : "/";
+
+	/**
+	 * OS information
+	 */
+	public static class OS {
+
+		private final String name;
+
+		private final String version;
+
+		private final String arch;
+
+		private final transient boolean isWindows;
+
+		public OS() {
+			this.name = getSystemProperty("os.name");
+			this.version = getSystemProperty("os.version");
+			this.arch = getSystemProperty("os.arch");
+			isWindows = name != null && name.toLowerCase().indexOf("win") >= 0;
+		}
+
+		/**
+		 * Returns the OS name.
+		 * 
+		 * @return the OS name.
+		 */
+		public String getName() {
+			return name;
+		}
+
+		/**
+		 * Returns the OS version.
+		 * 
+		 * @return the OS version.
+		 */
+		public String getVersion() {
+			return version;
+		}
+
+		/**
+		 * Returns the OS arch.
+		 * 
+		 * @return the OS arch.
+		 */
+		public String getArch() {
+			return arch;
+		}
+
+		public boolean isWindows() {
+			return isWindows;
+		}
+	}
+
+	/**
+	 * JVM information
+	 *
+	 */
+	public static class JVM {
+
+		/**
+		 * JVM memory information
+		 *
+		 */
+		public static class Memory {
+
+			private final long free;
+
+			private final long total;
+
+			private final long max;
+
+			private Memory() {
+				super();
+				this.free = Runtime.getRuntime().freeMemory();
+				this.total = Runtime.getRuntime().totalMemory();
+				this.max = Runtime.getRuntime().maxMemory();
+			}
+
+			public long getFree() {
+				return free;
+			}
+
+			public long getTotal() {
+				return total;
+			}
+
+			public long getMax() {
+				return max;
+			}
+
+		}
+
+		private final String name;
+
+		private final String runtime;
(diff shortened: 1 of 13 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/Platform.java
```diff
@@ -0,0 +1,182 @@
...
+package org.eclipse.lsp4xml;
+
+/**
+ * PLatform information about OS and JVM.
+ */
+public class Platform {
+
+	private static final OS os = new OS();
+
+	private static final JVM jvm = new JVM();
```
fbricon: no need to keep it static, as it should be used once on startup

angelozerr:
@fbricon here my vscode trace:

```json
[Trace - 15:05:19] Received notification 'telemetry/event'.
Params: {
    "version": "0.8.0-20190701-1255",
    "os": {
        "name": "Windows 10",
        "version": "10.0",
        "arch": "amd64"
    },
    "jvm": {
        "name": "OpenJDK 64-Bit Server VM",
        "version": "1.8.0_212-3-redhat",
        "memory": {
            "free": 520615944,
            "total": 534773760,
            "max": 1073741824
        }
    }
}

```

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/Platform.java
angelozerr: I agree for JVM, but for os it used to initialize isWindows static variable

fbricon: yes, that's why I only commented for JVM

angelozerr:
@fbricon I have add telemetry to track explicit action from user, that's to say:

 * rename
 * formatting
 * rangeFormatting

In vscode trace, you should see :

```
[Trace - 09:44:17] Received notification 'telemetry/event'.
Params: "textDocument/formatting"
```

if you format a XML file.

angelozerr:
> Looks OK, but please add some unit tests before we merge it.

It will hard to do that since it depends on OS where JUnit are executed (I think about JVM and OS information).

Or perhaps you wish I write test just with rename, formatting, rangeFormatting, and onInitialized but by avoiding value check of OS, JVM memories?

An another question, I tell me if we should give the capability to send telemetry or not. Here some ideas:

 * add settings to disable telemetry?
 * add client capability 'telemetry' to enable telemetry. For instance I think LSP4E will never used those telemetry events.

fbricon:
You can check telemetry data contain the same values as what's in Platform, and then check memory values are > 0.

Yes an `xml.telemetry.enabled` preference would be useful.

angelozerr:
> Looks OK, but please add some unit tests before we merge it.

Done with https://github.com/angelozerr/lsp4xml/pull/493/files#diff-06ace1651d73790f809dd84073bc3a86R45

angelozerr:
> You can check telemetry data contain the same values as what's in Platform, and then check memory values are > 0.

Ok test was updated.

> Yes an xml.telemetry.enabled preference would be useful.

Done in PR https://github.com/redhat-developer/vscode-xml/pull/165

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/settings/XMLTelemetrySettings.java
```diff
@@ -8,15 +8,21 @@
...
-public class OSUtils {
-	
-	public static final boolean isWindows = System.getProperty("os.name").toLowerCase().indexOf("win") >= 0;
-	public static String SLASH = isWindows ? "\\" : "/";
+public class XMLTelemetrySettings {
 
-	
-}
\ No newline at end of file
+	private boolean enabled = true;
```
fbricon: should be off by default, clients should explicitly enable it

angelozerr: Good catch! It was a compy from XMLSymbolSettings. Copy/paste generate so many errors :)

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/Platform.java
```diff
@@ -0,0 +1,183 @@
...
+	 * Returns the system property from the given key and null otherwise.
+	 * 
+	 * @param key the property system key
+	 * @return the system property from the given key and null otherwise.
+	 */
+	private static String getSystemProperty(String key) {
+		try {
+			return System.getProperty(key);
+		} catch (SecurityException e) {
+			return null;
```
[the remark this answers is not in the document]
angelozerr: @fbricon updated

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/Platform.java
```diff
@@ -0,0 +1,183 @@
...
+	 */
+	public static OS getOS() {
+		return os;
+	}
+
+	/**
+	 * Returns the system property from the given key and null otherwise.
+	 * 
+	 * @param key the property system key
+	 * @return the system property from the given key and null otherwise.
```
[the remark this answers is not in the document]
angelozerr: @fbricon updated

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/telemetry/TelemetryManager.java
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * The telemetry notification is sent from the server to the client to ask the
+	 * client to log a telemetry event.
+	 */
+	private void telemetryEvent(Object object) {
+		languageClient.telemetryEvent(object);
+	}
+
+}
```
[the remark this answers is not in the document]
angelozerr: @fbricon updated

angelozerr:
Replaced with https://github.com/eclipse/lemminx/pull/989

PR #497, opened 2019-07-02
Fix error range for s4s-elt-invalid-content.3

### LINKED ISSUE #496: Error range for s4s-elt-invalid-content.3
With the given XSD file:

```xml
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified">
	<xs:element name="project" type="xs:string"></xs:element>
	<xs:import ></xs:import>
</xs:schema>
```

`xs:import` should be higlighted as error with message:

`s4s-elt-invalid-content.3: Elements of type 'import' cannot appear after declarations as children of a <schema> element.
`

We could add a code action to remove this xs:import element.

xorye:
Fixes #496 

image


angelozerr:
It's perfect @xorye. Do you want to try to implement codeaction too in this PR to remove the <xs:include> or do you prefer doing that in another PR?

xorye:
@angelozerr I can do that in this PR.

My most recent change implements a code action and I've updated the test case to check for the code action as well.

I've taken care of the case where the element to be removed is a self closing element, or not.

image

angelozerr:
Well done @xorye !

PR #507, opened 2019-07-11
Cast to Number instead of Double in ExtractAbbreviation.java (Emmet extension)

jonburs:
Fixes https://github.com/angelozerr/lsp4xml/issues/219

### DIFF
--- a/extensions/org.eclipse.lsp4xml.extensions.emmet/src/main/java/org/eclipse/lsp4xml/extensions/emmet/emmetio/extract/ExtractAbbreviation.java
+++ b/extensions/org.eclipse.lsp4xml.extensions.emmet/src/main/java/org/eclipse/lsp4xml/extensions/emmet/emmetio/extract/ExtractAbbreviation.java
@@ -75,9 +75,9 @@ public class ExtractAbbreviation {
 				return null;
 			}
 			String abbreviation = (String) s.get("abbreviation");
-			Double location = (Double) s.get("location");
-			Double start = (Double) s.get("start");
-			Double end = (Double) s.get("end");
+			Number location = (Number) s.get("location");
+			Number start = (Number) s.get("start");
+			Number end = (Number) s.get("end");
 			return new ExtractAbbreviationResult(abbreviation, location != null ? location.intValue() : 0,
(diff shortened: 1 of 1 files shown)


extensions/org.eclipse.lsp4xml.extensions.emmet/src/main/java/org/eclipse/lsp4xml/extensions/emmet/emmetio/extract/ExtractAbbreviation.java
```diff
@@ -75,9 +75,9 @@ public static ExtractAbbreviationResult extractAbbreviation(String line, Integer
 				return null;
 			}
 			String abbreviation = (String) s.get("abbreviation");
-			Double location = (Double) s.get("location");
-			Double start = (Double) s.get("start");
-			Double end = (Double) s.get("end");
+			Number location = (Number) s.get("location");
```
angelozerr: To support both JDK8 and JDK11, I suggets you to have a private  static method `getInt` which manages Number and Double like this:

```java
private static int getInt(Bindings bindings, String key) {
	Object result = bindings.get(key);		
	if (result instanceof Number) {
		return ((Number) result).intValue();
	}
	if (result instanceof Double) {
		return ((Double) result).intValue();
	}
	return 0;
}
```

After that you call it like this:

`int location = getInt(s, "location");`

jonburs: I don't believe this is necessary. `Double` and `Integer` (et al.) both implement `Number` (see https://docs.oracle.com/javase/8/docs/api/java/lang/Number.html), so the current change works on both platforms -- were this not the case the CI run would have failed.

angelozerr:
You are right! Please ignore my comment. I will merge your pr but emmet
support should ne rewritten in pure java to avoid using JS engine.

My idea is to provide an emmet support based on XML schema. I will describe
more my idea in an issue. Are you interested with our emmet support?

Le jeu. 11 juil. 2019 18:31, Jonathan Burstein <notifications@github.com> a
écrit :

angelozerr:
Thanks @jonburs !

Issue #506, opened 2019-07-10
Improve revalidate XML files when XSD files is saved


angelozerr:
When a XSD file is saved, it loop for all opened document and for each XML document, it checks if the saved XSD file bind the the current XML file. To do that DOMDocument#useSchemas is called. This method must be improved

1)  Windows OS problem

When XSD file is saved on Windows OS, it should revalidate all bound XML files, but it didn't do that.

After debugging, I have this error:

java.nio.file.InvalidPathException: Illegal char <:> at index 2: /C:/Users/azerr/Downloads/data_gxl_minimal++_minimal++_OHNE_DOC.gxl/book.xsd

2) Manage only xs:schemaLocation, xs:noNamespaceSchemaLocation binding

If opened XML file is bound with a XML catalog, XML file assocotaions, DOMDocumnt#useSchemas doesn't takes care of this binding kind.

To fix both 1) and 2), the DOMDocumnt#useSchemas  must be rewritten by using resolverExtensionManager#resolve.

Issue #132, opened 2018-09-20
Only end tag is ignored by XML parser


angelozerr:
The following XML content:

```xml
<\a>
```
or 

```xml
<root>
  <\a>
</root>
```
ignores the end tag element `<\a>` by the XML parser which causes many problem with 

 * higlight (root is not higlighted)
 * rename (root cannot be rename)
 * document symbols doesn't return the element a
 * format remove the element a.

To fix thoses issues, the XML parser should take care of this end tag only case.

PR #509, opened 2019-07-11
Validate open .xml files when bounded .xsd files are externally saved

xorye:

Fixes [vscode-xml #132](https://github.com/redhat-developer/vscode-xml/issues/132).

This PR was tested with this `xml` document:
```
<root
    xsi:noNamespaceSchemaLocation="sequence.xsd">
  <tag></tag>
  <optional></optional>
  <optional></optional>
  <optional></optional>
</root>
```
With this `xsd` schema:
```
<xs:schema
    elementFormDefault="qualified">
  <xs:element name="root">
    <xs:complexType>
      <xs:sequence>
        <xs:element name="tag"/>
        <xs:element
            name="optional"
            minOccurs="0"
            maxOccurs="2"/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
</xs:schema>
``` 
In the `xsd` file, I simply change the `maxOccurs` value from `2` to `3` and back to `2` in a separate editor to test this fix.

This PR will register all `xsd` files as watched files, therefore when the `xsd` files are saved externally, the client would send the `didChangeWatchedFiles`  notification. The server will check through every open xml document and validate if it is bound to the recently saved `xsd` file.

[This](https://github.com/xorye/lsp4xml/blob/f71bade25c1a02d115937bcab3c1a35e6a39c377/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/ContentModelPlugin.java#L59) check was needed because `document` is null if the document representing `documentURI` is not opened on the client side. In our case, `document` is null when an `xsd` that is not currently open in the client, has been modified and saved externally.

Note that this PR works whether or not the `xsd` file is currently open in the client. I have only written the unit test for the unopened case. I am more than willing to write one for the opened case if desired.

For the unit test, I created [this](https://github.com/xorye/lsp4xml/blob/f71bade25c1a02d115937bcab3c1a35e6a39c377/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java) new file. One thing I am not sure about are the two sleep calls [here](https://github.com/xorye/lsp4xml/blob/f71bade25c1a02d115937bcab3c1a35e6a39c377/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java#L132-L140). I added them because it seems like validation for the xml file is being done in a separate thread, and I wanted to wait until the validation was over, so I could read the resulting diagnostics/errors.

Validation for the `xml` document takes place when I call `didOpen()` with the xml document [here](https://github.com/xorye/lsp4xml/blob/f71bade25c1a02d115937bcab3c1a35e6a39c377/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java#L204) which returns `void`.

I did not test this on Windows yet, however a fix for Windows overlaps with #506 .

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLWorkspaceService.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLWorkspaceService.java
@@ -15,10 +15,11 @@ import java.util.concurrent.CompletableFuture;
 
 import org.eclipse.lsp4j.DidChangeConfigurationParams;
 import org.eclipse.lsp4j.DidChangeWatchedFilesParams;
+import org.eclipse.lsp4j.FileEvent;
 import org.eclipse.lsp4j.SymbolInformation;
 import org.eclipse.lsp4j.WorkspaceSymbolParams;
 import org.eclipse.lsp4j.services.WorkspaceService;
-
+import org.eclipse.lsp4xml.XMLTextDocumentService;
 /**
  * XML workspace service.
  *
@@ -44,9 +45,12 @@ public class XMLWorkspaceService implements WorkspaceService {
 
 	@Override
 	public void didChangeWatchedFiles(DidChangeWatchedFilesParams params) {
-
+		XMLTextDocumentService xmlTextDocumentService = (XMLTextDocumentService) xmlLanguageServer.getTextDocumentService();
+		List<FileEvent> changes = params.getChanges();
+		for (FileEvent change: changes) {
+			if (!xmlTextDocumentService.documentIsOpen(change.getUri())) {
+				xmlTextDocumentService.doSave(change.getUri());
+			}
+		}
 	}
-
-	
-
 }
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/ContentModelPlugin.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/ContentModelPlugin.java
@@ -56,17 +56,17 @@ public class ContentModelPlugin implements IXMLExtension {
 			// The save is done for a given XML file
 			String documentURI = context.getUri();
 			DOMDocument document = context.getDocument(documentURI);
-			if (DOMUtils.isCatalog(document)) {
+			if (document != null && DOMUtils.isCatalog(document)) {
 				// the XML document which has changed is a XML catalog.
 				// 1) refresh catalogs
 				contentModelManager.refreshCatalogs();
-				// 2) Validate all opened XML files except the catalog which have changed
-				context.collectDocumentToValidate(d -> {
-					DOMDocument xml = context.getDocument(d.getDocumentURI());
-					xml.resetGrammar();
-					return !documentURI.equals(d.getDocumentURI());
-				});
 			}
+			// 2) Validate all opened XML files except the catalog which have changed
+			context.collectDocumentToValidate(d -> {
+				DOMDocument xml = context.getDocument(d.getDocumentURI());
+				xml.resetGrammar();
+				return !documentURI.equals(d.getDocumentURI());
+			});
 		} else {
 			// Settings
 			updateSettings(context);
@@ -89,6 +89,9 @@ public class ContentModelPlugin implements IXMLExtension {
 				// Validate all opened XML files
 				context.collectDocumentToValidate(d -> {
 					DOMDocument xml = context.getDocument(d.getDocumentURI());
+					if (xml == null) {
+						return false;
+					}
 					xml.resetGrammar();
 					return true;
 				});
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/settings/capabilities/XMLCapabilityManager.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/settings/capabilities/XMLCapabilityManager.java
@@ -36,12 +36,18 @@ import static org.eclipse.lsp4xml.settings.capabilities.ServerCapabilitiesConsta
 import static org.eclipse.lsp4xml.settings.capabilities.ServerCapabilitiesConstants.TEXT_DOCUMENT_LINK;
 import static org.eclipse.lsp4xml.settings.capabilities.ServerCapabilitiesConstants.TEXT_DOCUMENT_REFERENCES;
 import static org.eclipse.lsp4xml.settings.capabilities.ServerCapabilitiesConstants.TEXT_DOCUMENT_RENAME;
+import static org.eclipse.lsp4xml.settings.capabilities.ServerCapabilitiesConstants.WORKSPACE_WATCHED_FILES;
+import static org.eclipse.lsp4xml.settings.capabilities.ServerCapabilitiesConstants.WORKSPACE_WATCHED_FILES_ID;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import org.eclipse.lsp4j.ClientCapabilities;
+import org.eclipse.lsp4j.DidChangeWatchedFilesRegistrationOptions;
+import org.eclipse.lsp4j.FileSystemWatcher;
 import org.eclipse.lsp4j.Registration;
 import org.eclipse.lsp4j.RegistrationParams;
 import org.eclipse.lsp4j.Unregistration;
@@ -153,7 +159,19 @@ public class XMLCapabilityManager {
 		if (this.getClientCapabilities().isReferencesDynamicRegistrationSupported()) {
 			registerCapability(REFERENCES_ID, TEXT_DOCUMENT_REFERENCES);
 		}
+		if (this.getClientCapabilities().isDidChangeWatchedFilesRegistered()) {
+			registerWatchedFiles();
+		}
+		
 		syncDynamicCapabilitiesWithPreferences();
+	}
+
+	private void registerWatchedFiles() {
+		List<FileSystemWatcher> watchers = new ArrayList<>(2);
+		watchers.add(new FileSystemWatcher("**/*.xsd"));
+		watchers.add(new FileSystemWatcher("**/*.dtd"));
+		DidChangeWatchedFilesRegistrationOptions options = new DidChangeWatchedFilesRegistrationOptions(watchers);
+		registerCapability(WORKSPACE_WATCHED_FILES_ID, WORKSPACE_WATCHED_FILES, options);
 	}
 
 	/**
--- a/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
+++ b/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
@@ -1,0 +1,249 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4xml.extensions.contentmodel;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.net.URI;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+import com.google.common.base.Charsets;
+import com.google.common.io.Files;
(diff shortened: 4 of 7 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLWorkspaceService.java
```diff
@@ -44,9 +45,15 @@ public void didChangeConfiguration(DidChangeConfigurationParams params) {
 
 	@Override
 	public void didChangeWatchedFiles(DidChangeWatchedFilesParams params) {
-
+		saveChangeWatchedFiles(params);
 	}
 
-	
+	private void saveChangeWatchedFiles(DidChangeWatchedFilesParams params) {
```
fbricon: there's no added value in providing this private method. Might as well move the code directly under didChangeWatchedFiles

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/ContentModelPlugin.java
```diff
@@ -56,17 +56,17 @@ public void doSave(ISaveContext context) {
...
-				// 2) Validate all opened XML files except the catalog which have changed
-				context.collectDocumentToValidate(d -> {
-					DOMDocument xml = context.getDocument(d.getDocumentURI());
-					xml.resetGrammar();
-					return !documentURI.equals(d.getDocumentURI());
-				});
 			}
+			// 2) Validate all opened XML files except the catalog which have changed
+			context.collectDocumentToValidate(d -> {
+				DOMDocument xml = context.getDocument(d.getDocumentURI());
```
fbricon: should you check `xml` for null?

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/settings/capabilities/XMLCapabilityManager.java
```diff
@@ -153,9 +159,20 @@ public void initializeCapabilities() {
...
+		if (this.getClientCapabilities().isDidChangeWatchedFilesRegistered()) {
+			registerWatchedFiles();
+		}
+		
 		syncDynamicCapabilitiesWithPreferences();
 	}
 
+	private void registerWatchedFiles() {
+		List<FileSystemWatcher> watchers = new ArrayList<>();
+		watchers.add(new FileSystemWatcher("**/*.xsd"));
```
fbricon: why not also `**/*.dtd`?

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/settings/capabilities/XMLCapabilityManager.java
```diff
@@ -153,9 +159,20 @@ public void initializeCapabilities() {
...
 		}
+		if (this.getClientCapabilities().isDidChangeWatchedFilesRegistered()) {
+			registerWatchedFiles();
+		}
+		
 		syncDynamicCapabilitiesWithPreferences();
 	}
 
+	private void registerWatchedFiles() {
+		List<FileSystemWatcher> watchers = new ArrayList<>();
```
fbricon: `new ArrayList<>(1);` if you know the list size in advance,

If you now the size will be 1, then you can also do:
```java
List<FileSystemWatcher> watchers = Collections.singletonList(new FileSystemWatcher("**/*.xsd")));
```

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
```diff
@@ -0,0 +1,219 @@
...
+import org.eclipse.lsp4j.services.LanguageClient;
+import org.eclipse.lsp4xml.XMLLanguageServer;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.After;
+import org.junit.Test;
+
+/**
+ * This class tests the functionality where an XML file undergoes
+ * validation against its schema when the XSD schema file has been modified
```
fbricon: and DTD?

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
```diff
@@ -0,0 +1,219 @@
...
+ * and saved externally through means, not within the LS client.
+ * 
+ * @author David Kwon
+ *
+ */
+public class XMLExternalTest {
+
+	private String tempDirPath = "./src/test/resources/temp";
```
fbricon: never write into the source folders, prefer `target/` instead. I believe there are other test cases that do create temp files like that

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
```diff
@@ -0,0 +1,219 @@
...
+	private String xsdPath = tempDirPath + "/sequence.xsd";
+
```
fbricon: wrap such code with `//@formatter:off` and `//@formatter:on`, to prevent file formatting to bork the contents.

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
```diff
@@ -0,0 +1,219 @@
...
+	public void externalXSDTest() {
+		List<PublishDiagnosticsParams> actualDiagnostics = new ArrayList<>();
+		XMLLanguageServer languageServer = createServer(actualDiagnostics);
+		TextDocumentItem xmlTextDocument = getXMLTextDocumentItem("test.xml");
+		File testXsd = new File(xsdPath);
+
+		clientOpenFile(languageServer, xmlTextDocument);
+
+		try {
+			Thread.sleep(1000);
```
fbricon: that's too long. try slightly longer than the validation delay

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+			Assert.assertEquals(1, actualDiagnostics.size());
+			Assert.assertEquals(0, actualDiagnostics.get(0).getDiagnostics().size());
+
+			editXSDFileMaxOccurs(testXsd, 2);
+			didChangedWatchedFiles(languageServer, testXsd);
+
+			Thread.sleep(1000);
```
fbricon: too long

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
```diff
@@ -0,0 +1,219 @@
...
+			Assert.fail(e.getMessage());
+		}
+	}
+
+	@After
+	public void after() {
+		deleteTempDirIfExists();
+	}
+
+	private void deleteTempDirIfExists() {
```
fbricon: simply use guava's MoreFiles.deleteRecursively method

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+	/**
+	 * Creates temporary directory containing the xsd file for
+	 * the test
+	 */
+	@Before
+	public void before() {
+		deleteTempDirIfExists();
+		createTempDir();
+
```
fbricon: Use guava's `Files.asCharSink(new File(xsdPath), Charsets.UTF_8).write(xsdContents)`

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
```diff
@@ -0,0 +1,219 @@
...
+		try {
+			createXSDFile();
+		} catch (FileNotFoundException e) {
```
fbricon: no need to catch the exception, just add `throws` to the method's signature, the test will fail anyway

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLWorkspaceService.java
```diff
@@ -44,9 +45,15 @@ public void didChangeConfiguration(DidChangeConfigurationParams params) {
...
 	public void didChangeWatchedFiles(DidChangeWatchedFilesParams params) {
-
+		saveChangeWatchedFiles(params);
 	}
 
-	
+	private void saveChangeWatchedFiles(DidChangeWatchedFilesParams params) {
+		XMLTextDocumentService xmlTextDocumentService = (XMLTextDocumentService) xmlLanguageServer.getTextDocumentService();
+		List<FileEvent> changes = params.getChanges();
+		for (FileEvent change: changes) {
```
fbricon: if the file is already opened, i.e. tracked by the server, then it should not be revalidated, or else the job will be done twice

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/settings/capabilities/XMLCapabilityManager.java
angelozerr: DOMUtils provides constants for xsd and dtd extension file. I think we should have a constants which build this regexp in DOMUtils. I say that because there are other extension file like ent for dtd. And perhaps we should give the capability to add other extension file. See my comment https://github.com/angelozerr/lsp4xml/issues/380#issuecomment-494191582

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/ContentModelPlugin.java
angelozerr: This case could occur when DOM document is parsing, but never seen this case.

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/extensions/contentmodel/XMLExternalTest.java
xorye: The test currently only tests with an xsd file. Should I add another test that uses a dtd file?

fbricon: yes

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLWorkspaceService.java
xorye: > if the file is already opened,

Are you referring to the xml file or xsd file? In my test case, the xsd file is not opened while the xml is. I'm not sure when the validation would be done twice, because, without this code snippet, validation for opened xmls won't happen at all when xsd files (open or unopen) are changed externally. Could you explain a bit more about when validation might happen twice?

fbricon: if the xsd is also opened in lsp4xml, saving it would trigger both documentChange and fileWatchChanged events. But only the document change is interesting in that case

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/settings/capabilities/XMLCapabilityManager.java
xorye: @angelozerr for now, should I just add `**/*.dtd` here, manually in the code?

angelozerr: Do like this for the moment, but we need to change it in the future to give the capability to user to define custom file extension to watch.

xorye:
I've made the changes.

Instead of sleeping for 1000 milliseconds, it now sleeps for 600. It can go as low as around 530ms, but when it is that low, sometimes the validation takes longer so the tests fail. Or maybe it depends on the machine running the tests?

The temp folder is now being created in the target directory.

angelozerr:
Works like a charm on Windows OS (after applying my future work https://github.com/angelozerr/lsp4xml/tree/improve-revalidate)

Well done @xorye !

PR #511, opened 2019-07-15
Fix NPE with XSD highlight for attribute without value

angelozerr:
Fix NPE with XSD highlight for attribute without value

Take this XML Schema:

```xml
<?xml version="1.0" ?>

	<xs:element name="resources">
		<xs:complexType>
			<xs:sequence>
				<xs:element ref
```
and set cursor in ref (without value), NPE will occurs with master


xorye:
Looks good!

PR #513, opened 2019-07-16
Improve revalidate XML files when XSD files is saved

### LINKED ISSUE #506: Improve revalidate XML files when XSD files is saved

angelozerr:
Fix #506


xorye:
Works for me. I've edited an xsd (bounded to xml files with an xml catalog) from within VSCode and out of VSCode and xml validation was working.

angelozerr:
Thank's for your feedback! @fbricon I merge this PR. @xorye has checked that I have not broken something and existing test are working.

PR #515, opened 2019-07-17
XSD 1.1 working

### LINKED ISSUE #363: Does not support XSD 1.1
I just got with a document instance, that is specced with XSD v1.1:

`cos-all-limited.2: The {max occurs} of an element in an 'all' model group must be 0 or 1. The value 'unbounded' for element 'Bcc' is invalid.`

However, when validating with Xerxes2 I get no complaints. Could it be, that XSD 1.1 is not supported?

angelozerr:
XSD 1.1 working

Fixes #363 

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Features
 * [textDocument/hover](https://microsoft.github.io/language-server-protocol/specification#textDocument_hover).
 * [textDocument/rangeFormatting](https://microsoft.github.io/language-server-protocol/specification#textDocument_rangeFormatting)
 * [textDocument/rename](https://microsoft.github.io/language-server-protocol/specification#textDocument_rename).
+* XML Schema `1.0` and `1.1` support.t
 
 See screenshots in the [wiki](https://github.com/angelozerr/lsp4xml/wiki/Features).
 
--- a/org.eclipse.lsp4xml/pom.xml
+++ b/org.eclipse.lsp4xml/pom.xml
@@ -1,4 +1,6 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 	<parent>
 		<groupId>org.lsp4xml</groupId>
@@ -96,11 +98,20 @@
 			<groupId>org.eclipse.lsp4j</groupId>
 			<artifactId>org.eclipse.lsp4j.jsonrpc</artifactId>
 		</dependency>
+
+		<!-- Xerces 1.1 -->
 		<dependency>
-			<groupId>xerces</groupId>
+			<groupId>org.exist-db.thirdparty.xerces</groupId>
 			<artifactId>xercesImpl</artifactId>
 			<version>2.12.0</version>
+			<classifier>xml-schema-1.1</classifier>
 		</dependency>
+		<dependency>
+			<groupId>com.rackspace.eclipse.webtools.sourceediting</groupId>
+			<artifactId>org.eclipse.wst.xml.xpath2.processor</artifactId>
+			<version>2.1.100</version>
+		</dependency>
+
 		<dependency>
 			<groupId>xml-apis</groupId>
 			<artifactId>xml-apis</artifactId>
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/ContentModelPlugin.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/ContentModelPlugin.java
@@ -37,7 +37,7 @@ public class ContentModelPlugin implements IXMLExtension {
 
 	private final IHoverParticipant hoverParticipant;
 
-	private final IDiagnosticsParticipant diagnosticsParticipant;
+	private IDiagnosticsParticipant diagnosticsParticipant;
 
 	private final ICodeActionParticipant codeActionParticipant;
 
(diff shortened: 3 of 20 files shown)


angelozerr:
@fbricon I fixed the OutOfMemory with XSD 1.1 which was a problem coming from LSP4XML (an infinite loop). All tests work. If you are OK to merge it, it misses tests:

 * test with LSPMessageFormatter because there are no test with LSPMessageFormatter
 * test with a sample of validation with XSD 1.1.

If you wish I work today on it, please tell me.

fbricon:
@angelozerr tests for 1.1 support are needed indeed

angelozerr:
Ok I will write test today.

angelozerr:
> @angelozerr tests for 1.1 support are needed indeed

Done.

To test this PR, create you a xsd file like this:

```xml
<xs:schema 
	xmlns:t="test" 
	xmlns:vc="http://www.w3.org/2007/XMLSchema-versioning" 
	targetNamespace="test" 
	vc:minVersion="1.0">
	
	<xs:complexType name="testType">
		<xs:all>
			<xs:element name="testEle" minOccurs="1" maxOccurs="unbounded" type="xs:string"/>
			<xs:element name="testEleTwo" type="xs:string"/>
		</xs:all>
	</xs:complexType>
	
</xs:schema>
```

It's XSD 1.0, you should see the following error:

image

After that change vc:minVersion attribute with 1.1 like this:

```xml
<xs:schema 
	xmlns:t="test" 
	xmlns:vc="http://www.w3.org/2007/XMLSchema-versioning" 
	targetNamespace="test" 
	vc:minVersion="1.0">
	
	<xs:complexType name="testType">
		<xs:all>
			<xs:element name="testEle" minOccurs="1" maxOccurs="unbounded" type="xs:string"/>
			<xs:element name="testEleTwo" type="xs:string"/>
		</xs:all>
	</xs:complexType>
	
</xs:schema>
```
The error should gone.

fbricon:
please update the feature list in the readme

angelozerr:
@fbricon please don't review this PR. There are some trouble with completion.

xorye:
@angelozerr Validation with the example you provided works on my end.

I also briefly tested validation with some XSD 1.1 specific features found here: https://blogs.oracle.com/rammenon/xml-schema-11-ndash-what-you-need-to-know and it seems to work correctly (although I am not too knowledgable about XSD).

NikolasKomonen:
Also tested with some XSD 1.1 stuff. Works for me as well.

angelozerr:
> Also tested with some XSD 1.1 stuff. Works for me as well.

Thank's for your feedback @NikolasKomonen. I need to finish to write tests to accept this PR and if @fbricon is OK.

deathaxe:
Looks like it needs some work?

GitMensch:
Could anyone with write access try to fix the conflicts allowing others to test it (for example in the redhat XML plugin)?

It seems the following is far away:

> I'm working on this issue with PR https://github.com/angelozerr/lsp4xml/pull/515 It should be available for 0.8.0

_Originally posted by @angelozerr in https://github.com/eclipse/lemminx/issues/363#issuecomment-512193756_

Given the issues mentioned in https://github.com/redhat-developer/vscode-xml/issues/222

angelozerr:
> If so, [I've contacted](https://lists.apache.org/thread/wkg4qo0smrnq2q4oj64w083338m2543c) the Xerces-J maintainers, and they answered that they won't publish on Maven, but can accept volunteers that do that.

If I remember @apupier wanted to do that (not sure), but we would like to avoid publishing a non official Xerces XSD 1.1 maven project

apupier:
> > If so, [I've contacted](https://lists.apache.org/thread/wkg4qo0smrnq2q4oj64w083338m2543c) the Xerces-J maintainers, and they answered that they won't publish on Maven, but can accept volunteers that do that.
> 
> If I remember @apupier wanted to do that (not sure), but we would like to avoid publishing a non official Xerces XSD 1.1 maven project

"wanted" is a bit too strong ;-)
I published one version of xerces-j xsd 1.0
I guess I should have rights to publish another version.
if i remember well I didn't published the 1.1 version published because it was never done before and some specific artifacts for it were missing (and I wasn't able to build locally)

mlocati:
I managed to use the "xsd-1.1" variant of Xerces with Maven without much problems: see https://github.com/concrete5-community/concrete-cif

In particular, here's what I do in order to achieve that:

1. I download the Xerces archive (xsd-1.1 variant) and "install" its jars in the local m2 archive with `mvn install:install-file` ([see here](https://github.com/concrete5-community/concrete-cif/blob/0a35b5d08792ebca58ea7065d720d25c88d2058f/bin/install-xerces-m2.sh))
2. I reference those packages in my pom file ([see here](https://github.com/concrete5-community/concrete-cif/blob/0a35b5d08792ebca58ea7065d720d25c88d2058f/pom.xml#L29-L59))
3. and finally I'm able to use an XSD 1.1 schema ([see here](https://github.com/concrete5-community/concrete-cif/blob/0a35b5d08792ebca58ea7065d720d25c88d2058f/src/main/java/org/concretecms_community/concrete_cif/SchemaValidator.java#L18-L19)).

mlocati:
> I wasn't able to build locally

@apupier what about asking the [j-dev mailing list](https://xerces.apache.org/mail.html#xerces-j-dev)?

apupier:
> > I wasn't able to build locally
> 
> @apupier what about asking the [j-dev mailing list](https://xerces.apache.org/mail.html#xerces-j-dev)?

This is anyway not a good way. I should not build an artifact locally and publish it myself. The build needs to be done from a trusted place.
I also do not have time to look deep into this.
You can find what I documented when i published a version here https://issues.apache.org/jira/browse/XERCESJ-1724 and the mention that there was nothing available for xsd 1.1.

mlocati:
Publishing to Maven seems rather complicated.

But... is it really necessary? Can't we simply install xerces in the local m2 repository [like I do in my project](https://github.com/concrete5-community/concrete-cif)?

mickaelistria:
lemminx-maven is mainly distributed and built as a Maven artifact. Any dependency to a library that is not available in a Maven repository would make it impossible to consume transitively from Maven, because of the missing transitive dependency.
So if lemminx-maven is to use a library that is not published in a Maven repo, then it should consider incuding it in its own Maven repository. As this wouldn't be an official artifact, it should keep the org.eclipse.lemminx groupId.

But overall, the fact that the artifact is not published on Maven with regular Xerces can be interpreted as a sign of immaturity, and do we want lemminx to build on top immature blocks?

I don't think we should aim at a workaround in lemminx, instead, effort should be invested in xerces project directly to make the bundle available. @mlocati As you seem to need it most, you should be the one getting in touch with the xerces support channels and leading the effort of getting the artifact published to Maven. Then we could just use it in lemminx without extra burden.

mlocati:
> But overall, the fact that the artifact is not published on Maven with regular Xerces can be interpreted as a sign of immaturity, and do we want lemminx to build on top immature blocks?

I'd disagree: Xerces is very stable and mature. The problem is simply that the Xerces team doesn't need to publish on Maven.

> @mlocati As you seem to need it most, you should be the one getting in touch with the xerces support channels and leading the effort of getting the artifact published to Maven

I play with Java only occasionally, I wouldn't call me an expert: it'd me much better if this task could be done by someone more skilled by be. And Maven is an absolute obscure thing for me.

irbian:
Hey, just wondering if it has been any advance regarding this, that I understand is the root of why XSLT 3.0 files fail to validate on Eclipse with the message "There are '45' errors in 'schema-for-xslt30.xsd'."

apupier:
for reference, the ticket asking for a xerces with xml schema 1.1 https://issues.apache.org/jira/browse/XERCESJ-1736

eirikwitt:
I would like to add my voice to advocating for this.
For reference, this is the newest ticket asking for xsd 1.1 maven distribution: https://issues.apache.org/jira/browse/XERCESJ-1788

angelozerr:
Thanks @eirikwitt for this great info!

Issue #380, opened 2019-05-20
.ent and .mod files should be parsed as DTDs


zwaldowski:
DTDs frequently modularize long lists of entities using external files with the `.ent` extension. These should get syntax-checked as DTDs as well.

Technically, any extension can be used for this purpose; I've also seen `.mod`.

angelozerr:
Perhaps lsp4xml should provide a preference for thoses DTD file extensions?

angelozerr:
> Technically, any extension can be used for this purpose; I've also seen .mod.

Yes sure but it's a little hard to do that because there are synatx coloration to manage in vscode, etc. After reading some documentation, it seems there are 3 extensions (as you said):

 * dtd
 * mod
 * ent

We need to support those extension

angelozerr:
For syntax coloration, we need wait for accept of my PR for vscode https://github.com/microsoft/vscode/pull/77496

PR #516, opened 2019-07-17
.ent and .mod files should be parsed as DTDs

angelozerr:
See #380


angelozerr:
> Can you please add some unit tests?

Done.

PR #527, opened 2019-07-22
Hover markup kind response ignores the hover client capability

### LINKED ISSUE #525: Hover markup kind response ignores the hover client capability
Today XML Schema, DTD returns a MARKDOWN kind although it should check that client has this capability

angelozerr:
Fix #525


angelozerr:
This PR gives the capability to check client capablity for hover to returns MARKDPON markup kind if client support it.

This PR 

 * rename CompletionSettings to XMLCompletionSettings to follow the same rules than other settings.
 * fix bug to get folding client capability

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/XMLFoldings.java
```diff
@@ -153,8 +154,7 @@ public TagInfo(int startLine, String tagName) {
 			}
 		} catch (BadLocationException e) {
 			LOGGER.log(Level.SEVERE, "Foldings received a BadLocation while scanning the document", e);
-		}
-		catch(StackOverflowError e) {
+		} catch (StackOverflowError e) {
```
fbricon: this should be logged at the very least

angelozerr: @fbricon updated

Issue #518, opened 2019-07-17
Error range for EntityNotDeclared


angelozerr:
Given this XML file:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE article [
  <!ELEMENT article (#PCDATA)>
]>          
<article>
	&nbsp;
</article>
```

&nbsp; should be highlighted. (It should be managed in XMLSyntaxErrorCode).

CodeAction could create the proper entity in the DOCTYPE like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE article [
  <!ELEMENT article (#PCDATA)>
  <!ENTITY nbsp "entity-value">
]>           
<article>
	&nbsp;
</article>
```
We must take care of existing of doctype. For instance if we have

```xml
<?xml version="1.0" encoding="UTF-8"?>          
<article>
	&nbsp;
</article>
```

CodeAction should generate

```xml
<!DOCTYPE article [
  <!ENTITY nbsp "entity-value">
]>
```

fbricon:
Error range fix has been merged. Code action will be handled in #532

PR #531, opened 2019-07-22
Fix error range for EntityNotDeclared

xorye:
Fixes the error range mentioned in #518 

image

I am using the name of the undeclared entity, (in this case, "nbsp"), looking for the substring "&nbsp" in the line provided by xerces to determine the error range.

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/DTDErrorCode.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/DTDErrorCode.java
@@ -17,6 +17,7 @@ import org.apache.xerces.xni.XMLLocator;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4xml.commons.BadLocationException;
+import org.eclipse.lsp4xml.commons.TextDocument;
 import org.eclipse.lsp4xml.dom.DOMDocument;
 import org.eclipse.lsp4xml.dom.DOMElement;
 import org.eclipse.lsp4xml.extensions.contentmodel.participants.codeactions.ElementDeclUnterminatedCodeAction;
@@ -36,6 +37,7 @@ public enum DTDErrorCode implements IXMLErrorCode {
 	AttTypeRequiredInAttDef,
 	ElementDeclUnterminated,
 	EntityDeclUnterminated,
+	EntityNotDeclared,
 	ExternalIDorPublicIDRequired,
 	IDInvalidWithNamespaces,
 	IDREFInvalidWithNamespaces,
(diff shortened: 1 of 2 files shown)


fbricon:
Could we underline both the starting `&` and the closing `;`? @angelozerr wdyt?

fbricon:
That's how eclipse does it:

Screen Shot 2019-07-22 at 7 28 26 PM

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/DTDErrorCode.java
```diff
@@ -154,7 +156,24 @@ public static Range toLSPRange(XMLLocator location, DTDErrorCode code, Object[]
...
+		case EntityNotDeclared: {
+			TextDocument textDocument = document.getTextDocument();
+			try {
+				Position position = textDocument.positionAt(offset);
+				int line = position.getLine();
+				String text = textDocument.lineText(line);
+				String name = (String) arguments[0];
+				String subString = "&" + name;
+				int start = text.indexOf(subString) + 1;
+				int end = start + name.length();
```
fbricon:
```java
int start = text.indexOf(subString);
int end = start + subString.length();
```

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/DTDErrorCode.java
```diff
@@ -154,7 +156,24 @@ public static Range toLSPRange(XMLLocator location, DTDErrorCode code, Object[]
...
+				Position startPosition = new Position(line, start);
+				Position endPosition = new Position(line, end);
+				return new Range(startPosition, endPosition);
```
fbricon: note for later: we should have a `public static Range newRange(line, start, end)` utility method somewhere

xorye: should this change take place in a different PR?

fbricon: definitely yes

angelozerr:
> note for later: we should have a public static Range newRange(line, start, end) utility method somewhere

XMLPositionUtility is the best place to add this method

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/DTDErrorCode.java
angelozerr: XMLPositionUtility is the best place to add this method

PR #542, opened 2019-07-23
Fix error range for `cvc-datatype-valid-1-2-1`

### LINKED ISSUE #323: Fix error range for `cvc-datatype-valid-1-2-1`
If only integers are allowed as content and you have:

```xml
<a xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
   xsi:noNamespaceSchemaLocation="intType.xsd">

  TEXT
  

</a>
```

with XSD:

```xml
  <xs:element name="a" type="xs:integer">
  </xs:element>
</xs:schema>
```

The highlight could be cleaned up to the first and last character of text here:
Selection_050

And in the case a user puts in a tag the highlight, error message should be adjusted to include the first tag it comes across:
Selection_051

xorye:
Fixes #323 

Xerces gives us the location of the element `<a>`. This PR looks at the contents of `<a>` to determine the error range.

image
image

If there are no non-whitespace characters in the element's contents, the error range looks like this:
image

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/XMLSchemaErrorCode.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/XMLSchemaErrorCode.java
@@ -20,6 +20,7 @@ import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4xml.commons.BadLocationException;
 import org.eclipse.lsp4xml.dom.DOMDocument;
+import org.eclipse.lsp4xml.dom.DOMElement;
 import org.eclipse.lsp4xml.dom.DOMNode;
 import org.eclipse.lsp4xml.dom.NoNamespaceSchemaLocation;
 import org.eclipse.lsp4xml.dom.SchemaLocation;
@@ -32,6 +33,7 @@ import org.eclipse.lsp4xml.extensions.contentmodel.participants.codeactions.cvc_
 import org.eclipse.lsp4xml.extensions.contentmodel.participants.codeactions.cvc_type_3_1_1CodeAction;
 import org.eclipse.lsp4xml.services.extensions.ICodeActionParticipant;
 import org.eclipse.lsp4xml.services.extensions.diagnostics.IXMLErrorCode;
+import org.eclipse.lsp4xml.utils.DOMUtils;
 import org.eclipse.lsp4xml.utils.XMLPositionUtility;
 
 /**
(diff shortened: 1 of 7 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/dom/DOMElement.java
```diff
@@ -377,6 +377,13 @@ public boolean isStartTagClosed() {
 	public boolean isEndTagClosed() {
 		return getEndTagCloseOffset() != NULL_VALUE;
 	}
+
+	/**
+	 * Returns true if element contains only DOMText
+	 */
+	public boolean containsTextOnly() {
```
angelozerr: Please remove this method in static method in DOMUtils for instance.

xorye: You mean, I should add it into DOMUtils as a static method?

xorye: and remove it from DOMelement.java?

angelozerr: yes it's exactly that. We should avoid adding new method (if it used once time) in DOMElement which is an API. If we see that it's very used, we could move to DOMElement again.

xorye: Oops, my bad. We spoke about this in yesterday's meeting

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/DOMUtils.java
```diff
@@ -89,4 +89,16 @@ public static boolean isDTD(String uri) {
 		return uri != null
 				&& (uri.endsWith(DTD_EXTENSION) || uri.endsWith(ENT_EXTENSION) || uri.endsWith(MOD_EXTENSION));
 	}
+
+	/**
+	 * Returns true if element contains only DOMText
+	 */
```
fbricon: please fix the javadoc

xorye: fixed

angelozerr:
@fbricon do you think we can merge this PR?

fbricon:
@angelozerr I trust you to review and test it

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/XMLSchemaErrorCode.java
```diff
@@ -192,11 +194,27 @@ public static Range toLSPRange(XMLLocator location, XMLSchemaErrorCode code, Obj
 		case cvc_type_3_1_1:
 			return XMLPositionUtility.selectAllAttributes(offset, document);
 		case cvc_complex_type_2_1:
-		case cvc_type_3_1_3:
 		case cvc_elt_3_2_1:
 			return XMLPositionUtility.selectContent(offset, document);
+		case cvc_type_3_1_3:
+		case cvc_datatype_valid_1_2_1: {
+			
+			String attrValue = (String) arguments[0];
```
NikolasKomonen: Could you grab this value using the StringUtils helper: https://github.com/angelozerr/lsp4xml/blob/master/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/StringUtils.java#L355

It handles the case the object is not a String.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/utils/StringUtils.java
```diff
@@ -263,6 +263,48 @@ public static int getOffsetAfterWhitespace(String text, int endOffset) {
...
+	}
+
+	/**
+	 * Returns the number of consecutive whitespace characters from
+	 * the end of text
+	 * @param text String of interest
+	 * @return the number of consecutive whitespace characters from
+	 * the end of text
+	 */
+	public static int getBackWhitespaceLength(String text) {
```
NikolasKomonen: rename to "getTrailingWhitespaceLength()"

NikolasKomonen:
@xorye Works well, just some small things to change. Should be good to merge after.

PR #548, opened 2019-07-25
Cache completion based on XML Schema/DTD

### LINKED ISSUE #547: Cache completion based on XML Schema/DTD
When completion based on XML Schema/DTD is done and XML Schema/DTD  comes from the file system:

 * the DTD/Schema URI use file system
 * the DTD/Schema URI use http but cache is activated

The XMl Schema / DTD is every time parsed and loaded. For a big XML Schema / DTD, there are some trouble:

 * completion can be slow.
 * if it's too slow, completion returns nothing.

To fix this issue, we should cache the completion (the CMDocument) like we have done for http but we should track all files changed of the XML Schema (which can imported other XML Schema) to clean the cache when the XMl Schema / DTD is modified.

angelozerr:
Fixes #547

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/model/FilesChangedTracker.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/model/FilesChangedTracker.java
@@ -1,0 +1,100 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4xml.extensions.contentmodel.model;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.attribute.FileTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * File changed tracker.
+ * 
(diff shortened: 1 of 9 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/model/FilesChangedTracker.java
```diff
@@ -0,0 +1,68 @@
...
+
+	}
+
+	private final List<FileChangedTracker> files;
+
+	public FilesChangedTracker() {
+		files = new ArrayList<>();
+	}
+
+	public void addFileURI(String fileURI) {
```
fbricon: you need a way to remove tracked files

angelozerr: why?

angelozerr:
@fbricon working on test

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/model/FilesChangedTracker.java
fbricon: to prevent memory leaks?

angelozerr: The track is very simply, it checks just the last modified of file eeach time. I don't use watcher files system.

fbricon: so what happens if the tracked file is deleted?

angelozerr: Good catch! XSD Schema was stored in memory and completion continued to work. It should not. I have fixed the problem and cover that in my test.

angelozerr:
@fbricon all tests was done, is there something that you wish I do to merge this PR?

PR #557, opened 2019-08-19
Fixes issue with error messages not showing

### LINKED ISSUE #553: Validation Error Message Fails on Certain Cases
First found from this issue: https://github.com/redhat-developer/vscode-xml/issues/177#issuecomment-520132932

The line of code that looks like this, https://github.com/angelozerr/lsp4xml/blob/fee585992ab5c1748708cf92aada4c7f01c7f220/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/XMLSchemaErrorCode.java#L155

fails because the type we are casting it to is not always a string in certain cases.

Instead use the toString() method, make a StringUtils method to handle this issue.

NikolasKomonen:
Fixes #553

Signed-off-by: NikolasKomonen <nikolaskomonen@gmail.com>


angelozerr:
Thanks @NikolasKomonen !

Issue #177, opened 2018-10-28
no validation or code completion on nested elements


ehigiepaul:
Good day,
I was referred by @angelozerr to report the issue I have using Red hat vscode xml extension.

I created a custom xsd schema for a xml file. Am using Vscode with Red hat vscode-xml extension.

The schema works for only the top elements giving me all its attributes and validation from my xsd file but not on the nested elements in the xml file.

I used the same xsd schema file in Jetbrain Intellij and it worked fine.

Here is an example of my issue.
![xml_schema_error](https://user-images.githubusercontent.com/8577824/47619779-9e69d480-dae2-11e8-8616-5608c03176fb.gif)

Please also find below are my xsd and xml file.
[kll Page.zip](https://github.com/angelozerr/lsp4xml/files/2523037/kll.Page.zip)

I get swirling lines on the top element every time I reload vscode. My fix for that was to remove or add attribute to page tag.

Please can you kindly resolve this issue.  Thanks

angelozerr:
Problem ame from tht use of `xs:any`. I have fixed the problem with completion.

For validation, it should work since Xerces is ised to validate the XML content. You should have the same validation between vscode-xml and Intellij  (I think Intellij  uses Xerces too).

ehigiepaul:
@angelozerr thanks for the fast fix, I will try it out now.

ehigiepaul:
Had some issue to build the lsp4xml but succeeded by changing the protocol to http due to ssl problem. 
But after this fix and successful build. Am happy to say that it **works like a charm**. Thanks for this amazing support. I wish you all the best in all you future plan.

angelozerr:
Glad it works for you.

ehigiepaul:
Sorry but have to reopen this again. The completion is working fine but not validation available anymore

fbricon:
Completion shows duplicates of nodes already present in the document:
<img width="653" alt="screen shot 2018-11-02 at 1 40 33 pm" src="https://user-images.githubusercontent.com/148698/47931376-07679880-dea5-11e8-9b0c-4666c8ab9176.png">

fbricon:
Can no longer reproduce the duplicates issue. Closing

PR #563, opened 2019-08-22
Fix completion duplicate with xs:any

angelozerr:
See #177


fbricon:
On completion, all tags already in the document are missing the documentation part which points to their source (tns.xsd). All other tags have it

fbricon:
well actually no, this is not the proper cause, eg. DockLayout has no `source tns.xsd` and it's not already defined in the test-page.xml either

fbricon:
Seems to be concerning all *Layout tags. Probably a different bug then

angelozerr:
> Seems to be concerning all *Layout tags. Probably a different bug then

I think it's because I provide local and global elements both. In DockLayout case, local element is added beforethe global element and it seems I cannot get the source in this case. To fix the problem completion provides only global elements.

fbricon:
so just add local elements after global ones?

angelozerr:
> so just add local elements after global ones?

I collect now local and global elements and fix bug with source. I have written test for that.

PR #570, opened 2019-10-08
Completion doesn't use file cache for included XML schema.

angelozerr:
Completion doesn't use file cache for included XML schema.

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/model/ContentModelManager.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/model/ContentModelManager.java
@@ -117,8 +117,8 @@ public class ContentModelManager {
 	private CMDocument findCMDocument(String uri, String publicId, String systemId,
 			ContentModelProvider modelProvider) {
 		// Resolve the XML Schema/DTD uri (file, http, etc)
-		String key = resolverManager.resolve(uri, publicId, systemId);
-		if (key == null) {
+		String resolvedUri = resolverManager.resolve(uri, publicId, systemId);
+		if (resolvedUri == null) {
 			return null;
 		}
 		// the XML Schema, DTD can be resolved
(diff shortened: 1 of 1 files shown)


fbricon:
any tests?

angelozerr:
The PR fix is to check if the resolved uri (and not the uri) is an http uri to use the cache. See https://github.com/angelozerr/lsp4xml/pull/570/files#diff-3560eeb4e3d41113ab9ad5a099e6de95L139

This bug was introduced in 0.9.0 and problem comes from when XML schema include other XML schema (writing test is hard). I notice this problem when I try to understand the problem with https://github.com/redhat-developer/vscode-xml/issues/183

To test this PR remove the .lsp4xml cache at hand and write a XML file like this:

```xml
<IODevice xmlns="http://www.io-link.com/IODD/2010/10" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.io-link.com/IODD/2010/10 http://www.io-link.com/IODD/2010/10/IODD1.1.xsd">
</IODevice>
```

It will download the XML Schema http://www.io-link.com/IODD/2010/10/IODD1.1.xsd and some included XML schema but there will have an error with `xml.xsd` file to download

It's because the included XML schema https://io-link.com/IODD/2010/10/IODD-Primitives1.1.xsd include xml.xsd like this:

```xml
<xsd:schema xmlns="http://www.io-link.com/IODD/2010/10" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:iodd="http://www.io-link.com/IODD/2010/10" targetNamespace="http://www.io-link.com/IODD/2010/10" elementFormDefault="qualified" attributeFormDefault="unqualified" version="1.1">
<xsd:import namespace="http://www.w3.org/XML/1998/namespace" schemaLocation="xml.xsd"/>
```

It tries to download https://io-link.com/IODD/2010/10/xml.xsd although it doesn't exists. After that try to copy/pastehttps://github.com/angelozerr/lsp4xml/blob/master/org.eclipse.lsp4xml/src/main/resources/schemas/xsd/xml.xsd  in the cache folder where IODD1.1.xsd  was downloaded and relaunch the completion. Without this PR it continues to fail, with this PR it should download other XML Schema and have none error.

angelozerr:
> any tests?

It's very hard to write tests. See the usecases that I explained below.

PR #579, opened 2019-10-18
Navigation and intellisense for xs:include-ed types

angelozerr:
Navigation and intellisense for xs:include-ed types

See https://github.com/redhat-developer/vscode-xml/issues/188

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/contentmodel/CMXSDDocument.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/xsd/contentmodel/CMXSDDocument.java
@@ -11,7 +11,6 @@
 package org.eclipse.lsp4xml.extensions.xsd.contentmodel;
 
 import java.lang.reflect.Field;
-import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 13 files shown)


angelozerr:
This PR gives the capability to support xs:include schema location:

 * for definition
 * for completion

Test has been written for that. As highlight use the same search code thandefinition and completion, I had to disable search with xs:include for highlight to avoid return the range of xs:element coming from the xs:include XML Schema because LSP highlighting support higlight only for one file and not for external file.

To test this PR, you can use [SchemaA.xsd]( https://github.com/angelozerr/lsp4xml/pull/579/files#diff-09d93288567cbb5bde7afa6c76924d77):

```xml
<xs:schema id="tns" xmlns:xs="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified" attributeFormDefault="unqualified">
	<xs:include schemaLocation="SchemaB.xsd" />
  
	<xs:complexType name="Bar">
		<xs:sequence>
			<xs:element name="FooBar" type="xs:string" />
			<xs:element ref="TypeFromB" />
			<xs:element ref="TypeFromC" />
		</xs:sequence>
	</xs:complexType>

	<xs:complexType name="Root">
		<xs:sequence>
			<xs:element name="Bar" type="Bar" minOccurs="1" maxOccurs="unbounded" />
		</xs:sequence>
	</xs:complexType>

	<xs:element name="Root" type="Root"></xs:element>
</xs:schema>
```
which includes XSD SchemaB.xsd which include SchemaC.xsd.

Here a demo:

XSIncludeDefAndCompletionDemo

NikolasKomonen:
@angelozerr works great, ready to merge!

angelozerr:
Thanks @NikolasKomonen for your review!

PR #584, opened 2019-10-24
improve getCurrentAttribute method for AbstractPositionReqest

Seiphon:


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/AbstractPositionRequest.java
```diff
@@ -101,6 +102,9 @@ public String getCurrentTag() {
 
 	@Override
 	public String getCurrentAttributeName() {
+		if (node != null && node.isAttribute() && ((DOMAttr) node).getName() != null) {
+			return ((DOMAttr) node).getName();
+		}
 		return currentAttributeName;
```
angelozerr: Remove the setCurrentAttributeName method and currentAttributeName field.

angelozerr: Remove the setCurrentAttributeName method and currentAttributeName field.

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/XMLAssert.java
```diff
@@ -578,6 +579,10 @@ public static void assertHover(XMLLanguageService xmlLanguageService, String val
 		}
 	}
 
+	public static void assertHoverAttributeName(IHoverRequest request) {
```
angelozerr: Remove this method test

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/extensions/AggregetedHoverValuesTest.java
```diff
@@ -91,6 +91,7 @@ public String onTag(IHoverRequest request) throws Exception {
 
 			@Override
 			public String onAttributeName(IHoverRequest request) throws Exception {
```
angelozerr: Replace with 

class AggregatedHoverParticipant extends HoverParticipantAdapter {

```java
			@Override
			public String onTag(IHoverRequest request) throws Exception {
				if ("bean".equals(request.getCurrentTag())) {
					return TEST_FOR_TAG_HOVER;
				}
				return null;
			}

			@Override
			public String onAttributeName(IHoverRequest request) throws Exception {
				if ("class".equals(request.getCurrentAttributeName())) {
					return TEST_FOR_ATTRIBUTENAME_HOVER;
				}
				return null;
			}
		}
```

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/AbstractPositionRequest.java
Seiphon: Hi @angelozerr , I am afraid we can not just remove them. If we remove them, it will break something [here ](https://github.com/angelozerr/lsp4xml/blob/4eaea24d41b23dd09d76737483b3557f7bbb71e4/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/ContentModelCompletionParticipant.java#L313).
In this case, the DOMNode can not be recognized as a attribute, so it will always return a null which is not excepted. However if by the original scan way, we can get the attribute name. So maybe we should keep them, if there is no a better way.

angelozerr: I executed our tests by removing currentAttributeName  and everything is working? Have you a usecase which causes the trouble (have you a test for that).

Seiphon: After removed the setCurrentAttributeName method and currentAttributeName field,
https://github.com/angelozerr/lsp4xml/blob/4eaea24d41b23dd09d76737483b3557f7bbb71e4/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/XMLCompletions.java#L90
and 
https://github.com/angelozerr/lsp4xml/blob/4eaea24d41b23dd09d76737483b3557f7bbb71e4/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/XMLCompletions.java#L123
should also be removed.

When I executed tests, completionOnAttributeName() of XMLSchemaCompletionExtensionsTest, and testHTMLAttributeValueCompletion() of HTMLCompletionExtensionsTest will failed.

https://github.com/angelozerr/lsp4xml/blob/4eaea24d41b23dd09d76737483b3557f7bbb71e4/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/extensions/contentmodel/participants/ContentModelCompletionParticipant.java#L313
attributeName will always get a null here.

angelozerr: Ok I see the problem. When completion on attribute value is done, the scanner starts with attribute value token and doesn't get the currentAttributeName. We should remove this currentAttributeName and work only with DOM node because I find it's not very clean to use currentAttributeName when completion is done on attribute name and use node in case . I suggest you to implement `getCurrentAttributeName` like this:

```java
	@Override
	public String getCurrentAttributeName() {
		DOMAttr attr = getCurrentAttribute();
		return attr != null ? attr.getName() : null;
	}

	/**
	 * Returns the current attribute at the given offset and null otherwise.
	 * 
	 * @return the current attribute at the given offset and null otherwise.
	 */
	private DOMAttr getCurrentAttribute() {
		if (node == null) {
			return null;
		}
		switch (node.getNodeType()) {
		case Node.ELEMENT_NODE:
			return node.findAttrAt(offset);
		case Node.ATTRIBUTE_NODE:
			return ((DOMAttr) node);
		default:
			return null;
		}
	}

```

angelozerr:
Thanks @Seiphon !

PR #598, opened 2019-12-10
Allows File Associations to be used without Workspace

NikolasKomonen:
Fixes #redhat-developer/vscode-xml/issues/202

Signed-off-by: Nikolas Komonen <nikolaskomonen@gmail.com>


NikolasKomonen:
Previously, the systemId was only handled if the rootUri was present from the workspace. Now, it doesn't require a workspace and will work with an absolute systemId

PR #608, opened 2020-02-06
Unregister language server extension on LS shutdown

### LINKED ISSUE #605: Unregister language server extension on LS shutdown
For the LemMinX/lsp4xml maven extension testing suite, it's required that `IXMLExtension.stop()` be called when the LS gets shutdown. In our case, file locks for Maven index files get released when the extension is unregistered. However, currently the extensions aren't unregistered when the server is shutdown.

Unless there is a suitable workaround, having XML LS extensions get unregistered on LS shutdown would be greatly appreciated.

AObuchow:
fix #605

### DIFF
--- a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/XMLExtensionsRegistry.java
+++ b/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/XMLExtensionsRegistry.java
@@ -87,7 +87,14 @@ public class XMLExtensionsRegistry implements IComponentProvider {
 
 	public void initializeParams(InitializeParams params) {
 		if (initialized) {
-			extensions.stream().forEach(extension -> extension.start(params, this));
+			extensions.stream().forEach(extension -> {
+				try {
+					extension.start(params, this);
+				} catch (Exception e) {
+					LOGGER.log(Level.SEVERE, "Error while starting extension <" + extension.getClass().getName() + ">",
+							e);
+				}
+			});
 		} else {
 			this.params = params;
 		}
@@ -169,7 +176,7 @@ public class XMLExtensionsRegistry implements IComponentProvider {
 	}
 
 	private synchronized void initialize() {
-		
+
 		if (initialized) {
 			return;
 		}
@@ -194,8 +201,22 @@ public class XMLExtensionsRegistry implements IComponentProvider {
 	}
 
 	void unregisterExtension(IXMLExtension extension) {
-		extensions.remove(extension);
-		extension.stop(this);
+		try {
+			extensions.remove(extension);
+			extension.stop(this);
+		} catch (Exception e) {
+			LOGGER.log(Level.SEVERE, "Error while stopping extension <" + extension.getClass().getName() + ">", e);
+		}
+	}
+
+	/**
+	 * Unregisters all registered extensions.
+	 */
+	public void dispose() {
+		// Copy the list of extensions to avoid ConcurrentModificationError
+		List<IXMLExtension> extensionReferences = new ArrayList<>();
+		extensions.forEach(extensionReferences::add);
+		extensionReferences.forEach(this::unregisterExtension);
 	}
 
 	public void registerCompletionParticipant(ICompletionParticipant completionParticipant) {
--- a/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/extensions/ExtensionRegistryDisposeTest.java
+++ b/org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/extensions/ExtensionRegistryDisposeTest.java
@@ -1,0 +1,62 @@
+/*******************************************************************************
+* Copyright (c) 2020 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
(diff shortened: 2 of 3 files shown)


org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/XMLExtensionsRegistry.java
```diff
@@ -197,6 +197,12 @@ void unregisterExtension(IXMLExtension extension) {
 		extensions.remove(extension);
 		extension.stop(this);
 	}
+	
+	public void disposeExtensions() {
```
angelozerr: Please add javadoc which explains it unregister extension and please rename this method to dispose only because in the future we could dispose other things than extension.

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/XMLExtensionsRegistry.java
```diff
@@ -197,6 +197,12 @@ void unregisterExtension(IXMLExtension extension) {
 		extensions.remove(extension);
 		extension.stop(this);
 	}
+	
+	public void disposeExtensions() {
+		for (IXMLExtension extension : extensions) {
+			unregisterExtension(extension);
```
angelozerr: Not tested but are you sure it works since unregisterEXtension remove the extension? Please add tests which create XML extension registry add some extension and callb dispose. You could check the state of extension like stopped and check there are none extensions

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/XMLExtensionsRegistry.java
AObuchow: Okay will do

org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/XMLExtensionsRegistry.java
AObuchow: Ok no problem

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/extensions/ExtensionRegistryDisposeTest.java
```diff
@@ -0,0 +1,45 @@
+package org.eclipse.lsp4xml.services.extensions;
```
angelozerr: Add header

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/extensions/ExtensionRegistryDisposeTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import static org.junit.Assert.assertTrue;
+
+import org.eclipse.lsp4j.InitializeParams;
+import org.eclipse.lsp4xml.services.extensions.save.ISaveContext;
+import org.junit.Test;
+
+public class ExtensionRegistryDisposeTest {
```
angelozerr: Add javadoc

AObuchow: Will do

org.eclipse.lsp4xml/src/test/java/org/eclipse/lsp4xml/services/extensions/ExtensionRegistryDisposeTest.java
AObuchow: Good catch

angelozerr:
Last thing to do, do try/cacth inside unregisterExtension like it's done in registerExtension to be sure to call stop for each extension even if there is an extension which fails.

https://github.com/angelozerr/lsp4xml/blob/1057f73f1d29c91b0bcb20a1c78ae72e03d8c21a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/XMLExtensionsRegistry.java#L90

I notice that in https://github.com/angelozerr/lsp4xml/blob/1057f73f1d29c91b0bcb20a1c78ae72e03d8c21a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/XMLExtensionsRegistry.java#L90 no try catch is done, please do it.

Thanks!

AObuchow:
> Last thing to do, do try/cacth inside unregisterExtension like it's done in registerExtension to be sure to call stop for each extension even if there is an extension which fails.
> 
> https://github.com/angelozerr/lsp4xml/blob/1057f73f1d29c91b0bcb20a1c78ae72e03d8c21a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/XMLExtensionsRegistry.java#L90
> 
> I notice that in
> 
> https://github.com/angelozerr/lsp4xml/blob/1057f73f1d29c91b0bcb20a1c78ae72e03d8c21a/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/XMLExtensionsRegistry.java#L90
> 
> no try catch is done, please do it.
> Thanks!

Ok, it's done

angelozerr:
It's perfect @AObuchow, thanks!

AObuchow:
@angelozerr thanks for the review! Now the Maven extension's test should be happy :)

angelozerr:
> @angelozerr thanks for the review! 

You are welcome!

> Now the Maven extension's test should be happy :)

Nice :)

Issue #613, opened 2020-03-05
NPE when document contains an empty tag


mickaelistria:
Trying to complete
```
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
	<modelVersion>4.0.0</modelVersion>

	<groupId>org.test</groupId>
	<artifactId>test</artifactId>
	<version>0.0.1-SNAPSHOT</version>
	<packaging>pom</packaging>
	
	<build>
		<plugins>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-dependency-plugin</artifactId>
				<version>3.1.1</version>
				<executions>
					<execution>
						<goals><goal>list</goal></goals>
						<configuration>
							<>
						</configuration>
					</execution>
				</executions>
			</plugin>
		</plugins>
	</build>
</project>
```
in the empty tag inside `configuration`, I get
```
SEVERE: While performing ICompletionParticipant#onTagOpen
java.lang.IllegalArgumentException: Property must not be null: label
	at org.eclipse.lsp4j.util.Preconditions.checkNotNull(Preconditions.java:29)
	at org.eclipse.lsp4j.CompletionItem.<init>(CompletionItem.java:136)
	at org.eclipse.lsp4xml.extensions.contentmodel.participants.ContentModelCompletionParticipant.addTagName(ContentModelCompletionParticipant.java:204)
	at org.eclipse.lsp4xml.extensions.contentmodel.participants.ContentModelCompletionParticipant.addTagName(ContentModelCompletionParticipant.java:212)
	at org.eclipse.lsp4xml.extensions.contentmodel.participants.ContentModelCompletionParticipant.addTagName(ContentModelCompletionParticipant.java:212)
	at org.eclipse.lsp4xml.extensions.contentmodel.participants.ContentModelCompletionParticipant.addTagName(ContentModelCompletionParticipant.java:212)
	at org.eclipse.lsp4xml.extensions.contentmodel.participants.ContentModelCompletionParticipant.addTagName(ContentModelCompletionParticipant.java:212)
	at org.eclipse.lsp4xml.extensions.contentmodel.participants.ContentModelCompletionParticipant.addTagName(ContentModelCompletionParticipant.java:212)
[...]

PR #620, opened 2020-03-12
In Maven <configuration>, all known XML elements from schema are suggested as completion

### LINKED ISSUE #612: In Maven <configuration>, all known XML elements from schema are suggested as completion
In Maven <configuration>, all known XML elements from schema are suggested as completion.
This is typically not desired as the Maven extension can provide higher quality proposals here and the extra items are useless and generate noise reducing the usability.

angelozerr:
In Maven <configuration>, all known XML elements from schema are suggested as completion

Fixes #612

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCompletionParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCompletionParticipant.java
@@ -30,6 +30,7 @@ import org.eclipse.lemminx.services.extensions.ICompletionRequest;
 import org.eclipse.lemminx.services.extensions.ICompletionResponse;
 import org.eclipse.lemminx.settings.XMLFormattingOptions;
 import org.eclipse.lemminx.uriresolver.CacheResourceDownloadingException;
+import org.eclipse.lemminx.utils.StringUtils;
 import org.eclipse.lsp4j.CompletionItem;
 import org.eclipse.lsp4j.CompletionItemKind;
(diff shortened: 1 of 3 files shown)


angelozerr:
This PR fixes too https://github.com/eclipse/lemminx/issues/613 even if we are not in this case since skip has the same behavior than strict.

angelozerr:
To test this PR, create a pom file like this and open the completion (see | inside configuration element):

```xml
3.
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
	<modelVersion>4.0.0</modelVersion>

	<groupId>org.test</groupId>
	<artifactId>test</artifactId>
	<version>0.0.1-SNAPSHOT</version>
	<packaging>pom</packaging>
	
	<build>
		<plugins>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-dependency-plugin</artifactId>
				<version>3.1.1</version>
				<executions>
					<execution>
						<goals><goal>list</goal></goals>
						<configuration>
							<|>
						</configuration>
					</execution>
				</executions>
			</plugin>
		</plugins>
	</build>
</project>
```

You should see just project root element instead of all elements from XML Schema.

image

@mickaelistria please review it.

mickaelistria:
Sorry, I cannot easily try lemminx at the moment as we cannot migrate the extension before LemMinX has a 1st release.
I trust you there, and if you think it's OK to merge, don't wait for me.

PR #621, opened 2020-03-12
Add support for hovering on tag text

### LINKED ISSUE #609: Add onTagText to IHoverParticipant (or extend IHoverParticipant)
For the LemMinX/lsp4xml Maven extension, I would like to add support for hover request when hovering over text within a tag. Currently, [IHoverParticipant](https://github.com/angelozerr/lsp4xml/blob/68d929216d7d459cca1309f0a23eaad09ca825ca/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/services/extensions/IHoverParticipant.java) only declares hovers for `onTag`, `onAttributeName` and `onAttributeValue`. 
Since the interface is already in use, perhaps a `IHoverParticipantExtension` should be added, which would declare `onTagText`. This function would only be called when the child of a node contains only text content (and not another node/tag).

This is required for https://github.com/AObuchow/lsp4xml-extensions-maven/issues/30 as well as other use cases. For a demo of the hover we are trying to implement, [see this GIF.](https://photos.app.goo.gl/Myy3yBwgnUG4aJVU6)

AObuchow:
Add onTagText to IHoverParticipant, fix #609

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLHover.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLHover.java
@@ -22,6 +22,7 @@ import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.dom.DOMText;
 import org.eclipse.lemminx.dom.parser.Scanner;
 import org.eclipse.lemminx.dom.parser.TokenType;
 import org.eclipse.lemminx.dom.parser.XMLScanner;
@@ -84,8 +85,34 @@ class XMLHover {
 			}
 			// Attribute is hover
 			return getAttrNameHover(hoverRequest, null);
+		} else if (node.isText()) {
+			// Text is hover
+			DOMText text = (DOMText) node;
+			try {
+				Range textRange = new Range(xmlDocument.positionAt(text.getStart()),
+						xmlDocument.positionAt(text.getEnd()));
+				return getTextHover(hoverRequest, textRange);
+			} catch (BadLocationException e) {
+				LOGGER.log(Level.SEVERE, "While creating Range in XMLHover the text's Offset was a BadLocation", e);
+			}
 		}
 		return null;
+	}
+
+	private Hover getTextHover(HoverRequest hoverRequest, Range textRange) {
+		hoverRequest.setTagRange(textRange);
+		List<String> contentValues = new ArrayList<String>();
+		for (IHoverParticipant participant : extensionsRegistry.getHoverParticipants()) {
+			try {
+				String contentValue = participant.onText(hoverRequest);
+				if (contentValue != null) {
+					contentValues.add(contentValue);
+				}
+			} catch (Exception e) {
+				LOGGER.log(Level.SEVERE, "While performing IHoverParticipant#onTagText", e);
+			}
+		}
+		return createHover(contentValues, hoverRequest);
 	}
 
 	/**
@@ -111,7 +138,7 @@ class XMLHover {
 				LOGGER.log(Level.SEVERE, "While performing IHoverParticipant#onTag", e);
 			}
 		}
-		return creatHover(contentValues, hoverRequest);
+		return createHover(contentValues, hoverRequest);
 	}
 
 	private Range getTagNameRange(TokenType tokenType, int startOffset, int offset, DOMDocument document) {
@@ -154,7 +181,7 @@ class XMLHover {
 				LOGGER.log(Level.SEVERE, "While performing IHoverParticipant#onTag", e);
 			}
 		}
-		return creatHover(contentValues, hoverRequest);
+		return createHover(contentValues, hoverRequest);
 	}
 
 	/**
@@ -178,7 +205,7 @@ class XMLHover {
 				LOGGER.log(Level.SEVERE, "While performing IHoverParticipant#onTag", e);
 			}
 		}
-		return creatHover(contentValues, hoverRequest);
+		return createHover(contentValues, hoverRequest);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/IHoverParticipant.java
```diff
@@ -41,5 +41,13 @@
...
 	 */
 	String onAttributeValue(IHoverRequest request) throws Exception;
+	
+	/**
+	 * onTagText method
+	 *
+	 * @param hoverRequest the hover request.
+	 * @return the Value of MarkupContent {@link String}
+	 */
+	String onTagText(IHoverRequest resquest) throws Exception;
```
AObuchow: Should this maybe be renamed to "onText"? What do you think @angelozerr

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLHover.java
```diff
@@ -188,7 +215,7 @@ private Hover getAttrValueHover(HoverRequest hoverRequest, Range attrRange) {
 	 * @param hoverRequest the hover request.
 	 * @return the aggregated LSP hover from the value list.
 	 */
-	private static Hover creatHover(List<String> contentValues, HoverRequest hoverRequest) {
+	private static Hover createHover(List<String> contentValues, HoverRequest hoverRequest) {
```
AObuchow: This was just a typo I found, not related to the objective of this PR.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/IHoverParticipant.java
angelozerr: Yes onText is better (it's a w3c DOM Text)

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLHover.java
angelozerr: Please change that is this same PR. Thanks for having see that.

AObuchow: Ok so the rename is okay?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/IHoverParticipant.java
AObuchow: Ok will update it to onText

angelozerr:
@AObuchow your PR looks promising, thanks! Do you think it could be possible for you to manage hover text for XML Schema? I think it can be a good usecase for maven, I mean :

 * text from the maven XML Schema description
 * text from your maven extension.

What do you think? This feature should be implemented in ContentModelHoverParticipant.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/IHoverParticipant.java
angelozerr: Please implement this method in HoverParticipantAdapter

angelozerr: Thanks

AObuchow:
> What do you think? This feature should be implemented in ContentModelHoverParticipant.

I have no issues with trying this :) I'll give it a try in a moment.

angelozerr:
> I have no issues with trying this :) I'll give it a try in a moment.

I don't know if maven provides that, so perhaps my idea is bad. However a thing which is more important than XML Schema text hover is to write tests.

I suggest that you add test in https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/HTMLHoverExtensionsTest.java

You could implement onText in https://github.com/eclipse/lemminx/blob/92256a1e1b9c5c8d1e51202271af49cf7cb8b48d/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/HTMLHoverExtensionsTest.java#L63 and write some tests. So please work on that.

If XML Schema onText is easy to implement, do it in this same PR otherwise, please give up this idea and we will see that in an another PR. Thanks!

AObuchow:
@angelozerr I added some tests. 

I don't have enough immediate knowledge on XML schema definitions/content model to quickly get it working for XML schema (unless I just copy the implementation of onTag from ContentModelHoverParticipant). So it might be better to do that in a seperate PR.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/HTMLHoverExtensionsTest.java
```diff
@@ -39,6 +39,9 @@ public void testSingle() throws BadLocationException {
 		assertHover("<html></htm|l>", "</html>", 8);
 		assertHover("<html></html|>", "</html>", 8);
 		assertHover("<html></html>|");
+		assertHover("<html>hover|Text</html>", "hoverText", 6);
+		assertHover("<html>h|overText</html>", "hoverText", 6);
+		assertHover("<html> |</html>", " ", 6);
```
angelozerr: Please add test with empty text

angelozerr:
No problem with XML schema text. We will see that in an another pr. Please add last test that i suggested you to write and After that lets go for the merge

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/HTMLHoverExtensionsTest.java
AObuchow: When there's no text (eg.` assertHover("<html>|</html>")`) the `DOMNode` is considered to be a DOMElement (not `DOMText`). In other words, [this if statement](https://github.com/eclipse/lemminx/pull/621/files/353b25cec17c55eb9572786bb15968f3e3b86980#diff-b8239e61a89785e705c99fc4ee5fc4a1R66) evaluates to true. 

Since we are already doing a test for empty text [here](https://github.com/eclipse/lemminx/pull/621/files/353b25cec17c55eb9572786bb15968f3e3b86980#diff-a769ff934bc2fac796856d092dc07122R34), I don't know if there's a way to test `onText` hover with empty text.

angelozerr:
Thanks @AObuchow !

PR #632, opened 2020-04-13
xsd:enumeration autocomplete don't work for text node

angelozerr:
xsd:enumeration autocomplete don't work for text node

Fixes https://github.com/redhat-developer/vscode-xml/issues/218

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/SnippetsBuilder.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/SnippetsBuilder.java
@@ -46,9 +46,25 @@ public class SnippetsBuilder {
 	}
 
 	/**
+	 * Returns the LSP choices snippets content.
 	 * 
-	 * @param index
-	 * @param values
+	 * @param index the snippet index.
+	 * @param values the values for the choice.
+	 * @return the LSP choices snippets content.
+	 * 
+	 * @see https://github.com/Microsoft/language-server-protocol/blob/master/snippetSyntax.md#choice
+	 */
+	public static String choice(int index, Collection<String> values) {
+		StringBuilder snippets = new StringBuilder();
+		choice(index, values, snippets);
+		return snippets.toString();
+	}
+
+	/**
+	 * Add LSP choices snippets in the given snippets content.
+	 * 
+	 * @param index the snippet index.
+	 * @param values the values for the choice.
 	 * @return
 	 * 
 	 * @see https://github.com/Microsoft/language-server-protocol/blob/master/snippetSyntax.md#choice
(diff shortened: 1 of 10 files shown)


angelozerr:
To test this PR, see usecase at https://github.com/redhat-developer/vscode-xml/issues/218#issue-547209289

 * when you apply completion to insert `ACTION-TYPE`, it should show choice of xsd:enumeration
 * when you have `<ACTION-TYPE>|<ACTION-TYPE>` it should show list of xsd:enumeration.

Here a demo which shows in action xs:enumeration on text:

XSDEnumerationOnText

fbricon:
Looks promising. Could you plug the "did you mean" feature in? I don't remember if we implemented it in lemminx

angelozerr:
This PR is ready (there are some tests), the only part which is not managed is the documentation for the xs:enumeration value (see the FIXME).

I don't know how to manage that for the moment. I think we should create an another issue to manage that and support hover together.

angelozerr:
> Looks promising. Could you plug the "did you mean" feature in? I don't remember if we implemented it in lemminx

What do ypou mean with  `Could you plug the "did you mean" feature in?`

> I don't remember if we implemented it in lemminx

No it was not implemented and my PR implements that.

xorye:
I came across some bugs:
On my machine, there are three failing tests:
```
[ERROR] Failures: 
[ERROR]   DTDCompletionExtensionsTest.testNoDuplicateCompletionItems:186->testCompletionFor:202 expected: <5> but was: <2>
[ERROR]   DTDCompletionWithCacheExtensionsTest.dtdCache:70 desc should only exist once: Actual: End with '</animate>',End with '</svg>',#region,#endregion,cdata,comment ==> expected: <1> but was: <0>
[ERROR]   XMLSchemaCompletionExtensionsTest.noNamespaceSchemaLocationCompletion:223 expected: <<OutOfBand></OutOfBand>> but was: <<OutOfBand>false</OutOfBand>>
```

Also, the completion and validation does not seem to work after I open AUTOSAR_00048.xsd:
demo

angelozerr:
> Also, the completion and validation does not seem to work after I open AUTOSAR_00048.xsd

Yes please don't open it:) The problem with this XML Schema is that it generates too many symbols (you can exclude the symbols for this files in the settings), but I'm glad that you have seen the problem and now I think you will understand why https://github.com/redhat-developer/vscode-xml/issues/237 is very important :)

angelozerr:
> On my machine, there are three failing tests:

That's strange, travis doesnt fail?

angelozerr:
@xorye please retry

xorye:
It works!

xorye:
> Yes please don't open it:) The problem with this XML Schema is that it generates too many symbols (you can exclude the symbols for this files in the settings), but I'm glad that you have seen the problem and now I think you will understand why redhat-developer/vscode-xml#237 is very important :)

I see. I still get this problem when I disable symbols/exclude the file. I can investigate more while taking care of redhat-developer/vscode-xml#237

PR #634, opened 2020-04-14
MSG_ATTRIBUTE_NOT_DECLARED must highlight attribute name instead of attribute value

angelozerr:
MSG_ATTRIBUTE_NOT_DECLARED must highlight attribute name instead of attribute value.


angelozerr:
Given this XML file:

```xml
<!DOCTYPE note [
	<!ELEMENT note (to,from,heading,body, note?)>
	<!ATTLIST note version CDATA #REQUIRED>

]>
<note XXXXXX="0.9.0" >  
</note>
```

`XXXXXX` attribute `name` should be highlighted as error. In master it's the `XXXXXX` attribute `value` which is highlighted which is wrong.

xorye:
I have tested this PR, this PR can be merged

angelozerr:
Thanks @xorye

Issue #67, opened 2018-08-18
Give more settings for Format


angelozerr:
It should be cool to give the capability to customize formatting. We need to collect which settings we need and implement the formatting with thoses settings.

Here settings for WTP/XML Editor:

![image](https://user-images.githubusercontent.com/1932211/44298666-6165b280-a2e7-11e8-8902-fb1417934bb8.png)

Here the HTML formatting option for  vscode html language service https://github.com/Microsoft/vscode-html-languageservice/blob/master/src/services/htmlFormatter.ts#L55

```ts
let htmlOptions: IBeautifyHTMLOptions = {
		indent_size: options.insertSpaces ? tabSize : 1,
		indent_char: options.insertSpaces ? ' ' : '\t',
		wrap_line_length: getFormatOption(options, 'wrapLineLength', 120),
		unformatted: getTagsFormatOption(options, 'unformatted', void 0),
		content_unformatted: getTagsFormatOption(options, 'contentUnformatted', void 0),
		indent_inner_html: getFormatOption(options, 'indentInnerHtml', false),
		preserve_newlines: getFormatOption(options, 'preserveNewLines', true),
		max_preserve_newlines: getFormatOption(options, 'maxPreserveNewLines', 32786),
		indent_handlebars: getFormatOption(options, 'indentHandlebars', false),
		end_with_newline: includesEnd && getFormatOption(options, 'endWithNewline', false),
		extra_liners: getTagsFormatOption(options, 'extraLiners', void 0),
		wrap_attributes: getFormatOption(options, 'wrapAttributes', 'auto'),
		wrap_attributes_indent_size: getFormatOption(options, 'wrapAttributesIndentSize', void 0),
		eol: '\n'
	};
```

@NikolasKomonen do you want to work on this issue?

NikolasKomonen:
@angelozerr sure, I'll start on it

angelozerr:
Thanks @NikolasKomonen but before starting develop something, please clarify which preferences you wish to support (by seeing samples that I have given below).

NikolasKomonen:
@angelozerr I will implement all the options on the Eclipse editor settings in the image you sent. After I will look and see if anything else should be added.

angelozerr:
And what about VSCode HTML Language Service?

```
et htmlOptions: IBeautifyHTMLOptions = {
		indent_size: options.insertSpaces ? tabSize : 1,
		indent_char: options.insertSpaces ? ' ' : '\t',
[...]

PR #649, opened 2020-04-21
Insert final newline depending on lsp4j formatting settings

xorye:
First step to fix https://github.com/redhat-developer/vscode-xml/issues/196

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
@@ -61,6 +61,7 @@ class XMLFormatter {
 		private DOMDocument rangeDomDocument;
 		private XMLBuilder xmlBuilder;
 		private int indentLevel;
+		private boolean linefeedOnNextWrite;
 
 		/**
 		 * XML formatter document.
@@ -70,6 +71,7 @@ class XMLFormatter {
 			this.range = range;
 			this.options = options;
 			this.emptyElements = options.getEmptyElements();
+			this.linefeedOnNextWrite = false;
 		}
 
 		/**
@@ -261,14 +263,15 @@ class XMLFormatter {
 
 		private void format(DOMNode node) throws BadLocationException {
 
+			if (linefeedOnNextWrite && (!node.isText() || !((DOMText) node).isWhitespace())) {
+				this.xmlBuilder.linefeed();
+				linefeedOnNextWrite = false;
+			}
+
 			if (node.getNodeType() != DOMNode.DOCUMENT_NODE) {
-				boolean doLineFeed;
-				if (node.getOwnerDocument().isDTD()) {
-					doLineFeed = false;
-				} else {
-					doLineFeed = !(node.isComment() && ((DOMComment) node).isCommentSameLineEndTag())
-							&& (!node.isText() || (!((DOMText) node).isWhitespace() && ((DOMText) node).hasSiblings()));
-				}
+				boolean doLineFeed = !node.getOwnerDocument().isDTD() &&
+						!(node.isComment() && ((DOMComment) node).isCommentSameLineEndTag()) &&
+						(!node.isText() || (!((DOMText) node).isWhitespace() && ((DOMText) node).hasSiblings()));
 
 				if (this.indentLevel > 0 && doLineFeed) {
 					// add new line + indent
@@ -317,12 +320,12 @@ class XMLFormatter {
 
 		/**
 		 * Format the given DOM prolog
-		 * 
+		 *
 		 * @param node the DOM prolog to format.
 		 */
 		private void formatProlog(DOMNode node) {
 			addPrologToXMLBuilder(node, this.xmlBuilder);
-			this.xmlBuilder.linefeed();
+			linefeedOnNextWrite = true;
 		}
 
 		/**
@@ -331,10 +334,13 @@ class XMLFormatter {
 		 * @param textNode the DOM text node to format.
 		 */
 		private void formatText(DOMText textNode) {
-			// Generate content
 			String content = textNode.getData();
-			xmlBuilder.addContent(content, textNode.isWhitespace(), textNode.hasSiblings(), textNode.getDelimiter(),
-					this.indentLevel);
+			if (textNode.equals(this.fullDomDocument.getLastChild())) {
+				xmlBuilder.addContent(content);
+			} else {
+				xmlBuilder.addContent(content, textNode.isWhitespace(), textNode.hasSiblings(),
+						textNode.getDelimiter());
+			}
 		}
 
 		/**
@@ -363,7 +369,8 @@ class XMLFormatter {
 				if (documentType.isClosed()) {
 					xmlBuilder.endDoctype();
 				}
-				xmlBuilder.linefeed();
+				linefeedOnNextWrite = true;
+
 			} else {
 				formatDTD(documentType, 0, this.endOffset, this.xmlBuilder);
 			}
@@ -373,7 +380,7 @@ class XMLFormatter {
 		 * Format the given DOM ProcessingIntsruction.
 		 * 
 		 * @param element the DOM ProcessingIntsruction to format.
-		 * 
+		 *
 		 */
 		private void formatProcessingInstruction(DOMNode node) {
 			addPIToXMLBuilder(node, this.xmlBuilder);
@@ -386,14 +393,14 @@ class XMLFormatter {
 		 * Format the given DOM Comment
 		 * 
 		 * @param element the DOM Comment to format.
-		 * 
+		 *
 		 */
 		private void formatComment(DOMComment comment) {
 			this.xmlBuilder.startComment(comment);
 			this.xmlBuilder.addContentComment(comment.getData());
 			this.xmlBuilder.endComment();
 			if (this.indentLevel == 0) {
-				this.xmlBuilder.linefeed();
+				linefeedOnNextWrite = true;
 			}
 		}
 
@@ -616,6 +623,19 @@ class XMLFormatter {
 			Position endPosition = this.textDocument.positionAt(this.endOffset);
 			Range r = new Range(startPosition, endPosition);
 			List<TextEdit> edits = new ArrayList<>();
+
+			// check if format range reaches the end of the document
+			if (this.endOffset == this.textDocument.getText().length()) {
+
+				if (options.isTrimFinalNewlines()) {
+					this.xmlBuilder.trimFinalNewlines();
+				}
+
+				if (this.options.isInsertFinalNewline() && !this.xmlBuilder.isLastLineEmptyOrWhitespace()) {
+					this.xmlBuilder.linefeed();
+				}
+			}
+
 			edits.add(new TextEdit(r, this.xmlBuilder.toString()));
 			return edits;
 		}
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLBuilder.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLBuilder.java
@@ -79,7 +79,7 @@ public class XMLBuilder {
 			xml.append(":");
 		}
 		xml.append(name);
-		if(isEndTagClosed) {
+		if (isEndTagClosed) {
 			xml.append(">");
 		}
 		return this;
@@ -107,18 +107,18 @@ public class XMLBuilder {
 	 * 
 	 * It will not perform any linefeeds and only basic indentation.
 	 * 
-	 * @param name attribute name
+	 * @param name  attribute name
 	 * @param value attribute value
 	 * @return
 	 */
 	public XMLBuilder addSingleAttribute(String name, String value, boolean surroundWithQuotes) {
 		xml.append(" ");
 		addAttributeContents(name, true, value, surroundWithQuotes);
-		
+
 		return this;
 	}
 
-	public XMLBuilder addAttributesOnSingleLine(DOMAttr attr, Boolean surroundWithQuotes) {
+	public XMLBuilder addAttributesOnSingleLine(DOMAttr attr, boolean surroundWithQuotes) {
 		xml.append(" ");
 		addAttributeContents(attr.getName(), attr.hasDelimiter(), attr.getValue(), surroundWithQuotes);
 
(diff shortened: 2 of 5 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
```diff
@@ -616,6 +616,15 @@ private boolean hasSingleAttributeInFullDoc(DOMElement element) {
 			Position endPosition = this.textDocument.positionAt(this.endOffset);
 			Range r = new Range(startPosition, endPosition);
 			List<TextEdit> edits = new ArrayList<>();
+
+			// check if format range reaches the end of the document
+			if (this.endOffset == this.textDocument.getText().length()) {
+				this.xmlBuilder.trimTrailingNewlines();
```
angelozerr, line 622: I think we should do that only if  if `options.isTrimFinalNewlines()` is equals to true

xorye: I'm not sure if I understand, are you suggesting:

```
if (options.isTrimFinalNewlines() && this.endOffset == this.textDocument.getText().length()) {
    this.xmlBuilder.trimTrailingNewlines();
```
?

angelozerr: No I mean 

```java
if (this.endOffset == this.textDocument.getText().length()) {
    if (options.isTrimFinalNewlines()) {
        this.xmlBuilder.trimTrailingNewlines();
    }
    ...
```

xorye: @angelozerr ok thank you.

Just to clarify, the `xml.format.preservedNewlines` should only be for newlines in between nodes? Therefore `xml.format.preservedNewlines` shouldn't affect the newlines at the end of an XML document?

xorye: I discovered that the root cause of https://github.com/redhat-developer/vscode-xml/issues/196 is a small bug with how `xml.format.preservedNewlines` is being used

xorye: I think it shouldn't affect the newlines at the end of the document because in the description:
image
it mentions "new lines that separate tags"

xorye:
In my latest commit, the code now trims the final newline only if the lsp4j trimFinalNewlines option is true. With vscode-xml, the option is currently null, so `FormattingOptions#isTrimFinalNewlines` returns false.

Some tests are failing right now, partly because the current code does not trim final spaces. (I think the code should always trim final spaces? If yes, I will go ahead and implement that)

For example, if `_` were spaces:
before formatting
```
<a           ></a>_____
```

after formatting
```
<a></a>_____
```

xorye:
And for cases where `FormattingOptions#isTrimFinalNewlines` is false, I think we should make sure to only trim final spaces, not final newlines.

Like so:

Before
```
<a      ></a>
__________\n
____\n
_____\n
```

After
```
<a></a>
\n
\n
\n
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
angelozerr: To be honest with you, I don't know this formatting option. This newline have started with the issue https://github.com/eclipse/lemminx/issues/67 So I suggest that you play with HTML language server to see in action the preserve newlines behavior and try to do the same thing for XML

> I discovered that the root cause of redhat-developer/vscode-xml#196 is a small bug with how xml.format.preservedNewlines is being used

Is it easy to fix it?

xorye:
The latest commit makes the formatting settings/behaviour be more consistent with HTML's

HTML's "preserve new lines" settings preserve newlines in between tags, therefore does not affect the newlines at the end of the document whatsoever. This PR implements that behaviour, the `xml.format.preservedNewlines` preference now only applies to newlines in between tags.

In HTML, there is no HTML setting to keep more than 1 newline at the end of the document after formatting. Either there are no newlines, or there are only one.
This can be configured using the `html.format.endWithNewline` setting.

To have this in vscode-xml, we should introduce a `xml.format.endWithNewline` preference, and maybe have the setting be passed into the formatting request, so that it is accessible from the lsp4j FormattingOptions.

So as it stands right now, to keep consistency with HTML, I think it makes sense to always trim final lines at the end, and only add the newline if `xml.format.endWithNewline` is true.

All of this is (with the exception of the vscode-xml preference, this would be handled in a PR in vscode-xml) implemented in the latest commit of this PR

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
xorye: Yes, my fix was to not use `xml.format.preservedNewlines` if the current textnode was the last node of the document:
https://github.com/eclipse/lemminx/blob/4fd0d5c99f8373847a7242340f07e48e0620c2d9/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLBuilder.java#L240

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
```diff
@@ -656,6 +674,25 @@ private static void addAttributes(DOMNode node, XMLBuilder xmlBuilder) {
...
+
+		/**
+		 * Returns true if <code>node</code> is the last non-whitespace
+		 * child of the node's owner document, and false otherwise
+		 * 
+		 * @param node the node to check
+		 * @return true if <code>node</code> is the last non-whitespace
+		 * child of the node's owner document, and false otherwise
+		 */
+		private static boolean isLastNonWhitespaceChild(DOMNode node) {
```
xorye, line 686: The reason why I have this, is to make sure that the code doesn't add new lines after the last node. BUT this is not needed if we are going to trim the end of the document anyways.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
```diff
@@ -616,6 +623,19 @@ private boolean hasSingleAttributeInFullDoc(DOMElement element) {
 			Position endPosition = this.textDocument.positionAt(this.endOffset);
 			Range r = new Range(startPosition, endPosition);
 			List<TextEdit> edits = new ArrayList<>();
+
+			// check if format range reaches the end of the document
+			if (this.endOffset == this.textDocument.getText().length()) {
+				
+				if (options.isTrimFinalNewlines()) {
```
xorye, line 630: To be consistent with HTML, this should always be true

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
xorye: In other words, if `options.isTrimFinalNewlines()` is always true, we don't need this method. 

If `options.isTrimFinalNewlines()` is false, we need this method.

angelozerr: @xorye if we could avoid doing that, it should be very nice, because it means that for each nodes that you wish to format, you need to loop for the DOM tree.

Today we call `this.xmlBuilder.linefeed();`  after the generate of processing instruction, prolog, etc

```java
private void formatProlog(DOMNode node) {
	addPrologToXMLBuilder(node, this.xmlBuilder);
	this.xmlBuilder.linefeed();
}
```

I wonder if it should be better to flag the xml builder to say "when you will write next content, add a new line before).

```java
private void formatProlog(DOMNode node) {
	addPrologToXMLBuilder(node, this.xmlBuilder);
	this.xmlBuilder.linefeedOnNextWrite();
}
```

and when this.xmlBuilder must be updated with new content a linefedd will be inserted?

xorye: Understood, fixed

xorye:
With this vscode-xml PR: https://github.com/redhat-developer/vscode-xml/pull/248 the `files.trimFinalNewlines` and `files.insertFinalNewline` vscode settings get mapped to `xml.format.trimFinalNewlines` and `xml.format.insertFinalNewline` respectively.

Because of this line: https://github.com/xorye/lsp4xml/blob/ea1669063effe5d842abcc11405b4613354c7179/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java#L354

If the formatting options from the formatting request has the `insertFinalNewline`or `trimFinalNewlines` setting, they will override `xml.format.insertFinalNewline`/`xml.format.trimFinalLines`.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLBuilder.java
```diff
@@ -170,21 +170,25 @@ public XMLBuilder addAttribute(DOMAttr attr, int level, boolean surroundWithQuot
 	/**
 	 * Builds the attribute {name, '=', and value}.
 	 * 
-	 * Never puts quotes around unquoted values unless indicated to by 'surroundWithQuotes'
+	 * Never puts quotes around unquoted values unless indicated to by
+	 * 'surroundWithQuotes'
 	 */
-	private void addAttributeContents(String name, Boolean equalsSign, String originalValue, boolean surroundWithQuotes) {
-		if(name != null) {
+	private void addAttributeContents(String name, Boolean equalsSign, String originalValue,
```
angelozerr, line 176: Replace all `Boolean `from this XMLBuilder class with `boolean`.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLBuilder.java
```diff
@@ -170,21 +170,25 @@ public XMLBuilder addAttribute(DOMAttr attr, int level, boolean surroundWithQuot
 	/**
 	 * Builds the attribute {name, '=', and value}.
 	 * 
-	 * Never puts quotes around unquoted values unless indicated to by 'surroundWithQuotes'
+	 * Never puts quotes around unquoted values unless indicated to by
+	 * 'surroundWithQuotes'
 	 */
```
angelozerr, line 175: Add Javadoc for each parameters.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFormatterTest.java
```diff
@@ -2055,6 +2054,150 @@ public void testNoSpacesOnNewLine() throws BadLocationException {
...
+		format(content, expected, formattingOptions);
+	}
+
+	@Test
+	public void testDontInsertFinalNewLine3() throws BadLocationException {
+		XMLFormattingOptions formattingOptions = createDefaultFormattingOptions();
+		formattingOptions.setTrimFinalNewlines(false);
+		formattingOptions.setInsertFinalNewline(true);
+		String content = "<a  ></a>" + lineSeparator() + "   ";
+		String expected = "<a></a>" + lineSeparator() + "   ";;
```
angelozerr, line 2082: Remove last ';'

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFormatterTest.java
```diff
@@ -2277,6 +2420,8 @@ private static void format(String unformatted, String expected, XMLFormattingOpt
 	}
 
 	private static XMLFormattingOptions createDefaultFormattingOptions() {
-		return new XMLFormattingOptions(2, true);
+		XMLFormattingOptions options = new XMLFormattingOptions(2, true);
```
angelozerr, line 2423: return new XMLFormattingOptions(2, true);

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFormatterTest.java
```diff
@@ -2055,6 +2054,150 @@ public void testNoSpacesOnNewLine() throws BadLocationException {
...
+	public void testInsertFinalNewLine1() throws BadLocationException {
+		XMLFormattingOptions formattingOptions = createDefaultFormattingOptions();
+		formattingOptions.setInsertFinalNewline(true);
+		String content = "<a></a>";
+		String expected = "<a></a>" + lineSeparator();
+		format(content, expected, formattingOptions);
+	}
+
+	@Test
+	public void testInsertFinalNewLine2() throws BadLocationException {
```
angelozerr, line 2096: I will add the same test but by forcing line separator in the XML.

```
String content = "<a></a>\r\n\r\n";
String expected = "<a></a>\r\n";
```

to check that insert final new line use the lien separator declared in the XML (and not the line separator from the OS)

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLBuilder.java
```diff
@@ -394,7 +433,7 @@ private boolean isPreserveEmptyContent() {
 	}
 
 	private int getPreservedNewlines() {
-		if(formattingOptions != null) {
+		if (formattingOptions != null) {
 			return formattingOptions.getPreservedNewlines();
 		}
 		return 2; // default
```
angelozerr, line 439: `return XMLFormattingOptions.DEFAULT_PRESERVE_NEW_LINES;`

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/settings/SettingsTest.java
```diff
@@ -168,7 +168,7 @@ public void formatSettings() {
 		formattingOptions.setTabSize(5);
 		formattingOptions.setInsertSpaces(false);
 
-		XMLFormattingOptions xmlFormattingOptions = new XMLFormattingOptions(formattingOptions);
+		XMLFormattingOptions xmlFormattingOptions = new XMLFormattingOptions(formattingOptions, false);
```
xorye, line 171: I did this because since `initializeDefaultSettings` calls `super.setTabSize(DEFAULT_TAB_SIZE);`, when merging `formattingOptions`, the `tabSize` will not get updated because it has already been set: https://github.com/eclipse/lemminx/blob/55e92f3c0f51717787f2c3b161181dc2654c293d/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java#L346-L356

angelozerr:
Great job @xorye ! Could you add a section which explains those 2 settings (bound with vscode-xml) in https://github.com/redhat-developer/vscode-xml/wiki/Formatting please

PR #652, opened 2020-04-23
Don't generate end element on apply completion if it exists

### LINKED ISSUE #651: Don't generate end element on apply completion if it exists
To fix https://github.com/redhat-developer/vscode-xml/pull/213#issuecomment-579328288 the apply completion (for XML and for XML based on a grammar) must generate the end element only if it needs.

angelozerr:
Don't generate end element on apply completion if it exists

Fixes #651

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
@@ -18,6 +18,7 @@ import static org.eclipse.lemminx.dom.DOMAttr.XMLNS_NO_DEFAULT_ATTR;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 
 import org.eclipse.lemminx.utils.StringUtils;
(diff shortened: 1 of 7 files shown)


angelozerr:
To test this PR you need to test XML completion based on XML Schema/DTD and without it and open completion before an orphelan end element.

Create a XML Schema choice.xsd like this:

```xml
<xs:schema
        attributeFormDefault="unqualified">
    <xs:element name="person">
        <xs:complexType>
            <xs:choice
                    minOccurs="1"
                    maxOccurs="2">
                <xs:element name="employee"/>
                <xs:element name="member"/>
            </xs:choice>
        </xs:complexType>
    </xs:element>
</xs:schema>
```

create a XML file like this:

```xml
<person xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="choice.xsd">
	<employee></employee>
	|</member>
</person>
```

open completion before member and apply it, it should generate only `<member>` without close it.

After write a XML file (without XSD) like this:

```xml
<person>
	<employee></employee>
	|</employee>
</person>
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
```diff
@@ -378,7 +379,49 @@ public boolean isEndTagClosed() {
 	}
 
 	/**
-	 * If Element has a closing end tag eg: <a> </a> -> true , <a> </b> -> false
+	 * Returns true if the given element is an orphan end tag (which have not a
```
fbricon, line 382: which has no start tag

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
```diff
@@ -378,7 +379,49 @@ public boolean isEndTagClosed() {
 	}
 
 	/**
-	 * If Element has a closing end tag eg: <a> </a> -> true , <a> </b> -> false
+	 * Returns true if the given element is an orphan end tag (which have not a
+	 * start tag, eg: </a>) and false otherwise.
+	 * 
+	 * @param tagName the end tag name.
+	 * @return true if the given element is an orphan end tag (which have not a
```
fbricon, line 386: has no start tag

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
angelozerr: fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
angelozerr: fixed

PR #654, opened 2020-04-24
Fix build with completion and orphan

angelozerr:
Check that element has end tag when completion is triggered inside start tag element for searching orphan.


fbricon:
Test case?

angelozerr:
> Test case?

Tests are already written and cause a build error https://ci.eclipse.org/lemminx/blue/organizations/jenkins/lemminx/detail/master/52/pipeline#step-30-log-290. 

This PR fixes those 2 tests.

PR #656, opened 2020-04-24
Cache uris that received "limit exceeded" warnings

xorye:
Fixes https://github.com/redhat-developer/vscode-xml/issues/243

To implement this fix, there is a hashset that keep tracks of uris for files that already received the notification. In this PR, the cache for the file uri is evicted only after the file has been closed.

To test this PR, set `xml.symbols.maxItemsComputed` to a very low value.

Demo:
demo

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
@@ -14,7 +14,6 @@ package org.eclipse.lemminx;
 
 import static org.eclipse.lsp4j.jsonrpc.CompletableFutures.computeAsync;
 
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -29,13 +28,13 @@ import java.util.function.BiFunction;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
-import org.eclipse.lemminx.client.ClientCommands;
 import org.eclipse.lemminx.client.ExtendedClientCapabilities;
+import org.eclipse.lemminx.client.LimitExceededWarnings;
+import org.eclipse.lemminx.client.LimitFeature;
 import org.eclipse.lemminx.commons.ModelTextDocument;
 import org.eclipse.lemminx.commons.ModelTextDocuments;
 import org.eclipse.lemminx.commons.TextDocument;
 import org.eclipse.lemminx.commons.TextDocuments;
-import org.eclipse.lemminx.customservice.ActionableNotification;
 import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMParser;
 import org.eclipse.lemminx.extensions.contentmodel.settings.XMLValidationSettings;
@@ -59,7 +58,6 @@ import org.eclipse.lsp4j.CompletionItem;
 import org.eclipse.lsp4j.CompletionList;
 import org.eclipse.lsp4j.CompletionParams;
 import org.eclipse.lsp4j.DefinitionParams;
-import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.DidChangeTextDocumentParams;
 import org.eclipse.lsp4j.DidCloseTextDocumentParams;
 import org.eclipse.lsp4j.DidOpenTextDocumentParams;
@@ -78,8 +76,6 @@ import org.eclipse.lsp4j.Hover;
 import org.eclipse.lsp4j.HoverParams;
 import org.eclipse.lsp4j.Location;
 import org.eclipse.lsp4j.LocationLink;
-import org.eclipse.lsp4j.MessageParams;
-import org.eclipse.lsp4j.MessageType;
 import org.eclipse.lsp4j.PublishDiagnosticsParams;
 import org.eclipse.lsp4j.ReferenceParams;
 import org.eclipse.lsp4j.RenameParams;
@@ -102,6 +98,7 @@ public class XMLTextDocumentService implements TextDocumentService {
 	private final XMLLanguageServer xmlLanguageServer;
 	private final TextDocuments<ModelTextDocument<DOMDocument>> documents;
 	private SharedSettings sharedSettings;
+	private LimitExceededWarnings limitExceededWarnings;
 
 	/**
 	 * Save context.
@@ -154,6 +151,7 @@ public class XMLTextDocumentService implements TextDocumentService {
 			return parser.parse(document, getXMLLanguageService().getResolverExtensionManager(), true, cancelChecker);
 		});
 		this.sharedSettings = new SharedSettings();
+		this.limitExceededWarnings = null;
 	}
 
 	public void updateClientCapabilities(ClientCapabilities capabilities,
@@ -251,7 +249,9 @@ public class XMLTextDocumentService implements TextDocumentService {
 					.collect(Collectors.toList());
 			
 			if (resultLimitExceeded) {
-				sendSymbolLimitNotification(xmlDocument, symbolSettings);
+				// send warning
+				getLimitExceededWarnings().onResultLimitExceeded(xmlDocument.getTextDocument().getUri(),
+						LimitFeature.SYMBOLS);
 			}
 			return symbols;
 		});
@@ -305,7 +305,8 @@ public class XMLTextDocumentService implements TextDocumentService {
 		TextDocumentIdentifier document = params.getTextDocument();
 		String uri = document.getUri();
 		xmlLanguageServer.getLanguageClient()
-				.publishDiagnostics(new PublishDiagnosticsParams(uri, new ArrayList<Diagnostic>()));
+				.publishDiagnostics(new PublishDiagnosticsParams(uri, Collections.emptyList()));
+		getLimitExceededWarnings().evict(uri);
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/LimitExceededWarnings.java
```diff
@@ -0,0 +1,119 @@
...
+public class LimitExceededWarnings {
+
+	/**
+	 * Default instance.
+	 */
+	private static final LimitExceededWarnings DEFAULT_INSTANCE = new LimitExceededWarnings();
+
+	private final Set<String /* file URI */> cache;
+
+	public LimitExceededWarnings() {
```
angelozerr, line 39: I will not use singleton. I think it should be better that this class should be instanciated in teh XMLTextDocumentService, because you have 2 parameters which change never XMLLanguageClientAPI , and SharedSettings.

 
In other words I will have this constructor to store the 2 final variables client and sharedSettings.

```java
public LimitExceededWarnings(XMLLanguageClientAPI client, SharedSettings sharedSettings) {
   this.client = client;
   this.sharedSettings = sharedSettings;
}
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/LimitExceededWarnings.java
```diff
@@ -0,0 +1,119 @@
...
+		cache = new HashSet<String>();
+	}
+
+	/**
+	 * Gets the default instance.
+	 *
+	 * @return The default instance
+	 */
+	public static LimitExceededWarnings getDefaultInstance() {
```
angelozerr, line 48: Remove this static method

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/LimitExceededWarnings.java
```diff
@@ -0,0 +1,119 @@
...
+	 * 
+	 * After sending the warning, the uri is added to the cache
+	 * 
+	 * @param client         the language client
+	 * @param uri            the file uri
+	 * @param resultLimit    the result limit
+	 * @param featureName    the feature name
+	 * @param sharedSettings the shared settings
+	 */
+	public void onResultLimitExceeded(XMLLanguageClientAPI client, String uri, int resultLimit, String featureName,
```
angelozerr, line 64: replace with 

`public void onResultLimitExceeded(String uri, int resultLimit, String featureName) `

create in this class the enum


```java
public enum LimitFeature {
    
	SYMBOLS("document symbols");
	
	private final String name;
	
	private LimitFeature(String name) {
		this.name = name;
	}
	
	public String getname() {
		return name;
	}
}
```
and add this method

```java
public void onResultLimitExceeded(String uri, LimitFeature feature) {
    int resultLimit = 0;
    switch(feature) {
        case SYMBOLS: 
            resultLimit = sharedSettings.getSymbolSettings().getMaxItemsComputed();
    }
    onResultLimitExceeded(uri, resultLimit, feature.getName());
}
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
```diff
@@ -306,6 +305,7 @@ public void didClose(DidCloseTextDocumentParams params) {
 		String uri = document.getUri();
 		xmlLanguageServer.getLanguageClient()
 				.publishDiagnostics(new PublishDiagnosticsParams(uri, new ArrayList<Diagnostic>()));
```
angelozerr, line 307: replace `new ArrayList<Diagnostic>()` with `Collections.emptyList()`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
```diff
@@ -78,8 +76,6 @@
 import org.eclipse.lsp4j.HoverParams;
 import org.eclipse.lsp4j.Location;
 import org.eclipse.lsp4j.LocationLink;
-import org.eclipse.lsp4j.MessageParams;
-import org.eclipse.lsp4j.MessageType;
 import org.eclipse.lsp4j.PublishDiagnosticsParams;
 import org.eclipse.lsp4j.ReferenceParams;
 import org.eclipse.lsp4j.RenameParams;
```
angelozerr, line 81: Declare field `private LimitExceededWarnings limitExceededWarnings`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
```diff
@@ -251,7 +247,10 @@ private XMLFormattingOptions getFormattingSettings(String uri) {
 					.collect(Collectors.toList());
 			
 			if (resultLimitExceeded) {
-				sendSymbolLimitNotification(xmlDocument, symbolSettings);
+				// send warning
+				LimitExceededWarnings.getDefaultInstance().onResultLimitExceeded(xmlLanguageServer.getLanguageClient(),
```
angelozerr, line 251: replace with getLimitExceededWarnings().onResultLimitExceeded(xmlDocument.getTextDocument().getUri(), Limitfeature.SYMBOLS)

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/LimitExceededWarnings.java
xorye: I've run into a problem, could I pass in `XMLLanguageServer` instead of `XMLLanguageClientAPI`?  Because when `XMLTextDocumentService` is first instantiated, the client is `null`, therefore it calls `new LimitExceededWarnings(null, sharedSettings)`. It is `null` because the client is set after instantiation here:
https://github.com/eclipse/lemminx/blob/c2f5dfd113d71a8c30055c2489843befa7a3fc7f/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLServerLauncher.java#L72

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
```diff
@@ -306,6 +305,7 @@ public void didClose(DidCloseTextDocumentParams params) {
 		String uri = document.getUri();
 		xmlLanguageServer.getLanguageClient()
 				.publishDiagnostics(new PublishDiagnosticsParams(uri, new ArrayList<Diagnostic>()));
+		LimitExceededWarnings.getDefaultInstance().evict(uri);
```
angelozerr, line 308: `getLimitExceededWarnings().evict(uri);`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/LimitExceededWarnings.java
xorye: nevermind that comment, I think its better to instantiate LimitExceededWarnings (if needed) in
`XMLTextDocumentService#getLimitExceededWarnings`

angelozerr: Sorry I had not finished :)

You must define a getLimitExceededWarnings() in the XMLTextDocumentService which checks the null of LimitExceededWarnings and in this case you create the instance of LimitExceededWarnings. At this moment the language client should not be null

xorye: ok understood, thank you!

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/LimitExceededWarnings.java
```diff
@@ -0,0 +1,124 @@
...
+	 * Sends the limit exceeded warning to the client only if the provided
+	 * uri does not exist in the cache
+	 * 
+	 * After sending the warning, the uri is added to the cache
+	 * 
+	 * @param uri            the file uri
+	 * @param resultLimit    the result limit
+	 * @param featureName    the feature name
+	 */
+	public void onResultLimitExceeded(String uri, int resultLimit, String featureName) {
```
angelozerr, line 73: Set this method in private. We will see if we need it in the future.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/LimitFeature.java
```diff
@@ -0,0 +1,27 @@
...
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.client;
+
+public enum LimitFeature {
```
angelozerr, line 14: Add Javadoc

angelozerr:
@xorye I have not seen that `onResultLimitExceeded(String uri, int resultLimit, String featureName) {` was used for test. So you can  set it in public if you wish.

xorye:
@angelozerr oops, thank you for reminding me, fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
```diff
@@ -29,13 +28,13 @@
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
-import org.eclipse.lemminx.client.ClientCommands;
 import org.eclipse.lemminx.client.ExtendedClientCapabilities;
```
angelozerr, line 31: Remove unsued `import org.eclipse.lsp4j.Diagnostic;`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/LimitExceededWarnings.java
```diff
@@ -0,0 +1,124 @@
...
+
+	/**
+	 * Send limit exceeded warning to client
+	 * 
+	 * @param uri            the file uri
+	 * @param resultLimit    the result limit
+	 * @param featureName    the feature name
+	 */
+	private void sendLimitExceededWarning(String uri, int resultLimit, String featureName) {
+		String filename = Paths.get(uri).getFileName().toString();
```
angelozerr, line 105: On Windows OS, this code throws an exception.

Replace this line with:

```java
String filename = Paths.get(URI.create(uri)).getFileName().toString();
```

angelozerr:
Great job @xorye !

PR #240, opened 2018-11-29
Hover show documentation for XSI attributes

NikolasKomonen:
Fixes #164

PR #670, opened 2020-05-01
Remove spacing when formatting processing instruction

xorye:
Fixes #240 

### DIFF
--- a/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFormatterTest.java
+++ b/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFormatterTest.java
@@ -283,15 +283,15 @@ public class XMLFormatterTest {
 
 	@Test
 	public void testProlog() throws BadLocationException {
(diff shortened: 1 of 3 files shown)


xorye:
Before formatting:
image

After formatting:
image

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFormatterTest.java
```diff
@@ -284,22 +284,22 @@ public void rangeSelectWithinText() throws BadLocationException {
 	@Test
 	public void testProlog() throws BadLocationException {
 		String content = "<?xml version=   \"1.0\"       encoding=\"UTF-8\"  ?>" + lineSeparator();
-		String expected = "<?xml version=\"1.0\" encoding=\"UTF-8\" ?>" + lineSeparator();
+		String expected = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>" + lineSeparator();
```
angelozerr, line 287: Remplace lineSeparator with "\r\n". The idea is to use lineSepatrator when it is required (ex : when XML doesnt define lien separator in the XML content and after formatting a line separator is generated. In this case lien separator is used.

PR #672, opened 2020-05-04
Set format.trimFinalNewlines default to true

xorye:


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java
```diff
@@ -343,6 +344,21 @@ public EmptyElements getEmptyElements() {
...
 	}
 
+	/**
+	 * Returns the value of trimFinalNewlines.
+	 * 
+	 * If the trimFinalNewlines does not exist, defaults to true.
+	 */
+	@Override
+	public boolean isTrimFinalNewlines() {
+		final Boolean value = this.getBoolean(TRIM_FINAL_NEWLINES);
```
fbricon, line 354: return (value == null)? true : value;

xorye: fixed

angelozerr: Please define a default value (like other value) instead of hard coding that.

`return (value == null)? DEFUALT_TRIM_FINAL_NEWLINES : value;`

PR #687, opened 2020-05-11
Read the cached XSD, DTD grammar file with lazy mode

angelozerr:
Read the cached XSD, DTD grammar file with lazy mode

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/uriresolver/XMLCacheResolverExtension.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/uriresolver/XMLCacheResolverExtension.java
@@ -13,6 +13,7 @@
 package org.eclipse.lemminx.extensions.contentmodel.uriresolver;
 
 import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/uriresolver/XMLCacheResolverExtension.java
```diff
@@ -32,6 +33,30 @@
...
+		public XMLFileInputSource(XMLResourceIdentifier resourceIdentifier, Path file) {
+			super(resourceIdentifier);
+			this.file = file;
+		}
+
+		@Override
+		public InputStream getByteStream() {
+			// Load the file input stream only if it is used
+			InputStream input = super.getByteStream();
+			if (input == null) {
```
fbricon, line 49: check if path exists and is a file

angelozerr: This test is done before (file is null when it doesn't exists), in otherwise don't need to do that here.

fbricon:
@xorye please test this one

PR #695, opened 2020-05-13
No hyperlink to DTD source on hover

### LINKED ISSUE #693: No hyperlink to DTD source on hover
Given this web.xml:
```xml
<!DOCTYPE web-app PUBLIC
 "-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN"
 "http://java.sun.com/dtd/web-app_2_3.dtd" >

<web-app>
  <display-name>Servlet 2.3 Web Application</display-name>
</web-app>
```

Source hyperlink to web-app_2_3.dtd is missing. It only seems to work for xsd schemas.

Screen Shot 2020-05-12 at 8 12 58 PM

angelozerr:
No hyperlink to DTD source on hover

Fixes #693


angelozerr:
The problem came from the file path used to generate documentation. Indeed we generate doc inHTML and convert in Markdow, when path is without `file:///`, the markdown converter doesn't generate hyperlink (DTD case).

The fix is to store DTD with URI instead of file path.

fbricon:
Thanks for the quick fix!

angelozerr:
You are welcome :)

PR #705, opened 2020-05-15
Find definition for locally declared entity

### LINKED ISSUE #625: Find definition for locally declared entity
Find definition for declared local entity .

Given this XML file:

```xml
<!DOCTYPE article [
  <!ENTITY mdash "&#x2014;">
]>
<root>
  &mdash;&mdash;
</root>
```

Ctrl+click on `&mdash;`should find definition of `mdash `

LocalEntityDefinitionDemo

angelozerr:
Find definition for locally declared entity

Fixes #625

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DTDElementDecl.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DTDElementDecl.java
@@ -12,7 +12,11 @@
  */
 package org.eclipse.lemminx.dom;
 
+import static org.eclipse.lemminx.utils.StringUtils.findEndWord;
+import static org.eclipse.lemminx.utils.StringUtils.findStartWord;
+
 import java.util.function.BiConsumer;
+import java.util.function.Predicate;
 
 /**
  * DTD Element Declaration <!ELEMENT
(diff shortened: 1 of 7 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DTDElementDecl.java
```diff
@@ -22,6 +26,8 @@
  */
 public class DTDElementDecl extends DTDDeclNode {
 
+	private static Predicate<Character> isValidChar = (c) -> Character.isJavaIdentifierPart(c) || c == '-';
```
fbricon, line 29: final

angelozerr: good catch, fixed

angelozerr:
Here a demo with this XML:

```xml
<!DOCTYPE article [
  <!ENTITY mdash "&#x2014;">
]>
<root>
  &mdash;&mdash;
</root>
```

LocalEntityDefinitionDemo

xorye:
It seems that
https://github.com/eclipse/lemminx/blob/6d2256fd3b9a880022ab4c9ec5644d80fc186e39/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DTDElementDecl.java#L91

should be changed to:
`* <!ELEMENT n|ote (to,from,heading,body)> will return null. `

PR #711, opened 2020-05-19
Avoid duplicate of snippet label

### LINKED ISSUE #710: Avoid duplicate of snippet label
Today snippet uses prefix as label, but it causes duplicate of completion label:

image

The prefix is used as label like vscode does with their snippet. The idea is to keep this default behavior and add a new JSON field like `useDescriptionAsLabel: true`

@fbricon what do you think about this idea?

angelozerr:
Avoid duplicate of snippet label

Fixes #710

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/snippets/Snippet.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/snippets/Snippet.java
@@ -23,6 +23,8 @@ import java.util.function.BiPredicate;
  */
 public class Snippet {
 
+	private String label;
+
 	private List<String> prefixes;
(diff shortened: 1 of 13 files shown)


angelozerr:
@fbricon here the result:

image

datho7561:
I did some manual testing to see if $prefix and $description can be both used in a snippet (it works). If description is set to an empty string and $description is used in the label, then it appears as $description. This makes sense, but it might make sense to document this in a unit test to show it's intended. Other than that, this looks good to me.

angelozerr:
> This makes sense, but it might make sense to document this in a unit test to show it's intended. 

You are right, I have written tests for that.

> Other than that, this looks good to me.

Thanks for your review.

Issue #699, opened 2020-05-13
[xml-model] Snippet to generate xml-model


angelozerr:
Snippet to generate xml-model and their attributes (with choice). See https://www.w3.org/XML/2010/01/xml-model/

To do that:

 * rename `prolog-snippets.json` to `processing-instruction-snippets.json`
 * write xml-model snippets in `processing-instruction-snippets.json`.

BalduinLandolt:
I'll give that a go, if that's allright and no one is working on it already?

angelozerr:
@datho7561 should work on this issue, but i think he has not started it. @datho7561 if you have not started something, are you OK to delegate this issue to @BalduinLandolt ?

datho7561:
I haven't started working on this issue, and I'm okay with @BalduinLandot working on it.

angelozerr:
> I haven't started working on this issue, and I'm okay with @BalduinLandot working on it.

Thanks @datho7561 for your answer.

@BalduinLandolt any contribution are welcome! I think we should have a snippets with required attribute like href and another (one or several) snippets with other attributes. Some attributes like schematypens should be declared as choice to have a list of values.

BalduinLandolt:
Allright then. :)

In fact, I'd rather leave the prolog snippet as it is, and make processing insructions a separate on. That is for two reasons:
1) the prolog is technically not a PI (see #727)
2) the PIs have different rules, where they can be triggered: the prolog can basically only be in the first position. the PIs can be everywhere before the root element, except before the prolog.

... unless, of course, there are objections to that?  
I'll just give it a go, and then draft a PR. If you don't like it, you'll see that right away, there.

Issue #702, opened 2020-05-14
CodeAction to create missing noNamespaceSchemaLocation and generate XSD that adheres to current XML


xorye:
This issue is an extension to https://github.com/eclipse/lemminx/issues/691

https://github.com/eclipse/lemminx/issues/691 implements a new code action to create a new missing XSD file with generic content (prolog). This issue is to improve upon this, by generating XSD that adheres to the newly created XSD file's corresponding XML.

angelozerr:
For your information, @xorye I have started to experiment to generate XSD from XML files and DTD from XML files. See my branch at  https://github.com/angelozerr/lemminx/tree/grammar-generator2

angelozerr:
Fixed with https://github.com/eclipse/lemminx/issues/778

PR #730, opened 2020-05-27
xml-model snippets

BalduinLandolt:
This PR adresses Issue #699.

To Do:

- [x] Ensure snippet can only be triggered in valid places.
- [x] Add Unit Test

I could also add a snippet that includes href, type and schematypens... But it's already four snippets, adding another one might be a bit too extensive?


BalduinLandolt:
Also: No idea, why the unit tests keep failing me... I didn't change any of these files.

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
```diff
@@ -0,0 +1,52 @@
...
+	  "label": "$description",
+	  "description": "Insert XML Model Processing Instruction for Schema Association including Type"
+	},
+	"Insert XML Model Schema Association with Schematypens": {
+	  "prefix": [
+		"<?xml-model"
+	  ],
+	  "suffix": "?>",
+	  "body": [
+		"<?xml-model href=\"${1:location}\" schematypens=\"${2:schematypens}\"?>${0}"
```
angelozerr, line 30: Schematypeens should be a choice

angelozerr:
> Also: No idea, why the unit tests keep failing me... I didn't change any of these files.

Test failed because we have tests which check the number of returned snippets, when we add new snippets we need to update those tests. I know it's annoying but we have robust test and we are sure number of snippets that we return.

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
```diff
@@ -0,0 +1,63 @@
...
+	  "label": "$description",
+	  "description": "Insert XML Model Processing Instruction for Schema Association including Schematypens"
+	},
+	"Insert XML Model Schema Association with Type and Schematypens": {
+	  "prefix": [
+		"<?xml-model"
+	  ],
+	  "suffix": "?>",
+	  "body": [
+		"<?xml-model href=\"${1:location}\" ${2|type=\"application/xml-dtd\",type=\"application/xml\" schematypens=\"http://www.w3.org/2001/XMLSchema\",type=\"application/xml\" schematypens=\"http://relaxng.org/ns/structure/1.0\",type=\"application/relax-ng-compact-syntax\",type=\"application/xml\" schematypens=\"http://purl.oclc.org/dsdl/schematron\",type=\"application/xml\" schematypens=\"http://purl.oclc.org/dsdl/nvdl/ns/structure/1.0\"|}?>${0}"
```
angelozerr, line 41: Why you have several type= attributes?

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
BalduinLandolt: Done.

I also added a snippet for both type and schematypens, that works entirely with options and gets the correlation right (in which case use which).  
That's actually a lot nicer and makes the snippet doe type and for schematypens obsolete.

There is just one problem:
grafik
The text is too long, so you can't see what they actually are... any idea how to get around this?

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
BalduinLandolt: see my reply above :)

BalduinLandolt:
> Test failed because we have tests which check the number of returned snippets, when we add new snippets we need to update those tests. I know it's annoying but we have robust test and we are sure number of snippets that we return.

Well... that makes a lot of sense.  
And in fact, it's very usefull, it showed me a couple of cases where cml-model snippet should not be showing up, but it did.

I can fix that, as soon as I know how many xml-model snippets there really are.  
Do you think it should be a combined snippet for type and schematypens (as they tend to be related) or one for each, or all three?

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
BalduinLandolt: I guess what I could do is to give an attribte `title="DTD Schema"`, `title="XSD Schema"`, etc. And if I put that attribute first, that would make it easily readable, what the options mean.

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
BalduinLandolt: > 
> 
> Why you have several type= and schematypens= attributes?

My reasoning was the following: if sou use an xsd schema, you'll allways want `type="application/xml"` and the according schematypens, if you have a dtd, you'll allways want `type="application/xml-dtd"` but you don't want the schematypens at all.  
So instead of having two different choices, where people can choose combinations that don't fit; and to avoid having the schematypens in cases where it can't be (DTD and compact RelaxNG), I just put both attributes in one choice.

I thoght that's actually more reliable and user friendly. But if you don't think it's a good idea, I can easily get rid of it again.

angelozerr:
> Do you think it should be a combined snippet for type and schematypens (as they tend to be related) or one for each, or all three?

The hard part with snippet is to provide the commons use and not to try to provide a lot and complex snippets. 

IMHO I think we should have a snippet per grammar type and that's all. As we don't support schematron and relax NG I will skip them. I mean we should have just 2 xml-model snippets which generate href, type and schematypens (without choice, sorry it was a bad idea). I mean we should provide the 2 snippets: 

 * `Insert XML Schemas association`: this snippet should generate `<?xml-model href=\"${1:file.xsd}\" schematypens="http://www.w3.org/2001/XMLSchema" ?>${0}`
 * `Insert DTD association` : this snippet should generate `<?xml-model href=\"${1:file.dtd}\" type="application/xml-dtd" ?>${0}`

The idea is to provide the minimal snippet and generate an href with the proper extension and provide the same feature than xs:noNamespaceShemaLocation (validation + code action to generate the XSD or DTD from the XML, see https://github.com/eclipse/lemminx/issues/702)

@BalduinLandolt what do you think about this idea?

BalduinLandolt:
> @BalduinLandolt what do you think about this idea?

Yes, that's definitely better, that keeps it a lot simpler. I'll implement that right away.

Another thing: I messed up the first couple of commit messages (the signed-off-by bit), and therefor the eclipsefdn/eca check fails. And I haven't figured out a good way to change this, now that it's pushed already.  
Can it still be merged? Or is that a no go?  
in the latter case, I'd just create a new branch and pull request that gets it right from the start.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/snippets/ProcessingInstructionSnippetContext.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.Map;
+
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.dom.DOMElement;
+import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.services.extensions.ICompletionRequest;
+import org.eclipse.lemminx.utils.DOMUtils;
+
+public class ProcessingInstructionSnippetContext implements IXMLSnippetContext {
```
angelozerr, line 22: Add Javadoc

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/snippets/ProcessingInstructionSnippetContext.java
```diff
@@ -0,0 +1,60 @@
+/*******************************************************************************
```
angelozerr, line 1: @fbricon is this header is correct?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/snippets/ProcessingInstructionSnippetContext.java
```diff
@@ -0,0 +1,60 @@
...
+* Copyright (c) 2020 Balduin Landolt and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
```
angelozerr, line 6: Please add 

```
 *
 * SPDX-License-Identifier: EPL-2.0
 *
```

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
```diff
@@ -0,0 +1,24 @@
+{
+	"Insert XML Model: XSD": {
+	  "prefix": [
+		"<?xml-model"
+	  ],
+	  "suffix": "?>",
+	  "body": [
+		"<?xml-model href=\"${1:file.xsd}\" type=\"application/xml\" schematypens=\"http://www.w3.org/2001/XMLSchema\"?>${0}"
+	  ],
+	  "label": "$description",
```
angelozerr, line 10: I think we should have a shorter label. Here a suggestion

`"label": "Insert XML Schema association",`

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
```diff
@@ -0,0 +1,24 @@
...
+	},
+	"Insert XML Model: DTD": {
+		"prefix": [
+		  "<?xml-model"
+		],
+		"suffix": "?>",
+		"body": [
+		  "<?xml-model href=\"${1:file.dtd}\" type=\"application/xml-dtd\"?>${0}"
+		],
+		"label": "$description",
```
angelozerr, line 21: I think we should have a shorter label. Here a suggestion

`"label": "Insert DTD association",`

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/DTDCompletionExtensionsTest.java
```diff
@@ -196,7 +196,7 @@ public void externalDTDCompletionAllDeclsSnippetsNotSupported() throws BadLocati
 	public void testNoDuplicateCompletionItems() throws BadLocationException {
 		// completion on <|
 		String xml = "<?xml version=\"1.0\" standalone=\"no\" ?>\n"
-				+ "<!DOCTYPE svg PUBLIC \"-//W3C//DTD SVG 1.0//EN\" \"http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg.dtd\">\n"
+				+ "<!DOCTYPE svg PUBLIC \"-//W3C//DTD SVG 1.0//EN\" \"http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg.dtd\">\n" // FIXME: I can't find this file. does it still exist?
```
angelozerr, line 199: Don't change this line, this test uses XML catalog and test is working (see CI build)

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/DTDCompletionExtensionsTest.java
```diff
@@ -216,7 +216,7 @@ public void elementCompletionWithDoctypeSubsetWithNoElements() throws BadLocatio
 				"<web-app>\r\n" + //
 				"  <display-name>Servlet 2.3 aWeb Application</display-name>\r\n" + //
 				"</web-app>";
-		testCompletionFor(xml, false, 76 + 2 /* CDATA and Comments */,
+		testCompletionFor(xml, false, 76 + 4 /* XML-Model, CDATA and Comments */,
```
angelozerr, line 219: Please use your constants, I mean PROCESSING_INSTRUCTION_SNIPPETS /* XML Model */  and use existing constants for CDATA and Comments

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLCompletionSnippetsTest.java
```diff
@@ -92,12 +108,19 @@ public void emptyXMLContent() throws BadLocationException {
...
+						r(0, 0, 0, 2), "<?xml-model"),
+				c("Insert XML Model Processing Instruction to Associate Document Type Description (DTD)", //
+						"<?xml-model href=\"file.dtd\" type=\"application/xml-dtd\"?>", //
+						r(0, 0, 0, 2), "<?xml-model"),
 				c("<!--", //
 						"<!-- -->", //
 						r(0, 0, 0, 2), "<!--"));
 
-		testCompletionFor("<!|", NEW_XML_SNIPPETS /* DOCTYPE snippets */ + //
+		testCompletionFor("<!|", NEW_XML_SNIPPETS /* DOCTYPE snippets */ + // TODO: why does that trigger these completions? is that intended?
```
angelozerr, line 121: Yes we return all snippets and filter is done by client. It gives the capability to benefit with fuzzy completion if client (like vscode, Eclipse IDE) supports it.

angelozerr:
> Another thing: I messed up the first couple of commit messages (the signed-off-by bit), and therefor the eclipsefdn/eca check fails. And I haven't figured out a good way to change this, now that it's pushed already.

Have you signed your ECA https://www.eclipse.org/legal/ecafaq.php ?

> Can it still be merged? Or is that a no go?

Please fix my comments and please merge your commits in one commit.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/snippets/ProcessingInstructionSnippetContext.java
fbricon: Yes it's correct

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/snippets/ProcessingInstructionSnippetContext.java
BalduinLandolt: For the headerI used the vscode snippet (as defined here: https://github.com/eclipse/lemminx/blob/master/.vscode/project.code-snippets). I guess that should be added there as well, then?

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
BalduinLandolt: aye

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/DTDCompletionExtensionsTest.java
BalduinLandolt: yes, that was just a temporary "note to self" :)

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/DTDCompletionExtensionsTest.java
BalduinLandolt: makes sense. done.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLCompletionSnippetsTest.java
BalduinLandolt: ok :)

BalduinLandolt:
> 
> 
> > Another thing: I messed up the first couple of commit messages (the signed-off-by bit), and therefor the eclipsefdn/eca check fails. And I haven't figured out a good way to change this, now that it's pushed already.
> 
> Have you signed your ECA https://www.eclipse.org/legal/ecafaq.php ?

Yes. The problem is that the Agreement only accepts one email address, and I accidentally pushed signed with various different addresses. I hope squashing will fix it, though.

> 
> > Can it still be merged? Or is that a no go?
> 
> Please fix my comments and please merge your commits in one commit.

Changed all necessary things. Will merge the commits into one tomorrow.

BalduinLandolt:
Well, I didn't find a good way of fixing these problems I accidentally caused.

What I did now was create a new branch with all the commits squashed into one, and a correct signed-off-by, so eclipsefdn doesn't complain anymore.  
This means, however, that I'll have to create a new PR from this branch, which will make this one obsolete... Sorry about the confusion/complication! (I'm learning... ;) )

PR #745, opened 2020-05-29
Fixed error range for SemicolonRequiredInReference

### LINKED ISSUE #664: Fix error range for `SemicolonRequiredInReference`
Given this XML file:

```xml
<!DOCTYPE article [
  <!ENTITY mdash "&#x2014;">
]>
<root>
  &mdash
</root>
```

`&mdash` should be ends with `;` and Xerces report the error as `SemicolonRequiredInReference`. We should adjust error range to highlight `&mdash`

datho7561:
The error range now goes from the start of the ampersand to
the end of the entity name.

SemicolonRequiredInReference

Fixes #664

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/XMLSyntaxErrorCode.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/XMLSyntaxErrorCode.java
@@ -31,6 +31,7 @@ import org.eclipse.lemminx.extensions.contentmodel.participants.codeactions.Root
 import org.eclipse.lemminx.services.extensions.ICodeActionParticipant;
 import org.eclipse.lemminx.services.extensions.diagnostics.IXMLErrorCode;
 import org.eclipse.lemminx.utils.XMLPositionUtility;
+import org.eclipse.lemminx.utils.XMLPositionUtility.EntityReferenceRange;
 import org.eclipse.lsp4j.Range;
 
 /**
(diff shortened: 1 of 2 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/XMLSyntaxErrorCode.java
```diff
@@ -167,6 +167,10 @@ public static Range toLSPRange(XMLLocator location, XMLSyntaxErrorCode code, Obj
 			String tag = getString(arguments[0]);
 			return XMLPositionUtility.selectChildEndTag(tag, offset, document);
 		}
+		case SemicolonRequiredInReference: {
```
angelozerr, line 170: I will trylike this:

```
case SemicolonRequiredInReference:
		case EntityNotDeclared: {
			EntityReferenceRange range = XMLPositionUtility.selectEntityReference(offset - 1, document, false);
			return range != null ? range.getRange() : null;
		}
```

datho7561: This doesn't include the ampersand, and just highlights the entity name. Should I include the ampersand in the highlight range?

angelozerr: It should include it because when entity reference doesn't exist it highlight &. Are you sure that you are using the last master branch?

datho7561: I checked out my master branch, and updated to eclipse/lemminx master. I rebuilt the server and ran the VSCode extension. I get this as the error range: 
ErrorRangeNoSemicolon
The default range must have changed since this issue was opened.

angelozerr: @datho7561 sorry with my mistake. Please keey EntityNotDeclared as today and add SemicolonRequiredInReference case like this:

```java
case SemicolonRequiredInReference: {
			EntityReferenceRange range = XMLPositionUtility.selectEntityReference(offset + 1, document, false);
			return range != null ? range.getRange() : null;
		}
```

datho7561: Thank you!

Issue #742, opened 2020-05-28
No entity completion for externally declared SYSTEM and PUBLIC entities


xorye:
**test.dtd**
```xml
<!ELEMENT author (#PCDATA)>
<!ENTITY writer PUBLIC "public id" "entity uri">
<!ENTITY copyright SYSTEM "uri">
```

**test.xml**
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE author SYSTEM "test.dtd">
<author>&writer;&copyright;</author>
```

There is no completion for `&writer;` and `&copyright;` in the xml file:
![image](https://user-images.githubusercontent.com/20326645/83179652-46e17f80-a0f0-11ea-925e-5825fc3bcfb7.png)

fbricon:
Fixed with #746

PR #746, opened 2020-05-29
Entity documentation has no value for entities declared with SYSTEM OR PUBLIC

### LINKED ISSUE #741: Entity documentation has no value for entities declared with SYSTEM OR PUBLIC
With this XML

```
<?xml version="1.0" standalone="no" ?>
<!DOCTYPE copyright [
  <!ELEMENT copyright (#PCDATA)>
  <!ENTITY c SYSTEM "http://www.xmlwriter.net/copyright.xml">
]>
<copyright>&c;</copyright>
```
The hover does not show the value of c:
image

The same thing happens if we have 
`<!ENTITY c PUBLIC "http://www.xmlwriter.net/copyright.xml">`
instead of
`<!ENTITY c SYSTEM "http://www.xmlwriter.net/copyright.xml">`

It would be great if the entity's value (`"http://www.xmlwriter.net/copyright.xml"` in this case) was displayed in the documentation.

angelozerr:
Entity documentation has no value for entities declared with SYSTEM OR PUBLIC

Fixes #741


angelozerr:
This PR:

 * fixes https://github.com/eclipse/lemminx/issues/741
 * fixes https://github.com/eclipse/lemminx/issues/742
 * add public + system ID in the documentation (hover + completion)
 * add a Source link.

angelozerr:
> To stay consistent, the #742's case should have SYSTEM ID instead of Value

Indeed it was a bug, it should be fixed.

> I understand that this is a draft PR, so sorry if the implementation wasn't done

No please give me feedback and don't hesitate to give me those kind of comments.

It's draft PR because I have not written test and not commented my code.

angelozerr:
Test are now written, please review it.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/entities/EntitiesDocumentationUtils.java
```diff
@@ -64,6 +65,19 @@ public String getValue() {
...
 
+	public static MarkupContent getDocumentation(String entityName, String entityValue, EntityOriginType type,
+			boolean markdown) {
+		return getDocumentation(entityName, entityValue, null, null, null, type, markdown);
+	}
+
+	public static MarkupContent getDocumentation(DTDEntityDecl entity, EntityOriginType type, boolean markdown) {
+		String systemID = entity.getSystemId();
+		String publicID = entity.getPublicId();
+		String targetURI = entity.getNameParameter().getTargetURI();
```
xorye, line 76: Do we need to check if  `entity.getNameParameter() == null` here? Although, I'm not sure if it would happen in our case

angelozerr: If entity has name, entity.getNameParameter() is not null, so we don't need to check that.

xorye:
This PR looks & works great

PR #754, opened 2020-06-01
Disable snippets in front of XML declaration.

### LINKED ISSUE #729: prolog snippet can be triggered right in front of prolog snippet
grafik

this would result in a double prolog, which would not be valid.

BalduinLandolt:
Fixes #729.

Signed-off-by: Balduin Landolt <balduin.landolt@hotmail.com>


BalduinLandolt:
Ready for review.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLCompletionSnippetsTest.java
```diff
@@ -537,4 +537,39 @@ public void emptyCompletionInsideDOCTYPE() throws BadLocationException {
 		testCompletionFor("<!DOCTYPE root-element |SYSTEM \"file.dtd\">", 0);
 		testCompletionFor("<!DOCTYPE root-element SYSTEM |\"file.dtd\">", 0);
 	}
+
+	// Tests triggering completion at the very beginning of the file.
+	// Should be possible if not in front of the xml declaration.
+	@Test
```
datho7561, line 543: Add a test case for completion at the beginning for a document with just the xml declaration:
xsi-appear-before-xml-decl
I think it might be a good idea to also add a test case with a completely empty file.

BalduinLandolt: > 
> 
> Add a test case for completion at the beginning for a document with just the xml declaration:
> xsi-appear-before-xml-decl

woops, yes.

> I think it might be a good idea to also add a test case with a completely empty file.

I think this is covered sufficiently by the empty xml test:
https://github.com/eclipse/lemminx/blob/c97206e6de3033f21445401b29d9a651dde6cd95/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLCompletionSnippetsTest.java#L37-L137

datho7561: Okay. Sorry, I didn't know that checks for empty documents were already there. I should have checked first.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/snippets/XMLDeclarationSnippetContext.java
```diff
@@ -37,8 +37,8 @@ public boolean isMatch(ICompletionRequest request, Map<String, String> model) {
 			// --> <?xml version="1.0" encoding="UTF-8" | ?>
 			return false;
 		}
-		// check if document already defined a xml processing instruction.
-		if (hasProlog(document, offset)) {
+		// check if document already defined a xml declaration.
+		if (hasProlog(document, offset) || (document.hasProlog() && offset == 0)) {
```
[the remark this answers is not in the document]
BalduinLandolt: done.

code looks a bit different than your suggestion, though, because there's no `getProlog()` (which could be added, but I wasn't sure if I should do that) and the refactored `inProlog(node, offset)` takes the document, not the node (which again, could be changed, but I didn't want to touch on that).

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLCompletionSnippetsTest.java
BalduinLandolt: no worries!

changed the rest, though.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/snippets/CommentSnippetContext.java
```diff
@@ -36,6 +37,13 @@ public boolean isMatch(ICompletionRequest request, Map<String, String> model) {
 			// completion was triggered inside doctype declaration, ignore the snippets
 			return false;
 		}
+		
+		DOMDocument document = request.getXMLDocument();
+		int offset = request.getOffset();
+		if (document.hasProlog() && offset == 0){
```
angelozerr, line 43: Not tested, but is it working if you have some spaces before the xml(declaration and offset is not equals to 0? I mean :

`       |       <?xml ?>`

BalduinLandolt: I think that would trigger, but I haven't given that much thought because it's marked as invalid anyways.  
Should I address it nonetheless?

angelozerr: Yes we must manage this case even if validation report an error. For instance if user is typing quickly, he cannot see the error validation. Imagine that validation could be disable (it's not the case), snippet must be shown only if it's valid.

BalduinLandolt: done.  
But in these cases the `#region` and `#endregion` snippets trigger anyways. This is clearly wrong, but I didn't want to touch them because their implementation differs from all other snippets...

angelozerr: We should switch #region and #endregion in snippets. Please create an issue for that and if you wish yo do that and assign you to the issue.

BalduinLandolt: ok, I open an issue for that.

BalduinLandolt: but for this PR, should I leave them in the unit test, so the test passes, even tough it is wrong? Or should I keep the test correct, which will than inevitably fail?

```Java
xml = "| <?xml version=\"1.0\" encoding=\"UTF-8\" ?>";
testCompletionFor(xml, REGION_SNIPPETS); // test passes, but it's wrong
```
or?

```Java
xml = "| <?xml version=\"1.0\" encoding=\"UTF-8\" ?>";
testCompletionFor(xml, 0); // test fails
```

angelozerr: CI build must not fails because it build LemMinx JAR on each commit (and vscode-xml as well). The test will be fixed when region will be fixed too.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/snippets/CommentSnippetContext.java
```diff
@@ -36,6 +37,13 @@ public boolean isMatch(ICompletionRequest request, Map<String, String> model) {
 			// completion was triggered inside doctype declaration, ignore the snippets
 			return false;
 		}
+		
+		DOMDocument document = request.getXMLDocument();
+		int offset = request.getOffset();
+		if (document.hasProlog() && offset <= document.getProlog().getStart()){
```
angelozerr, line 43: I will define in DOMDocument `isInBeforeProlog(int offset)` and I will use it in all *SnippetContext classes

BalduinLandolt: sounds reasonable. Did you mean something like this?

PR #755, opened 2020-06-02
Provide more server/build info on startup

fbricon:
Includes server version, Java home, git branch, short sha1 commit id and message

Screen Shot 2020-06-08 at 4 32 03 PM

Rationale behind this change is because I often find myself wondering whether I properly rebuilt the server when testing a new PR.


org.eclipse.lemminx/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "java.configuration.updateBuildConfiguration": "interactive",
```
xorye, line 2: Why do we need this?

xorye:
Looks & works great

org.eclipse.lemminx/.vscode/settings.json
fbricon: we don't

Issue #217, opened 2018-11-12
Validation needs additional <uri> catalog entry


ralfhandl:
To make validation and completion work I currently need two catalog entries, e.g.
```xml
  <public publicId="urn:oasis:names:tc:entity:xmlns:xml:catalog" uri="./schemas/catalog.xsd" />
  <uri name="urn:oasis:names:tc:entity:xmlns:xml:catalog" uri="./schemas/catalog.xsd" />
```

From the [XML Catalog specification](https://www.oasis-open.org/committees/download.php/14809/xml-catalogs.html#s.ext.ent) I would have expected to just need the `<uri>` entry because `<public>` is intended for resolving *external identifiers* and expressly *not* for resolving XML namespace names:

> External Identifiers, as defined in [Production 75] of [XML], identify the external subset, entities, and notations of an XML document. They are not used to identify other resources such as namespace names, stylesheets, and schema languages other than DTDs; URI entries are used for that purpose.

However with just a `<uri>` entry neither validation nor code completion are triggered.

Note: this issue is of low importance as validation and completion work fine as long as the catalog contains these "duplicate" entries. Mentioning this in documentation on "how to write a catalog.xml" would be sufficient.

angelozerr:
@ralfhandl I have tried to debug the xml-resolver project but to be honnest I'm a little lost -(

I have tried to use your XML catalog with Eclipse WTP, but only completion is working, the validation doesn't work. 

It should be really cool if somebody could help us to know if catalog is valid or not.

ralfhandl:
@angelozerr Strange. I'm also using Eclipse Photon with the editor from 
* Eclipse XML Editors and Tools 
* 3.10.0.v201804210200	
* org.eclipse.wst.xml_ui.feature.feature.group	
* Eclipse Web Tools Platform

The file `...eclipse-workspace-Photon\.metadata\.plugins\org.eclipse.wst.xml.core\user_catalog.xml` only contains `<uri>` entries, e.g.
```xml
  <uri name="http://docs.oasis-open.org/odata/ns/edm" uri="platform:/resource/OData-TC/odata-csdl-schemas/schemas/edm.xsd"/>
  <uri name="http://docs.oasis-open.org/odata/ns/edmx" uri="platform:/resource/OData-TC/odata-csdl-schemas/schemas/edmx.xsd"/>
```

Validation and code completion both work fine with this catalog.

angelozerr:
[...]

PR #757, opened 2020-06-02
Send notification to client when xml.catalogs contains an invalid path

xorye:
For #217 

My plan is to check for invalid paths from `XMLTextDocumentService`, and send a notification to the client if an invalid path exists here:
https://github.com/eclipse/lemminx/blob/977a96757b1f9cd472b81394e4cfd2dd866867b5/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java#L424-L429

The reason why I have this in `XMLTextDocumentService` is that it is easy detect incorrect paths and send notifications here, because there is easy access to `XMLLanguageServer#languageClient`.

The problem with this approach is that path validation happens twice, once in the code above and once here:
https://github.com/eclipse/lemminx/blob/84d268957060b53aee211bd75a34b4536e9dab1e/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/uriresolver/XMLCatalogResolverExtension.java#L102-L131

Would this be ok?


angelozerr:
> Would this be ok?

I would like to avoid doing that because it's a catalog feature, so it should be managed in the catalog extension (see XMLCatalogPlugin which manages is able to return the proper catalog.xsd from urn:oasis:names:tc:entity:xmlns:xml:catalog namespace). I mean if there are some validation to do, it should be done in `XMLCatalogPlugin#doSave(ISaveContext context)`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
```diff
@@ -565,4 +571,12 @@ public LimitExceededWarnings getLimitExceededWarnings() {
 		}
 		return this.limitExceededWarnings;
 	}
+
+	public PathWarnings getPathWarnings() {
```
angelozerr, line 575: This method should be moved in the `XMLCatalogPlugin`.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/PathWarnings.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Set;
+
+import org.eclipse.lemminx.customservice.ActionableNotification;
+import org.eclipse.lemminx.customservice.XMLLanguageClientAPI;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lsp4j.Command;
+import org.eclipse.lsp4j.MessageParams;
+import org.eclipse.lsp4j.MessageType;
+
+public class PathWarnings {
```
angelozerr, line 21: A lof of code are similar with LimitExceededWarnings. I suggest that you create an abstract class like `AbstractXMLNotification which manages cache and send of the notification and LimitExceededWarnings and PathWarnings  extend it.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
```diff
@@ -418,6 +422,8 @@ public void didSave(DidSaveTextDocumentParams params) {
 	 * @param settings
 	 */
 	public void updateSettings(Object settings) {
+		// here, read settings and check if xml.catalog path contains
```
angelozerr, line 425: Please do that in `XMLCatalogPlugin#doSave`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/PathWarnings.java
xorye: fixed

xorye:
I made an update to this PR with these changes:
1. create `AbstractXMLNotification.java` which manages sending the notification to the client
2. validating catalogs paths and sending the notification to the client from `XMLCatalogPlugin.java`

In order to do 2., I've implemented a way to access `XMLLangaugeClientAPI` from `XMLExtensionsRegistry`. 

Please let me know if I'm (or not) on the right track with this PR. 

What this PR is missing:
- [ ] Javadoc
- [ ] Tests
- [ ] Improved notification message

xorye:
Current error message (should be improved):
image

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/AbstractXMLNotification.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Collections;
+
+import org.eclipse.lemminx.customservice.ActionableNotification;
+import org.eclipse.lemminx.customservice.XMLLanguageClientAPI;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lsp4j.Command;
+import org.eclipse.lsp4j.MessageParams;
+import org.eclipse.lsp4j.MessageType;
+
+public abstract class AbstractXMLNotification {
```
datho7561, line 13: I think it would make more sense to call this class `AbstractXMLNotifier`, since this generates a notification, but doesn't represent one itself.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/PathWarnings.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.lemminx.customservice.XMLLanguageClientAPI;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lsp4j.Command;
+
+public class PathWarnings extends AbstractXMLNotification {
```
datho7561, line 12: It might make more sense to name this class `PathWarner`, change `cache` to a `Set`, and store `settingId` as a attribute that is set in the constructor. Then you can make a different `PathWarner` for each setting you want to watch. I think this makes sense in this context, since the only time this class is being used, it is only tracking one setting.

xorye:
This PR has been updated with changes from https://github.com/eclipse/lemminx/pull/794.

Demo:
demo

In the gif above, the first entry is an existing file. The others are not.

fbricon:
You should use error messages

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/PathWarnings.java
xorye: > I think it would make more sense to call this class AbstractXMLNotifier, since this generates a notification, but doesn't represent one itself.
> It might make more sense to name this class PathWarner

I like this idea. This means `LimitExceededWarnings` should change to `LimitExceededWarner` too. I will change this soon.

> change cache to a Set, and store settingId as a attribute that is set in the constructor.

I realized that for `LimitExceededWarnings`, the cache is being shared for all features: https://github.com/eclipse/lemminx/blob/c9fa0cc44faa55c80c5d06a5824039be948eec3c/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/LimitExceededWarnings.java#L53

Right now, we only have one, so its not a problem. In my latest PR, I separated the cache for each feature using a map (key is the feature, value is a Set), but perhaps its a bit messy. 

If we want one feature = one instance, we must remove this method signature: https://github.com/eclipse/lemminx/blob/c9fa0cc44faa55c80c5d06a5824039be948eec3c/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/LimitExceededWarnings.java#L53
and provide the feature in the constructor as you said.

I quickly made a branch that implements that change: https://github.com/xorye/lsp4xml/tree/217catalogs_path_cleaner_cache

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/IXMLNotificationService.java
```diff
@@ -0,0 +1,11 @@
+package org.eclipse.lemminx.services;
+
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lsp4j.Command;
+
+public interface IXMLNotificationService {
```
angelozerr, line 6: Add Javadoc for class + methods.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/IXMLNotificationService.java
```diff
@@ -0,0 +1,11 @@
+package org.eclipse.lemminx.services;
```
angelozerr, line 1: Add header

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/XMLExtensionsRegistry.java
```diff
@@ -330,5 +333,13 @@ public void setDocumentProvider(IXMLDocumentProvider documentProvider) {
 	public URIResolverExtensionManager getResolverExtensionManager() {
 		return resolverExtensionManager;
 	}
+	
+	public IXMLNotificationService getNotificationService() {
```
angelozerr, line 337: Add Javadoc

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/XMLExtensionsRegistry.java
```diff
@@ -330,5 +333,13 @@ public void setDocumentProvider(IXMLDocumentProvider documentProvider) {
 	public URIResolverExtensionManager getResolverExtensionManager() {
 		return resolverExtensionManager;
 	}
+	
+	public IXMLNotificationService getNotificationService() {
+		return notificationService;
+	}
+	
+	public void setNotificationService(IXMLNotificationService notificationService) {
```
angelozerr, line 341: Add Javadoc

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/AbstractNotifierTest.java
```diff
@@ -0,0 +1,110 @@
+package org.eclipse.lemminx.services;
```
angelozerr, line 1: Add header

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/AbstractNotifierTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.eclipse.lemminx.client.ExtendedClientCapabilities;
+import org.eclipse.lemminx.customservice.ActionableNotification;
+import org.eclipse.lemminx.customservice.XMLLanguageClientAPI;
+import org.eclipse.lsp4j.ClientCapabilities;
+import org.eclipse.lsp4j.MessageActionItem;
+import org.eclipse.lsp4j.MessageParams;
+import org.eclipse.lsp4j.PublishDiagnosticsParams;
+import org.eclipse.lsp4j.ShowMessageRequestParams;
+
+public abstract class AbstractNotifierTest {
```
angelozerr, line 19: Add Javadoc

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/InvalidPathWarningsTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+import java.util.Arrays;
+import java.util.HashSet;
+
+import org.eclipse.lemminx.client.PathFeature;
+import org.eclipse.lemminx.client.PathWarnings;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class InvalidPathWarningsTest extends AbstractNotifierTest {
```
angelozerr, line 22: Add Javadoc

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/PathWarnings.java
angelozerr: @xorye I thing your design is very good now. The validation is processed in plugin + AbstractXMLNotification manages cache. 

what is the problem with  public void `onResultLimitExceeded(String uri, LimitFeature feature)` signature?

xorye: @angelozerr If we follow this suggestion:

> It might make more sense to name this class PathWarner, change cache to a Set, and store settingId as a attribute that is set in the constructor. 

Changing the cache type for `PathWarnings` would change the cache type for `LimitExceededWarnings` too. The cache would only store info for one feature, since the cache would be a `Set<String>` instead of `Map<String, Set<String>>`.

Therefore, it would make `LimitExceededWarnings` work only for one feature, which would change this method signature:
```onResultLimitExceeded(String uri, LimitFeature feature)```
to
```onResultLimitExceeded(String uri)``` 

So, there's no problem with `onResultLimitExceeded(String uri, LimitFeature feature)` right now in this PR, its just that if the suggestion is followed, it does not make sense to provide `feature` in  `onResultLimitExceeded(String uri, LimitFeature feature)`

xorye:
> You should use error messages

Fixed:
image

Everything has been squashed to one commit.

angelozerr:
Great job @xorye !

PR #760, opened 2020-06-02
Quickfix to close open tag doesn't deal with attributes

### LINKED ISSUE #646: Quickfix to close open tag doesn't deal with attributes
Try to apply any of the proposed quickfixes to 
```xml
<foo>
  <bar action="toot" /
</foo>
```
Apr-20-2020 16-59-03

None generate proper code. I reckon it might be tricky to fix though

angelozerr:
Quickfix to close open tag doesn't deal with attributes

Fixes #646

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseStartTagCodeAction.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseStartTagCodeAction.java
@@ -1,0 +1,152 @@
+/*******************************************************************************
+* Copyright (c) 2020 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.extensions.contentmodel.participants.codeactions;
+
+import java.util.List;
+
+import org.eclipse.lemminx.commons.BadLocationException;
+import org.eclipse.lemminx.commons.CodeActionFactory;
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.dom.DOMElement;
+import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.dom.LineIndentInfo;
+import org.eclipse.lemminx.services.extensions.ICodeActionParticipant;
+import org.eclipse.lemminx.services.extensions.IComponentProvider;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lemminx.utils.XMLPositionUtility;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+
+/**
+ * Code action to fix close start tag element.
+ *
+ */
+public class CloseStartTagCodeAction implements ICodeActionParticipant {
+
+	@Override
+	public void doCodeAction(Diagnostic diagnostic, Range range, DOMDocument document, List<CodeAction> codeActions,
+			SharedSettings sharedSettings, IComponentProvider componentProvider) {
+		Range diagnosticRange = diagnostic.getRange();
+		try {
+			int startOffset = document.offsetAt(diagnosticRange.getStart());
+			DOMNode node = document.findNodeAt(startOffset);
+			if (node != null && node.isElement()) {
+				int diagnosticEndOffset = document.offsetAt(diagnosticRange.getEnd());
+				DOMElement element = (DOMElement) node;
+				if (!element.hasStartTag()) {
+					// </foo>
+					DOMElement parent = element.getParentElement();
+					if (parent != null && parent.getTagName() != null) {
+						// <a><b></c>
+						// Replace with 'b' closing tag
+						String tagName = parent.getTagName();
+						Range replaceRange = XMLPositionUtility.selectEndTagName(element);
+						CodeAction replaceTagAction = CodeActionFactory.replace("Replace with '" + tagName + "' closing tag",
+								replaceRange, tagName, document.getTextDocument(), diagnostic);
+						codeActions.add(replaceTagAction);
+					}
+				} else {
+					// <foo>
+					boolean startTagClosed = element.isStartTagClosed();
+					char c = document.getText().charAt(diagnosticEndOffset - 1);
+					if (c != '/') {
+						if (startTagClosed) {
+							// ex : <foo attr="" >
+							// // Close with '</$tag>
+							String tagName = element.getTagName();
+							if (tagName != null) {
+								String label = "</" + tagName + ">";
+								String insertText = label;
+								Position endPosition = null;
+								if (!element.hasChildNodes()) {
+									int endOffset = element.getStartTagCloseOffset() + 1;
+									endPosition = document.positionAt(endOffset);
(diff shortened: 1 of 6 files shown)


angelozerr:
This PR improves code action to close element for:

 * root element
 * element which are not in root element
 * Insert end element after the child elements (the limitation is that it doesn't take care of format settings).

Here a demo:

AutoCloseCodeAction

fbricon:
when adding `</root>` to `<root><foo></foo>`, it should be inserted using the same indentation as its opening tag

angelozerr:
> when adding `</root>` to `<root><foo></foo>`, it should be inserted using the same indentation as its opening tag

You mean manage proper formatting, is that?

angelozerr:
> when adding </root> to <root><foo></foo>, it should be inserted using the same indentation as its opening tag

I need to fix test, but it should respect the line indentation.

angelozerr:
Tests are fixed

angelozerr:
Please don't merge, I have found a usecase which is totally wrong:

```xml
<a>
	<b>
		</c>
```

angelozerr:
> Please don't merge, I have found a usecase which is totally wrong:

fixed, you shouls have a replace code action for this usecase.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseStartTagCodeAction.java
```diff
@@ -0,0 +1,153 @@
...
+				DOMElement element = (DOMElement) node;
+				if (!element.hasStartTag()) {
+					// </foo>
+					DOMElement parent = element.getParentElement();
+					if (parent != null && parent.getTagName() != null) {
+						// <a><b></c>
+						// Replace "c" with "b"
+						String tagName = parent.getTagName();
+						Range replaceRange = XMLPositionUtility.selectEndTagName(element);
+						CodeAction replaceTagAction = CodeActionFactory.replace("Replace with '" + tagName + "' tag",
```
datho7561, line 56: I think it might make more sense to the user if the message was "Replace with 'b' closing tag"

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseStartTagCodeAction.java
```diff
@@ -0,0 +1,153 @@
...
+								}
+								CodeAction closeEndTagAction = CodeActionFactory.insert("Close with '" + label + "'",
+										endPosition, insertText, document.getTextDocument(), diagnostic);
+								codeActions.add(closeEndTagAction);
+							}
+
+						} else {
+							// ex : <foo attr="
+							// Close with '/>
+							CodeAction autoCloseAction = CodeActionFactory.insert("Close with '/>'",
```
datho7561, line 107: I found this case which isn't handled correctly:
```xml
<foo action="toot"
    /
```

double-backslash-closing-element

I think this option should be hidden for this case.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseStartTagCodeAction.java
```diff
@@ -0,0 +1,153 @@
...
+										// remove new line
+										char ch = text.charAt(endOffset);
+										while (ch == '\r' || ch == '\n') {
+											endOffset--;
+											ch = text.charAt(endOffset);
+										}
+									}
+									endOffset++;
+									endPosition = document.positionAt(endOffset);
+									if (hasElements(element)) {
```
xorye, line 90: I think this check should be if the last child is an element? Should we indent even if the last child is a text node?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseStartTagCodeAction.java
```diff
@@ -0,0 +1,153 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
```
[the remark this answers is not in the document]
angelozerr: fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseStartTagCodeAction.java
angelozerr: good catch, fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseStartTagCodeAction.java
angelozerr: I agree, fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseStartTagCodeAction.java
angelozerr: I have not a strong opinion about that. My idea was to manage new line for mixed content (when element to close contains an element). 

@fbricon what do you think about that?

I mean if we have

```xml
<a>
   bcd<e/>fgh
```

the code action should generate (my PR)

```xml
<a>
   bcd<e/>fgh
</a>
```

or (suggestion of @xorye )

```xml
<a>
   bcd<e/>fgh</a>
```

fbricon: indentation should match the opening tag's. So what the PR does

angelozerr: ok thanks for your feedback. I let my code like this.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSyntaxDiagnosticsTest.java
```diff
@@ -217,7 +278,7 @@ public void testETagRequired() throws Exception {
 				"			</Nm>  ";
 		Diagnostic d = d(1, 5, 1, 7, XMLSyntaxErrorCode.ETagRequired);
 		testDiagnosticsFor(xml, d);
-		testCodeActionsFor(xml, d, ca(d, te(1, 8, 1, 8, "</Nm>")));
+		testCodeActionsFor(xml, d, ca(d, te(2, 2, 2, 2, "</Nm>")));
```
xorye, line 281: From https://github.com/eclipse/lemminx/pull/760#discussion_r435431827, it seems like "adding" whitespace to the end of the text node is ok, so what about "removing" whitespace at the end of the textnode?

The behaviour defined in this test is:
```
<UltmtDbtr>
	<Nm>Name
	</UltmtDbtr>
		</Nm>
```

to

```
<UltmtDbtr>
	<Nm>Name
	</Nm></UltmtDbtr>
		</Nm>
```

But would it be better if we "remove" the whitespace and newline from `Name\n____` to have
```
<UltmtDbtr>
	<Nm>Name</Nm>
	</UltmtDbtr>
		</Nm>
```
instead? Ie, insert the codeaction content right after `Name` instead of `Name\n____`?

angelozerr: I agree, fixed

PR #761, opened 2020-06-03
TargetNamespace.1/2 error range fix

### LINKED ISSUE #703: Fix error range TargetNamespace.2
This issue occurs when xml-model support will be merged.

Given this XML :

```xml
<?xml-model href="http://www.docbook.org/xml/5.0/xsd/docbook.xsd"?>
<book> 
	<title />
  	<BAD_ELEMENT />
</book> 
```

We have this error:

image

We should:

  * fix error range, instead of highlight ?, we must highlight `book` root element.
 * provide a code action, to insert in the book the text `xmlns="http://docbook.org/ns/docbook"`

### LINKED ISSUE #704: Fix error range TargetNamespace.1
This issue occurs when xml-model support will be merged.

Given this XML :

```xml
<?xml-model href="http://www.docbook.org/xml/5.0/xsd/docbook.xsd"?>
<book xmlns="BAD_NS"> 
	<title />
  	<BAD_ELEMENT />
</book> 
```

We have this error:

image

We should:

  * fix error range, instead of highlight >, we must highlight `BAD-NS` attribute value of xmlns.
 * provide a code action, to replace `BAD_NS` with `http://docbook.org/ns/docbook`

datho7561:
If the declared namespace of the root element is not the same as
the namespace declared in the schema, the `xmlns` attribute of the root
element is highlighted.

If there is no namespace declared in the root element, but one is
declared in the schema, the root element is highlighted.

Added a CodeAction attached to `TargetNamespace.1` that replaces
the namespace declared in the instance with the namespace declared in
the schema.

Added a CodeAction attached to `TargetNamespace.2` that adds the
`xmlns` attribute to the root element and fills it with the namespace
used in the schema.

Fixes #704, Fixes #703

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/XMLSchemaErrorCode.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/XMLSchemaErrorCode.java
@@ -19,11 +19,14 @@ import java.util.Map;
 
 import org.apache.xerces.xni.XMLLocator;
 import org.eclipse.lemminx.commons.BadLocationException;
+import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.dom.DOMNode;
 import org.eclipse.lemminx.dom.NoNamespaceSchemaLocation;
 import org.eclipse.lemminx.dom.SchemaLocation;
+import org.eclipse.lemminx.extensions.contentmodel.participants.codeactions.TargetNamespace_1CodeAction;
+import org.eclipse.lemminx.extensions.contentmodel.participants.codeactions.TargetNamespace_2CodeAction;
 import org.eclipse.lemminx.extensions.contentmodel.participants.codeactions.cvc_attribute_3CodeAction;
 import org.eclipse.lemminx.extensions.contentmodel.participants.codeactions.cvc_complex_type_2_1CodeAction;
 import org.eclipse.lemminx.extensions.contentmodel.participants.codeactions.cvc_complex_type_2_3CodeAction;
@@ -74,6 +77,7 @@ public enum XMLSchemaErrorCode implements IXMLErrorCode {
 	cvc_maxInclusive_valid("cvc-maxInclusive-valid"), // https://wiki.xmldation.com/Support/validator/cvc-maxinclusive-valid
 	cvc_minExclusive_valid("cvc-minExclusive-valid"), // https://wiki.xmldation.com/Support/validator/cvc-minexclusive-valid
 	cvc_minInclusive_valid("cvc-minInclusive-valid"), // https://wiki.xmldation.com/Support/validator/cvc-mininclusive-valid
+	TargetNamespace_1("TargetNamespace.1"), //
 	TargetNamespace_2("TargetNamespace.2"), SchemaLocation("SchemaLocation"), schema_reference_4("schema_reference.4"), //
 	src_element_3("src-element.3");
 
@@ -138,7 +142,6 @@ public enum XMLSchemaErrorCode implements IXMLErrorCode {
 		case cvc_elt_1_a:
 		case cvc_complex_type_4:
 		case src_element_3:
-		case TargetNamespace_2:
 			return XMLPositionUtility.selectStartTagName(offset, document);
 		case cvc_complex_type_3_2_2: {
 			String attrName = getString(arguments[1]);
@@ -245,6 +248,10 @@ public enum XMLSchemaErrorCode implements IXMLErrorCode {
 		}
 		case cvc_type_3_1_2:
 			return XMLPositionUtility.selectStartTagName(offset, document);
+		case TargetNamespace_1:
+			return XMLPositionUtility.selectRootAttributeValue(DOMAttr.XMLNS_ATTR, document);
+		case TargetNamespace_2:
+			return XMLPositionUtility.selectRootStartTag(document);
 		default:
 		}
 		return null;
@@ -260,5 +267,7 @@ public enum XMLSchemaErrorCode implements IXMLErrorCode {
 		codeActions.put(cvc_complex_type_3_2_2.getCode(), new cvc_complex_type_3_2_2CodeAction());
 		codeActions.put(cvc_enumeration_valid.getCode(), new cvc_enumeration_validCodeAction());
 		codeActions.put(cvc_complex_type_2_1.getCode(), new cvc_complex_type_2_1CodeAction());
+		codeActions.put(TargetNamespace_1.getCode(), new TargetNamespace_1CodeAction());
+		codeActions.put(TargetNamespace_2.getCode(), new TargetNamespace_2CodeAction());
 	}
 }
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
@@ -1,0 +1,70 @@
+/**
+ * Copyright (c) 2020 Red Hat Inc. and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ * Red Hat Inc. - initial API and implementation
+ */
+package org.eclipse.lemminx.extensions.contentmodel.participants.codeactions;
+
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.regex.Matcher;
+
+import org.eclipse.lemminx.commons.CodeActionFactory;
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.services.extensions.ICodeActionParticipant;
+import org.eclipse.lemminx.services.extensions.IComponentProvider;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lemminx.utils.StringUtils;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.Range;
+
+/**
+ * CodeAction to replace an incorrect namespace in an .xml document.
+ * 
+ * Changes the value of the xmlns attribute of the root element of the .xml
+ * document to the declared namespace of the referenced .xsd document.
+ */
+public class TargetNamespace_1CodeAction implements ICodeActionParticipant {
+
+	private static final Pattern NAMESPACE_EXTRACTOR = Pattern.compile("'([^']+)'\\.");
+
+	@Override
+	public void doCodeAction(Diagnostic diagnostic, Range range, DOMDocument document, List<CodeAction> codeActions,
+			SharedSettings sharedSettings, IComponentProvider componentProvider) {
+
+		String namespace = extractNamespace(diagnostic.getMessage());
+		if (StringUtils.isEmpty(namespace)) {
+			return;
+		}
+		String quote = sharedSettings.getPreferences().getQuotationAsString();
+		// @formatter:off
+		CodeAction replaceNamespace = CodeActionFactory.replace(
+				"Replace with '" + namespace + "'",
+				diagnostic.getRange(),
+				quote + namespace + quote,
+				document.getTextDocument(),
+				diagnostic);
+		// @formatter:on
+		codeActions.add(replaceNamespace);
+	}
+
+	private static String extractNamespace(String diagnosticMessage) {
+		// The error message has this form:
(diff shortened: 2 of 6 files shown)


fbricon:
See if you can fix #703 too, as part of this PR

datho7561:
@fbricon I'll add a fix for it as well. Should I include the code snippets that were originally mentioned in this PR, or open a separate issue to fix them?

fbricon:
Huh which snippers are you referring to?

datho7561:
@fbricon Oops. Sorry, I meant CodeActions. Angelo mentioned adding CodeActions in the issues. Should I make separate issues for these, or add them in this PR?

fbricon:
We'll add code actions separately, in 0.13.0 or later

fbricon:
I see you already have 1 code action. We should then have the other one as well, to be consistent. Hopefully it should be fairly straightforward to add now

datho7561:
Still working on the tests for them, but here are demos of the `CodeActions`:
TargetNamespace 1-fix-declaration
TargetNamespace 2-add-declaration

angelozerr:
Nice demo!

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
```diff
@@ -0,0 +1,55 @@
...
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.services.extensions.ICodeActionParticipant;
+import org.eclipse.lemminx.services.extensions.IComponentProvider;
+import org.eclipse.lemminx.settings.EnforceQuoteStyle;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.Range;
+
+public class TargetNamespace_1CodeAction implements ICodeActionParticipant {
```
angelozerr, line 29: Add javadoc

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
```diff
@@ -0,0 +1,55 @@
...
+
+	private static final Pattern NAMESPACE_EXTRACTOR = Pattern.compile("'([^']+)'\\.");
+
+	@Override
+	public void doCodeAction(Diagnostic diagnostic, Range range, DOMDocument document, List<CodeAction> codeActions,
+			SharedSettings sharedSettings, IComponentProvider componentProvider) {
+		
+		Matcher namespaceMatcher = NAMESPACE_EXTRACTOR.matcher(diagnostic.getMessage());
```
angelozerr, line 37: please add comment which explain that you extract namespace from the message (by given a sample of meessage in the comments)

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_2CodeAction.java
```diff
@@ -0,0 +1,65 @@
...
+import org.eclipse.lemminx.services.extensions.IComponentProvider;
+import org.eclipse.lemminx.settings.EnforceQuoteStyle;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lemminx.utils.XMLPositionUtility;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+
+public class TargetNamespace_2CodeAction implements ICodeActionParticipant {
```
angelozerr, line 32: Add Javadoc

datho7561:
> Nice demo!

Thank you!

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSyntaxDiagnosticsTest.java
```diff
@@ -19,6 +19,7 @@
 import static org.eclipse.lemminx.XMLAssert.testDiagnosticsFor;
 
 import org.eclipse.lemminx.XMLAssert;
+import org.eclipse.lemminx.extensions.contentmodel.participants.XMLSchemaErrorCode;
```
angelozerr, line 22: remove this unused package.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLPositionUtility.java
```diff
@@ -325,6 +325,21 @@ public static Range selectRootStartTag(DOMDocument document) {
 		return selectStartTagName(root);
 	}
 
+	public static Range selectRootAttributeValue(String attrName, DOMDocument document) {
```
angelozerr, line 328: Add Javadoc

angelozerr:
@datho7561 please fix the conflict

datho7561:
> @datho7561 please fix the conflict



> @datho7561 please fix the conflict

Fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
```diff
@@ -0,0 +1,70 @@
...
+				document.getTextDocument(), diagnostic);
+		// @formatter:on
+		codeActions.add(replaceNamespace);
+	}
+
+	private static String extractNamespace(String diagnosticMessage) {
+		// The error message has this form:
+		// TargetNamespace.1: Expecting namespace 'NaN', but the target namespace of the
+		// schema document is 'http://two-letter-name'.
+		Matcher nsMatcher = NAMESPACE_EXTRACTOR.matcher(diagnosticMessage);
```
xorye, line 62: My worry is what if the namespace contains a `'` (not url encoded like %XX) character?

@angelozerr do we need to worry about this?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
```diff
@@ -0,0 +1,70 @@
...
+public class TargetNamespace_1CodeAction implements ICodeActionParticipant {
+
+	private static final Pattern NAMESPACE_EXTRACTOR = Pattern.compile("'([^']+)'\\.");
+
+	@Override
+	public void doCodeAction(Diagnostic diagnostic, Range range, DOMDocument document, List<CodeAction> codeActions,
+			SharedSettings sharedSettings, IComponentProvider componentProvider) {
+
+		String namespace = extractNamespace(diagnostic.getMessage());
+		String quote = "\"";
```
xorye, line 44: It should just be `String quote = sharedSettings.getPreferences().getQuotationAsString();` because the `sharedSettings.getFormattingSettings().getEnforceQuoteStyle();` is only for formatting requests.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
datho7561: Good catch. I looked at the specification to see if `'` is allowed in a namespace: https://www.w3.org/TR/xml/#NT-Name 
My understanding from this document is that it isn't.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public class TargetNamespace_1CodeAction implements ICodeActionParticipant {
+
+	private static final Pattern NAMESPACE_EXTRACTOR = Pattern.compile("'([^']+)'\\.");
+
+	@Override
+	public void doCodeAction(Diagnostic diagnostic, Range range, DOMDocument document, List<CodeAction> codeActions,
+			SharedSettings sharedSettings, IComponentProvider componentProvider) {
+
+		String namespace = extractNamespace(diagnostic.getMessage());
```
fbricon, line 42: if namespace is empty/null, you should return early

datho7561: This does make more sense as a fail-safe behaviour; I'll implement it.

It also made me think of this test case. The expected behaviour would be hard to implement, since `null` is a valid namespace:

replace-namespace-with-null

datho7561:
There are some issues with the build that I don't seem to have on my computer.

datho7561:
I ran the tests under a fresh Fedora 32 VM with openjdk 11, using the included maven wrapper and there were no failed tests.

angelozerr:
If you repush your branch with amend (to restart the CI build) is it working better?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
angelozerr: With this usecase, You have found a new bug with error range with EmptyTargetNamespace (see orange warning at the end of xml-model). You can create a bug for this issue.

I think it's a bug of Xerces which should generate a message with 'empty namespace' instead of null. You can create a second bug for this issue.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
angelozerr: Xerces doesn't complain when you write targetNamespace="'", so I think we should take care of even if it's not a common usecases.

datho7561: Working on it and unit tests for it

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/XMLSchemaErrorCode.java
```diff
@@ -245,6 +247,10 @@ public static Range toLSPRange(XMLLocator location, XMLSchemaErrorCode code, Obj
 		}
 		case cvc_type_3_1_2:
 			return XMLPositionUtility.selectStartTagName(offset, document);
+		case TargetNamespace_1:
+			return XMLPositionUtility.selectRootAttributeValue("xmlns", document);
```
angelozerr, line 251: Use DOMAttr.XMLNS_ATTR

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
datho7561: I implemented it so that it will work with `'` and escapes it when it occurs, but it doesn't recognize the escaped version of `'`, so it still raises `TargetNamespace.1`:

DoesntFixErrorQuotes

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
datho7561: I created GitHub issues for these issues.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
datho7561: I've pushed the code that implements the above (with tests) into this PR. Right now, it always tries to escape single quotes. I left it as a separate commit in case it we want to revert it. Other options could be:

- Always use double quotes when declaring `xmlns` if there is a single quote embedded in the `targetNamespace`
- Don't give a code action if single quotes are prefered and there are single quotes in the `targetNamespace`

datho7561:
I've reverted my changes that try to resolve single quotes in the namespace. This PR is ready for review.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/TargetNamespace_1CodeAction.java
datho7561: ^This has been reverted back to the code that doesn't deal with single quotes in the namespace

Issue #675, opened 2020-05-04
XML format for bad processing instruction removes following element


angelozerr:
Given this XML content (the processing instruction is not closed). 

```xml
<?xml version="1.0" encoding="UTF-8"?
<a /> 
```

After formatting, processing instruction is closed (I'm not sure if we need to close it?) but the annoying behavior is that `<a />` disappear.

```xml
<?xml version="1.0" encoding="UTF-8"?>
```

angelozerr:
This issue should be fixed by the experimental formatter https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md#xmlformatexperimental

PR #766, opened 2020-06-07
Prevent formatting when invalid content exists inside tags

### LINKED ISSUE #679: Lone quote inside tag leads to catastrophic formatting
When formatting a document containing a single quote inside a tag, everything right to the quote disappears:

May-06-2020 00-23-37

xorye:
Fixes #679 and #675 

In `DOMNode`, his PR introduces a new boolean property: `containsUnknownContent`. The formatter will read this value and will abort formatting when it reaches a `DOMNode` for which `containsUnknownContent` is true.

Content that comes before the element, will still be formatted:
demo

The main goal of this PR is to not lose information when formatting.

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMNode.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMNode.java
@@ -64,6 +64,7 @@ public abstract class DOMNode implements Node, DOMRange {
 
 	private XMLNamedNodeMap<DOMAttr> attributeNodes;
 	private XMLNodeList<DOMNode> children;
+	private boolean containsUnknownContent = false;
 
 	final int start; // |<root> </root>
 	int end; // <root> </root>|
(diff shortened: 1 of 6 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMNode.java
```diff
@@ -843,4 +844,24 @@ public Object setUserData(String arg0, Object arg1, UserDataHandler arg2) {
...
+	public boolean isContainsUnknownContent() {
+		return containsUnknownContent;
+	}
+
+	/**
+	 * Set containsMalformedContent
+	 *
+	 * @param containsMalformedContent
+	 */
+	public void setContainsUnknownContent(boolean containsUnknownContent) {
```
angelozerr, line 863: remove public

rgrunber:
@angelozerr , I think this can be closed now that we have the experimental formatter. See https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md#xmlformatexperimental for details.

angelozerr:
Indeed, we can close it.

PR #770, opened 2020-06-08
XML Scanner with quote

angelozerr:

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/parser/XMLScanner.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/parser/XMLScanner.java
@@ -85,7 +85,7 @@ public class XMLScanner implements Scanner {
 		return !Character.isWhitespace(ch) && ch != _QMA && ch != _DQO && ch != _SIQ && ch != _LAN && ch != _RAN
 				&& ch != _FSL && ch != _EQS && !(ch >= 0x00 && ch <= 0x0F) && ch != 0x7F && !(ch >= 0x80 && ch <= 0x9F);
 	};
-	
+
 	MultiLineStream stream;
 	ScannerState state;
 	int tokenOffset;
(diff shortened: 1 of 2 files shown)


xorye:
One small problem I found is that the DOMDocument will lose text content `hello` in this case:
```
<a attr "value">
    hello
</a>
```

In master, the `"value"` is read as an attribute value. After formatting, the formatter inserts a delimiter:
image

In this PR, the `"value"` is skipped, and the scanner advances until it reaches the next `<` character. Therefore element `a` is being read as an empty tag, therefore after formatting we get: image

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/parser/XMLScanner.java
```diff
@@ -366,24 +363,17 @@ TokenType internalScan() {
...
+				state = ScannerState.WithinContent;
+				if (stream.advanceIfChar(_RAN)) { // >
 					return finishToken(offset, TokenType.StartTagSelfClose);
 				}
-				return finishToken(offset, TokenType.Unknown);
-			}
-			int c = stream.peekChar();
-			if (c == _DQO || c == _SIQ) { // " || '
-				state = ScannerState.BeforeAttributeValue;
-				return internalScan();
```
xorye, line 386: To prevent the issue I mentioned, I was thinking about adding something like this here:

```
int c = stream.peekChar();
if (c == _DQO || c == _SIQ) { // " || '
    int i = 1;
    int curr = stream.peekChar(i);
    while (curr != -1) {
        if (curr == c) {
            state = ScannerState.BeforeAttributeValue;
            return internalScan();
        }
        if (curr == _LAN) {
            break;
        }
        curr = stream.peekChar(i++);
    }
}
```

This is under the assumption that the `<` character cannot exist in an attribute value, and that `<` is always being interpreted as a start of a new element

angelozerr: To be honnest with you, I tried to do the same code than HTML Language Service but there are some limitations too. I think we should write more tests with tolerant parser (liek your case) and we will see what we must to do. 

In other words, this PR should be done and improved after the release.

fbricon:
Investigation closed

Issue #272, opened 2018-12-21
DTD ATTLIST parsing needs to consider Enumerated Type


NikolasKomonen:
```xml
<!ATTLIST graphicsFile
	type (jpeg | postscript | gif | pdf | misc) #REQUIRED
	href ENTITY #REQUIRED
>
```
Refer to: https://www.w3.org/TR/xml/#NT-EnumeratedType

PR #772, opened 2020-06-08
Preserve attribute line breaks

xorye:


xorye:
This PR is ready for review.

Instead of introducing a formatting setting that keeps all newlines from non-empty lines, this PR introduces a new formatting setting, (ie, in VS Code: `xml.format.preserveAttributeLineBreaks`) that selectively keeps newlines from non-empty lines that come before and after element attributes.

For example:
```
<a \nattribute=\n"value"\n></a>
    ^           ^        ^
    1           2        3
```
The formatter will keep `1` because it comes before an attribute.
The formatter will lose `2` because it exists inside an attribute.
The formatter will keep `3` because it comes after an attribute.

These characteristics are useful in these cases:

End tag closing bracket is in a new line
```
<a attr="value"></a
>
```
formats to
```
<a attr="value"></a>
```

Attribute name/delimiter/value on separate lines
```
<a attr
=
"value"></a>
```
formats to
```
<a attr="value"></a>
```

As a plus, this setting will not will not keep newlines inside of the XML prolog (`<?xml version="1.0" encoding="UTF-8"?>`) because the XML prolog must be in one line.

To review this PR with VS Code, please also check out https://github.com/redhat-developer/vscode-xml/pull/271, which adds the `xml.format.preserveAttributeLineBreaks` setting:
image

Demo:
```
<a>
    <li attribute="123456789" attribute="123456789" attribute="123456789"
        attribute="123456789" attribute="123456789" attribute="123456789"
        attribute="123456789" attribute="123456789" attribute="123456789"
        attribute="123456789" attribute="123456789" attribute="123456789"
    />
</a>
```
demo

xorye:
The gif is a bit misleading since the attribute names are all aligned since the tab size is 4 spaces.

This PR will indent the attribute names depending on the element's indentation and will not align attribute names for the moment. I created a new issue for attribute aligning here: #272

fbricon:
The closing bracket is moved to the previous line, it wasn't supposed to happen
Jun-10-2020 13-00-49

PR #777, opened 2020-06-11
Sort snippet

### LINKED ISSUE #692: Sort snippet
Ex : xml declaration should be before other snippets.

An idea is to add a sortText field in the  json snippet.

angelozerr:
Sort snippet

Fixes #692

### DIFF
--- a/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/xml-declaration-snippets.json
+++ b/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/xml-declaration-snippets.json
@@ -8,7 +8,8 @@
       "<?xml version=\"${1|1.0,1.1|}\" encoding=\"${2|UTF-8,ISO-8859-1,Windows-1251,Windows-1252,Shift JIS,GB2312,EUC-KR|}\"?>${0}"
     ],
     "label": "$description",
-    "description": "Insert XML Declaration"
+    "description": "Insert XML Declaration",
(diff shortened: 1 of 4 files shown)


angelozerr:
To test this PR, create an empty XML file and open completion, XML declaration snippet should appear on the top:

image

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/xml-declaration-snippets.json
```diff
@@ -19,6 +20,7 @@
       "<?xml version=\"${1|1.0,1.1|}\" encoding=\"${2|UTF-8,ISO-8859-1,Windows-1251,Windows-1252,Shift JIS,GB2312,EUC-KR|}\" standalone=\"${3|yes,no|}\"?>${0}"
     ],
     "label": "$description",
-    "description": "Insert XML Declaration with standalone"
+    "description": "Insert XML Declaration with standalone",
+	"sortText": "0"
```
xorye, line 24: Very minor question, but is there a reason why both `sortText` are `0` instead of one being `0` and another being `1`?

angelozerr: My idea was to group the snippet per type (here the type is XML declaration). Less we have values for sortText less it will be easier to manage sort.

xorye:
This PR works well & looks good

PR #780, opened 2020-06-11
Fix line break being incorrectly added when preserveAttrLineBreaks is

xorye:
While `xml.format.preserveAttributeLineBreaks` is enabled, in this case:
```
<a attr="value" attr="value"
    attr=
    "value" attr="value">
</a>
```

in master, formatting results in:
```
<a attr="value" attr="value"
    attr="value"
    attr="value">
</a>
```

There shouldn't be a newline after the 2nd line.

This PR fixes this problem. As a result, formatting now results in:
```
<a attr="value" attr="value"
    attr="value" attr="value">
</a>
```

Sorry, this should have been caught yesterday


datho7561:
I get
```xml
<a attr="value" attr="value"
    attr="value" attr="value"></a>
```
for
```xml
<a attr="value" attr="value"
    attr=
    "value" attr="value">
</a>
```

xorye:
Yes, that's correct behaviour

PR #785, opened 2020-06-16
Implement file path completion for DTD systemId

### LINKED ISSUE #738: Provide file completion support for local DTDs
When a DTD is referenced by a local path, completions for local files should be proposed
```xml
<!DOCTYPE foo SYSTEM "./|">
<foo>
</foo>
```

Similar to XSD local locations:

Screen Shot 2020-05-28 at 5 12 49 PM

xorye:
Fixes #738 

```
<!DOCTYPE foo SYSTEM "">
```

demo

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/XSISchemaModel.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/XSISchemaModel.java
@@ -167,6 +167,9 @@ public class XSISchemaModel {
 		
 		String actualPrefix = document.getSchemaInstancePrefix();
 		DOMAttr attrAtOffset = nodeAtOffset.findAttrAt(offset);
+		if (attrAtOffset == null) {
+			return;
+		}
 		String attrName = attrAtOffset.getName();
 		
 		if(attrName != null) {
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
@@ -31,6 +31,7 @@ import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.dom.DTDDeclParameter;
 import org.eclipse.lemminx.dom.parser.Scanner;
 import org.eclipse.lemminx.dom.parser.ScannerState;
 import org.eclipse.lemminx.dom.parser.TokenType;
@@ -122,6 +123,24 @@ public class XMLCompletions {
 				case DelimiterAssign:
 					if (scanner.getTokenEnd() == offset) {
 						collectAttributeValueSuggestions(offset, offset, completionRequest, completionResponse);
+						return completionResponse;
+					}
+					break;
+				case DTDStartDoctypeTag:
+
+					DTDDeclParameter systemId = xmlDocument.getDoctype().getSystemIdNode();
+					if (systemId == null) {
+						break;
+					}
+
+					if (DOMNode.isIncluded(systemId, offset)) {
+						/**
+						 * Completion invoked within systemId parameter
+						 * ie, completion offset is at | like so:
+						 * <!DOCTYPE foo SYSTEM "./|">
+						 */
+						collectDTDSystemIdSuggestions(systemId.getStart(), systemId.getEnd(),
+								completionRequest, completionResponse);
 						return completionResponse;
 					}
 					break;
@@ -830,7 +849,7 @@ public class XMLCompletions {
 			int valueContentStart = valueStart + 1;
 			int valueContentEnd = valueEnd;
 			// valueEnd points to the char after quote, which encloses the replace range
-			if (valueEnd > valueStart && text.charAt(valueEnd - 1) == text.charAt(valueStart)) {
+			if (text.charAt(valueEnd - 1) == text.charAt(valueStart)) {
 				valueContentEnd--;
 			}
 			valuePrefix = offset >= valueContentStart && offset <= valueContentEnd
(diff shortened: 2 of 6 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
```diff
@@ -125,6 +126,17 @@ public CompletionList doComplete(DOMDocument xmlDocument, Position position, Sha
 						return completionResponse;
 					}
 					break;
+				case DTDStartDoctypeTag:
+					DTDDeclParameter systemId = xmlDocument.getDoctype().getSystemIdNode();
+					if (systemId == null) {
+						break;
+					}
+					if (systemId.getStart() <= offset && offset <= systemId.getEnd()) {
+						collectAttributeValueSuggestions(systemId.getStart(), systemId.getEnd(),
```
xorye, line 135: Although the systemId parameter is not an attribute, I decided to call `collectAttributeValueSuggestions` since file path completion for attribute values were already implemented in `FilePathCompletionParticipant#onAttributeValue()`.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/XSISchemaModel.java
```diff
@@ -167,6 +167,9 @@ public static void computeValueCompletionResponses(ICompletionRequest request,
 		
 		String actualPrefix = document.getSchemaInstancePrefix();
 		DOMAttr attrAtOffset = nodeAtOffset.findAttrAt(offset);
+		if (attrAtOffset == null) {
+			return;
+		}
```
xorye, line 172: The problem with calling `onAttributeValue()` though, is that the request range's offset does not correspond to a `DOMAttr`. Therefore every completion participant that implements `onAttributeValue()` must be careful.

If this is a problem, maybe its a better idea to add new API in `CompletionParticipantAdapter`?  Something along the lines of `onDTDContent()`?

xorye:
An enhancement for this feature would be to provide folders and `.dtd` files in the completion options. Currently, the completion options will show any file `.dtd`, `.xsd`, `.txt`, etc.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
```diff
@@ -125,6 +126,17 @@ public CompletionList doComplete(DOMDocument xmlDocument, Position position, Sha
 						return completionResponse;
 					}
 					break;
+				case DTDStartDoctypeTag:
+					DTDDeclParameter systemId = xmlDocument.getDoctype().getSystemIdNode();
+					if (systemId == null) {
+						break;
+					}
+					if (systemId.getStart() <= offset && offset <= systemId.getEnd()) {
```
angelozerr, line 134: I think you could replace with `DOMNode.isIncluded(systemId, offset)`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
xorye: @angelozerr @datho7561 actually, maybe it might be better to move the contents of https://github.com/eclipse/lemminx/blob/59c04ef37c7959d95044f61564939c2f139b2d06/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java#L48

into a new static method and have:
```
case DTDStartDoctypeTag:
...
...
call static method
```
instead of
```
case DTDStartDoctypeTag:
...
...
collectAttributeValueSuggestions()
```

xorye: I will make a new commit for this

angelozerr: You mean that XMLCompletion (core) will depends on FilePathCompletionParticipant(extension). I would liek to avoid doing that.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/XSISchemaModel.java
angelozerr: Yes it should be a better idea to do that.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
angelozerr: I think the clean mean is to have a onDTDSystemId

angelozerr:
> An enhancement for this feature would be to provide folders and .dtd files in the completion options. Currently, the completion options will show any file .dtd, .xsd, .txt, etc.

I agree but I think a very clean mean is to mark attribute for file patch completion type (ex : schemaLocation must shows only .xsd)

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
xorye: Ok thank you, I created a new commit that creates onDTDSystemId. If it looks good, I will squash my two commits into one

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
xorye: fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/ICompletionParticipant.java
```diff
@@ -27,5 +27,8 @@ void onAttributeName(boolean generateValue, ICompletionRequest request, IComplet
 
 	void onAttributeValue(String valuePrefix, ICompletionRequest request, ICompletionResponse response)
 			throws Exception;
+	
+	void onDTDSystemId(String valuePrefix, ICompletionRequest request, ICompletionResponse response)
```
angelozerr, line 31: Add Javadoc

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
```diff
@@ -125,6 +126,18 @@ public CompletionList doComplete(DOMDocument xmlDocument, Position position, Sha
...
 					}
 					break;
+				case DTDStartDoctypeTag:
+					DTDDeclParameter systemId = xmlDocument.getDoctype().getSystemIdNode();
+					if (systemId == null) {
+						break;
+					}
+
+					if (DOMNode.isIncluded(systemId, offset)) {
+						collectDTDSystemIdSuggestions(systemId.getStart(), systemId.getEnd(),
```
angelozerr, line 136: Add a comment with sample DTD system (and with | to show where is the completion offset)

xorye: fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/ICompletionParticipant.java
xorye: fixed

angelozerr:
It works great but it can be improved for:

 * filter files with dtd file extension.
 * apply should replace the full attribute value.

To do that we must implement https://github.com/eclipse/lemminx/issues/786

PR #787, opened 2020-06-18
noNamespaceSchemaLocation file missing CodeAction

### LINKED ISSUE #691: CodeAction to fix bad value for noNamespaceSchemaLocation
Given this XML file:

```xml
<root-element
  xsi:noNamespaceSchemaLocation="file.xsd">
  
</root-element>

```

and file.xsd which doesn't exists:

image

We should have code actions:

 * replace with similar file name 
 * create the file.xsd and generate the file content by using the XML content. We  need a generate of XSD based on XML which could use too in https://github.com/eclipse/lemminx/issues/151

angelozerr:
A CodeAction for the error `schema-reference.4` that
creates the file referenced in
`xsi:noNamespaceSchemaLocation` if its missing.

Closes #691

### DIFF
--- a/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
+++ b/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
@@ -13,9 +13,13 @@
 package org.eclipse.lemminx.extensions.contentmodel;
 
 import static org.eclipse.lemminx.XMLAssert.ca;
+import static org.eclipse.lemminx.XMLAssert.createFile;
 import static org.eclipse.lemminx.XMLAssert.d;
 import static org.eclipse.lemminx.XMLAssert.te;
+import static org.eclipse.lemminx.XMLAssert.teOp;
 import static org.eclipse.lemminx.XMLAssert.testCodeActionsFor;
+
+import java.util.Arrays;
 
 import org.eclipse.lemminx.XMLAssert;
 import org.eclipse.lemminx.extensions.contentmodel.participants.XMLSchemaErrorCode;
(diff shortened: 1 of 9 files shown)


org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
```diff
@@ -643,6 +650,52 @@ public void testTargetNamespace_2SingleQuotes() throws Exception {
...
+		WorkspaceClientCapabilities workspace = new WorkspaceClientCapabilities();
+		WorkspaceEditCapabilities workspaceEdit = new WorkspaceEditCapabilities();
+		workspaceEdit.setResourceOperations(Arrays.asList(ResourceOperationKind.Create));
+		workspace.setWorkspaceEdit(workspaceEdit);
+		settings.getWorkspaceSettings().setCapabilities(workspace);
+
+		XMLAssert.testCodeActionsFor(xml, missingSchema, settings,
+				ca(missingSchema,
+				createFile("/salad.xsd", false), 
+				teOp("/salad.xsd", 0, 0, 0, 0, "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<xs:schema xmlns:xs=\"http://www.w3.org/2001/XMLSchema\">\n\n</xs:schema>")));
```
datho7561, line 678: I needed to change both occurrences of `"/salad.xsd"` to `"file:///salad.xsd"` in order to get this to pass under Linux. I think we need to adapt some of the test functions to work cross platform.

angelozerr: indeed, in Windows OS too, fixed

PR #799, opened 2020-06-22
Invalid location hint in xsi:schemaLocation

### LINKED ISSUE #782: Highlight error for bad xsd file in xsi:schemaLocation with the xsd file range
Given this XML :

```xml
<note xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	  xmlns="https://www.w3schools.com"
	  xsi:schemaLocation="https://www.w3schools.com XXXXXXXXX.xsd">
	<to></to>
	<from></from>
	<heading></heading>
	<body></body>
</note>
```

The `XXXXXXXXX.xsd` from xsi:schemaLocation="https://www.w3schools.com XXXXXXXXX.xsd" must be a highlighted as an error instead of the full attribute value.

Pay attention that xsi:schemaLocation can contain several xsd .

@datho7561 are you interested to work on that since you have fixed https://github.com/eclipse/lemminx/issues/666

datho7561:
When non-existent schemas are referenced using `xsi:schemaLocation`,
the location hint is now highlighted instead of the entire attribute
value.

Fixes #782

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/SchemaLocation.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/SchemaLocation.java
@@ -44,11 +44,10 @@ public class SchemaLocation {
 				break;
 			}
 			DOMNode valNode = attr.getNodeAttrValue();
-			// matcher matches 1 char ahead, but valNode's start is at beginning of "
-			int start = valNode.getStart() + locPairMatcher.start(2);
-			// matcher matches end correctly, but range needs to be one past end
-			// to highlight the end, and one to make up for "
-			int end = valNode.getStart() + locPairMatcher.end(2) + 2;
+			// http://example.org/schema/root |root.xsd http://example.org/schema/bison bison.xsd
+			int start = valNode.getStart() + locPairMatcher.start(2) + 1;
+			// http://example.org/schema/root root.xsd| http://example.org/schema/bison bison.xsd
+			int end = valNode.getStart() + locPairMatcher.end(2) + 1;
 			schemaLocationValuePairs.put(namespaceURI,
 					new SchemaLocationHint(start, end, locationHint, this));
 		}
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/XMLSchemaErrorCode.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/XMLSchemaErrorCode.java
@@ -17,6 +17,8 @@ import static org.eclipse.lemminx.utils.StringUtils.getString;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.xerces.impl.XMLEntityManager;
+import org.apache.xerces.util.URI.MalformedURIException;
 import org.apache.xerces.xni.XMLLocator;
 import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMDocument;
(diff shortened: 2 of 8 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/SchemaLocation.java
```diff
@@ -44,11 +44,8 @@ public SchemaLocation(DOMAttr attr) {
 				break;
 			}
 			DOMNode valNode = attr.getNodeAttrValue();
-			// matcher matches 1 char ahead, but valNode's start is at beginning of "
-			int start = valNode.getStart() + locPairMatcher.start(2);
-			// matcher matches end correctly, but range needs to be one past end
-			// to highlight the end, and one to make up for "
-			int end = valNode.getStart() + locPairMatcher.end(2) + 2;
+			int start = valNode.getStart() + locPairMatcher.start(2) + 1;
```
angelozerr, line 47: Please add a comments with String which uses | to understand what is the start and the end.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/XMLSchemaErrorCode.java
```diff
@@ -185,6 +189,20 @@ public static Range toLSPRange(XMLLocator location, XMLSchemaErrorCode code, Obj
 				} else {
 					SchemaLocation schemaLocation = document.getSchemaLocation();
 					if (schemaLocation != null) {
+						String invalidSchemaPath = arguments[0] instanceof String ? (String) arguments[0] : null;
+						
+						if (invalidSchemaPath != null) {
+							for (SchemaLocationHint locHintRange : schemaLocation.getSchemaLocationHints()) {
+								try {
+									String expandedHint = XMLEntityManager.expandSystemId(locHintRange.getHint(), document.getDocumentURI(), false);
```
angelozerr, line 197: create a private static methd getResolvedLocation (it should help to fix https://github.com/eclipse/lemminx/issues/801)

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLPositionUtility.java
```diff
@@ -997,6 +997,21 @@ public static DocumentLink createDocumentLink(DOMRange target, String location)
...
 
+	/**
+	 * Create a document link without adjusting the range to account for quotes.
+	 * 
+	 * @param target The range in the document that should be the link
+	 * @param location URI where the link should point
+	 * @return A DocumentLink over the range <code>target</code> that points to <code>location</code>.
+	 * @throws BadLocationException if <code>target</code> is out of the bounds of the document
+	 */
+	public static DocumentLink createDocumentLinkNoAdjust(DOMRange target, String location) throws BadLocationException {
```
angelozerr, line 1008: I will update the existing createDocumentLink to `createDocumentLink(DOMRange target, String location, boolean adjust)` and createDocumentLink((DOMRange target, String location) will call `createDocumentLink(target, location, true)` to share the same code.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLPositionUtility.java
```diff
@@ -983,17 +983,28 @@ public static Location createLocation(DOMRange target) {
...
+	 * @throws BadLocationException if <code>target</code> is out of the bounds of
+	 *                              the document
 	 */
-	public static DocumentLink createDocumentLink(DOMRange target, String location) throws BadLocationException {
+	public static DocumentLink createDocumentLink(DOMRange target, String location, boolean adjust)
+			throws BadLocationException {
 		DOMDocument document = target.getOwnerDocument();
-		Position start = document.positionAt(target.getStart() + 1);
-		Position end = document.positionAt(target.getEnd() - 1);
+		Position start, end;
```
angelozerr, line 1000: Position start  = document.positionAt(target.getStart() + adjust ? 1 : 0);

xorye:
Works well for me. After https://github.com/eclipse/lemminx/pull/799#discussion_r444257245 is addressed, I am good with merging this PR.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLPositionUtility.java
datho7561: I was thinking about implementing it this way, but then I would have to check `adjust` twice instead of once. Do you think this is worthwhile trade-off for readability?

angelozerr: checking a boolean is fast. Me I prefer doing that because the code is little and if you need to change target.getStart() with other things you change just in this line and not in the 2 line code.

angelozerr:
@datho7561 please fix conflicts.

datho7561:
> @datho7561 please fix conflicts.

Should be up to date now

angelozerr:
Good job @datho7561 !

PR #802, opened 2020-06-23
Fix catalog test on Windows OS

angelozerr:
Fix catalog test on Windows OS


angelozerr:
@xorye on Windows OS test fails with catalog. Here the fix. Please merge it.

xorye:
@angelozerr works on my machine

PR #803, opened 2020-06-23
Added catalog snippets

### LINKED ISSUE #708: New snippet to generate a catalog
catalog prefix ->
```xml
<catalog xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog">
    <public publicId="" uri="" />
</catalog>
```

datho7561:
Added two snippets that will generate a catalog.
One binds to the catalog grammar using DTD, and
the other binds with XSD.

Fixes #708

### DIFF
--- a/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/catalog-snippets.json
+++ b/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/catalog-snippets.json
@@ -1,0 +1,43 @@
+{
+  "New catalog bound using DTD": {
+    "prefix": "<catalog",
+    "suffix": ">",
+    "body": [
+      "<!DOCTYPE catalog",
+      "\tPUBLIC \"-//OASIS//DTD Entity Resolution XML Catalog V1.0//EN\"",
+      "\t\"http://www.oasis-open.org/committees/entity/release/1.0/catalog.dtd\">",
+      "<catalog xmlns=\"urn:oasis:names:tc:entity:xmlns:xml:catalog\" prefer=\"public\">",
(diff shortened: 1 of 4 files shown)


angelozerr:
The PR looks great. I think we should add a snippet for catalog without DTD / XSD : 

```xml
<?xml version="1.0"?>
  <catalog xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog">

    <system systemId="http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd"
            uri="dtd/xhtml1/xhtml1-strict.dtd"/>

    <system systemId="http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"
            uri="dtd/xhtml1/xhtml1-transitional.dtd"/>

    <system systemId="http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd"
            uri="dtd/xhtml11/xhtml11-flat.dtd"/>

  </catalog>
```

LemMinx is available to resolve  xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog" and bound id to the DTD catalog

org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/catalog-snippets.json
```diff
@@ -0,0 +1,43 @@
+{
+  "New catalog bound using DTD": {
+    "prefix": "<catalog",
```
xorye, line 3: I think the prefix should be `<!DOCTYPE`? @angelozerr what do you think?

angelozerr:
It works like a charm, thanks @datho7561 !

PR #815, opened 2020-07-03
Snippets for xml-stylesheet processing instruction.

### LINKED ISSUE #728: xml-stylesheet snippet
Manage snippet for generating xml-stylesheet processing instruction.

To do that:

 * rename `prolog-snippets.json` to `processing-instruction-snippets.json`
 * write xml-stylesheet snippets in `processing-instruction-snippets.json`.

BalduinLandolt:
Add snippets for xml-stylesheet processing instruction (`<?xml-stylesheet ...?>`).

Fixes #728

Signed-off-by: Balduin Landolt <balduin.landolt@hotmail.com>

### DIFF
--- a/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
+++ b/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/processing-instruction-snippets.json
@@ -8,7 +8,8 @@
 		"<?xml-model href=\"${1:file.xsd}\" type=\"application/xml\" schematypens=\"http://www.w3.org/2001/XMLSchema\"?>${0}"
 	  ],
 	  "label": "Insert XML Schema association",
-	  "description": "Insert XML Model Processing Instruction to Associate XSD Schema"
+	  "description": "Insert XML Model Processing Instruction to Associate XSD Schema",
+	  "sortText": "model"
 	},
 	"Insert XML Model: DTD": {
 		"prefix": [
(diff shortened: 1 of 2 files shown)


BalduinLandolt:
It works, I guess. Two questions, however:
1. in the tests (`XMLCompletionSnippetsTest.java`), do I need to add all the `c("Instert ..." [etc])` to each `testCompletionFor()`?
1. should I adress the order of the snippets (with `"sortText": ...`)?  
Alphabetic order seems suboptimal here:  
grafik  
I'd rather group xml-model and xml-stylesheet.

angelozerr:
> I'd rather group xml-model and xml-stylesheet.

Indeed, to do that you can use sortText declared in snippet (see https://microsoft.github.io/language-server-protocol/specification#textDocument_completion).

BalduinLandolt:
> Indeed, to do that you can use sortText declared in snippet (see https://microsoft.github.io/language-server-protocol/specification#textDocument_completion).

Done. Looks better now:  
grafik

angelozerr:
> Done. Looks better now:

Definitely, thanks!

Just one question I wonder if we could have several xml-stylesheet processing with the same type in XML document? if you think you can manage that it's fine, otherwise I'm OK to merge it.

BalduinLandolt:
> Just one question I wonder if we could have several xml-stylesheet processing with the same type in XML document? if you think you can manage that it's fine, otherwise I'm OK to merge it.

How do you mean exactly?  
Are you asking if it is supposed to be possible to have multiple? Or are you asking if we should have a snippet to add multiple at once?

As far as I'm aware, nothing speaks against having multiple stylesheets (see the examples here: https://www.w3.org/TR/2010/REC-xml-stylesheet-20101028/ ).

Of course I could have added attribtes like title and alternate, but I don't think they are commonly used, so I left them out. If you think they should be added, that's done quickly.

angelozerr:
> Are you asking if it is supposed to be possible to have multiple?

Yes I mean that.

> As far as I'm aware, nothing speaks against having multiple stylesheets (see the examples here: https://www.w3.org/TR/2010/REC-xml-stylesheet-20101028/ ).

yes in this case perhaps we should improve the filter to check we have one xml-stylesheet?

> Of course I could have added attribtes like title and alternate, but I don't think they are commonly used, so I left them out. If you think they should be added, that's done quickly.

Indeed I agree with you, too many snippets will kill the snippets powerful

BalduinLandolt:
> yes in this case perhaps we should improve the filter to check we have one xml-stylesheet?

Well, as it is now, it allows for multiple stylesheets, which is what we want, I think.
 
> Indeed I agree with you, too many snippets will kill the snippets powerful

Ok. In that case I think it's good to merge.

angelozerr:
Thanks @BalduinLandolt !

PR #834, opened 2020-08-05
Don't send invalid catalog notifications for paths with file scheme

xorye:
Fixes https://github.com/redhat-developer/vscode-xml/issues/289

The reason why completion from the catalog was working even though there was an error notification was because `XMLCatalogResolverExtension` used `new URI(catalogFile).getPath()` while `XMLCatalogPlugin` used `new URI(catalogFile).toString()`.

I removed `isXMLCatalogFileValid()` from `XMLCatalogResolverExtension` and `XMLCatalogPlugin`, and moved it into a utils class.

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/catalog/XMLCatalogPlugin.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/catalog/XMLCatalogPlugin.java
@@ -12,21 +12,20 @@
  */
 package org.eclipse.lemminx.extensions.catalog;
 
-import java.io.File;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.nio.file.Files;
 import java.util.Arrays;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import org.eclipse.lemminx.client.PathFeature;
 import org.eclipse.lemminx.client.InvalidPathWarner;
+import org.eclipse.lemminx.client.PathFeature;
 import org.eclipse.lemminx.extensions.contentmodel.settings.ContentModelSettings;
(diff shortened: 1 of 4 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/FilesUtils.java
```diff
@@ -262,6 +264,21 @@ public static Path getPath(String uri) {
...
 	}
 
+	/**
+	 * Returns true if the provided path exists.
+	 * Returns false otherwise.
+	 * 
+	 * @param path the path to check
+	 * @return true if the provided path exists and false otherwise
+	 */
+	public static boolean pathExists(String path) {
```
angelozerr, line 274: I wonder if you could use the existing FilesUtils.getPath to avoid creating a new utility method?

xorye: Fixed. I removed this method and now I am using `Files.notExists(FilesUtils.getPath(c));` and `Files.exists(FilesUtils.getPath(c));`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/FilesUtils.java
```diff
@@ -262,6 +262,21 @@ public static Path getPath(String uri) {
 		return Paths.get(uri);
 	}
 
+	// /**
```
angelozerr, line 265: Could you remove this comment please.

xorye: fixed

angelozerr:
Thanks @xorye !

PR #432, opened 2019-06-12
Preference 'xml.symbols.exclude' for an array of patterns to exclude symbols for file(s)

NikolasKomonen:
No tests yet

The structure of this PR.

So for the patterns it will take in an array of patterns as strings.
The server will convert those patterns to XMLExcludedSymbolFiles classes.

Each TextDocument has a doSymbols Boolean.
When symbols is requested on it it will check if that Boolean has been set yet, or else it will compute a pattern match.

Let me know if this works or could be improved.

PR #836, opened 2020-08-05
Type rename provider

xorye:
Server side part for https://github.com/redhat-developer/vscode-xml/pull/304


xorye:
Added tests

angelozerr:
Today this type rename provider became linked editing range with a LSP specification. This PR is replaced with #432

PR #837, opened 2020-08-07
Extension completion should use node text range

### LINKED ISSUE #723: completionRequest.getReplaceRange() is erroneous in text that contains /
In LemMinX extensions, `CompletionParticipantAdapter.onXMLContent`, when on a node that contains `/` (eg `path/to/something`), then `request.getReplaceRange()` returns only the offset location (resulting in an insertion) while in other cases, it returns the whole text node area (which allows filtering and replacement).
I would expect to always get the whole text node range.
If it's not possible to always have the whole text node range, then it should always be the offset; or whatever over sane convention as long _it's always the same one_.

AObuchow:
Fix #723

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
@@ -31,6 +31,7 @@ import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.dom.DOMText;
 import org.eclipse.lemminx.dom.DTDDeclParameter;
 import org.eclipse.lemminx.dom.parser.Scanner;
 import org.eclipse.lemminx.dom.parser.ScannerState;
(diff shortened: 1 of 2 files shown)


AObuchow:
This probably needs a test.

I can confirm however that it fixes the issue related to https://github.com/eclipse/lemminx-maven/pull/127 (I removed the workaround but kept the test from that PR)

angelozerr:
> This probably needs a test.

Yes please.

AObuchow:
@angelozerr should the test just extend `CompletionParticipantAdapter`, perform a completion and test that the completion replacement range is the whole node's text?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
```diff
@@ -753,6 +754,18 @@ private void collectInsideContent(CompletionRequest request, CompletionResponse
 			collectCloseTagSuggestions(tagNameRange, true, true, false, request, response);
 		}
 		// Participant completion on XML content
+		if (request.getNode() instanceof DOMElement) {
```
AObuchow, line 757: I think this section about DOMElement probably should be removed, I don't think it actually addresses the issue that was reported (even though it made downstream lemminx-maven workaround tests pass..). 

The original bug was about triggering a completion on text, not on a node?

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/HTMLCompletionExtensionsTest.java
```diff
@@ -118,6 +119,19 @@ public void testHTMLAttributeNameCompletion() throws BadLocationException {
...
+		Position start = new Position(0, 7);
+		Position end = new Position(0, 26);
+		TextEdit edit = new TextEdit(new Range(start, end), "replacement text");
+		completion.setTextEdit(edit);
+		testCompletionFor("<input>some extisti|ng text</input>", completion );
+	}
+	
+	@Test
+	private void testHTMLOnXMLContentCompletion() throws BadLocationException {
+		// TODO: This isin't getting called by Junit
```
AObuchow, line 134: I don't know why I couldn't get my test to be called :P Is there anything I'm missing to get it called? For now I put the test code in `testHTMLAttributeValueCompletion` so that it'd be called on CI

fbricon: needs to be public

angelozerr:
This PR is replaced with https://github.com/eclipse/lemminx/pull/866

PR #844, opened 2020-08-18
How to set catalog prefer as public in vscode setting?

angelozerr:
How to set catalog prefer as public in vscode setting?

See https://github.com/redhat-developer/vscode-xml/issues/295

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/ContentModelSettings.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/ContentModelSettings.java
@@ -24,6 +24,8 @@ public class ContentModelSettings {
 
 	private String[] catalogs = null;
 
+	private XMLCatalogSettings catalog;
+
 	private XMLFileAssociation[] fileAssociations;
 
 	private XMLValidationSettings validation;
@@ -54,6 +56,7 @@ public class ContentModelSettings {
 	 *
 	 * @param catalogs
 	 */
+	@Deprecated
 	public void setCatalogs(String[] catalogs) {
 		this.catalogs = catalogs;
 	}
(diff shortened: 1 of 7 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/ContentModelSettings.java
```diff
@@ -64,7 +67,13 @@ public void setCatalogs(String[] catalogs) {
...
 	 */
 	public String[] getCatalogs() {
-		return catalogs == null ? new String[0] : catalogs;
+		XMLCatalogSettings catalog = getCatalog();
+		if (catalog != null) {
+			if (catalog.getFiles() != null) {
+				return catalog.getFiles();
+			}
+		}
+		return catalogs;
```
fbricon, line 76: it used to never return null, you should keep that

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLCatalogSettings.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public class XMLCatalogSettings {
+
+	public static Prefer DEFAULT_PREFER_PUBLIC = Prefer.PUBLIC;
+
+	public static boolean DEFAULT_USE_PREFER_PUBLIC = true;
+
+	public static boolean DEFAULT_USE_LITERAL_SYSTEM_ID = true;
+
+	public static final XMLCatalogSettings DEFAULT_CATALOG = new XMLCatalogSettings();
```
fbricon, line 26: default catalog should be immutable

fbricon:
please use a proper commit message (explaining the change)

angelozerr:
I close this issue since original issue https://github.com/redhat-developer/vscode-xml/issues/295 doesn't require this new settings.

PR #855, opened 2020-08-25
Fix completion & hover based on XSD with elementFormDefault="unqualified"

angelozerr:
Fix completion & hover based on XSD with elementFormDefault="unqualified"

See https://github.com/redhat-developer/vscode-xml/issues/311

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/model/ContentModelManager.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/model/ContentModelManager.java
@@ -68,9 +68,9 @@ public class ContentModelManager {
 	/**
 	 * Returns the owner document of the declared element which matches the given
 	 * XML element and null otherwise.
-	 * 
+	 *
 	 * @param element the XML element
-	 * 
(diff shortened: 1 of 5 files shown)


angelozerr:
To test this PR:

 * open `foo-qualified.xml`  which is in this PR
 * open completion after bar, you should see item (the fix).

or try  the user usecase https://github.com/redhat-developer/vscode-xml/issues/311#issue-682116842

datho7561:
For the user use case, I tried deleting `<subtitle></subtitle>` element, and then opening completion. It is specified as min 0 max 1, but it isn't suggested. If I open completion after `<subtitle></subtitle>`, it suggests adding another subtitle element:
SubtitleDoesntAppearUntilAfter

angelozerr:
@datho7561 have you configured the XML catalog?

datho7561:
> @datho7561 have you configured the XML catalog?

Thanks oops.

datho7561:
I configured the catalog, and now it works for the root level, but not nested elements:

NestedStillDoesntWork

angelozerr:
> I configured the catalog, and now it works for the root level, but not nested elements:

Indeed there were a problem with XML catalog. Please retry it.

PR #859, opened 2020-08-27
Add AbstractClasspathResourceResolver

rzgry:
Adds an abstract class to make it easier to deploy and resolve a classpath resource. Ie. XSD schema
@angelozerr mentioned this might be useful

Can be used like this:

```
package io.openliberty.lemminx.liberty;

import io.openliberty.lemminx.liberty.util.LibertyUtils;

public class LibertyXSDURIResolver extends AbstractClasspathResourceResolver {

  private static final String XSD_RESOURCE_URL = "https://example.com/server.xsd";
  private static final String XSD_CLASSPATH_LOCATION = "/schema/server.xsd";

  LibertyXSDURIResolver() {
    super(XSD_RESOURCE_URL, XSD_CLASSPATH_LOCATION);
  }

  @Override
  boolean shouldResolveResource(String baseLocation, String publicId, String systemId) {
    return LibertyUtils.isServerXMLFile(baseLocation);
  }
}

```

Signed-off-by: Ryan Zegray <ryan.zegray@ibm.com>


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/AbstractClasspathResourceResolver.java
```diff
@@ -0,0 +1,69 @@
+package org.eclipse.lemminx.uriresolver;
```
datho7561, line 1: Please add a copyright header

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/AbstractClasspathResourceResolver.java
```diff
@@ -0,0 +1,80 @@
+/*******************************************************************************
+* Copyright (c) 2020 IBM Corporation and others.
+* All rights reserved. This program and the accompanying materials
```
fbricon, line 3: header needs to be updated to dual license

fbricon:
unit tests would be nice ;-)

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/AbstractClasspathResourceResolver.java
rzgry: Most files seem to only have the 1 license in the lemminx repo. Should I still update to dual license?

angelozerr:
@rzgry it should be nice if XSLURIResolverExtension will extend your AbstractClasspathResourceResolver . I'm not sure that it works since XSLURIResolverExtension manage several XSD and choose it according the version attribute.

rzgry:
I can take a look at adapting the AbstractClasspathResourceResolver to support managing multiple XSDs.

rzgry:
@angelozerr Updated to instead have an abstract method `public abstract ResourceToDeploy resourceToResolve(String baseLocation, String publicId, String systemId);` Where you return the resource you want to resolve/deploy given the parameters.  And updated the implementation of XSLURIResolverExtension.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/AbstractClasspathResourceResolver.java
fbricon: You're right, sorry my brains were still on lsp4mp mode.

mickaelistria:
As a consumer, I'd rather not have to extend 1 class to provide a schema to LemMinX. I would hope for this basic use-case to be something more built-in directly in LemMinX API (see suggestions in #1007). One way to achieve that would be to provide a concrete `ClasspathResourceResolver(URI schemaUri, String classpathResource)` which we'd just have to instantiate with eg `new ClasspathResourceResolver(URI.create("http://maven.apache.org/maven-v4_0_0.xsd"), "/xsd/maven-v4_0_0.xsd")`.
Could this PR help towards that goal?

PR #866, opened 2020-09-02
Extension completion should use node text range

### LINKED ISSUE #723: completionRequest.getReplaceRange() is erroneous in text that contains /

angelozerr:
Extension completion should use node text range

Fixes #723


angelozerr:
@mickaelistria this PR could interest you

PR #873, opened 2020-09-11
Redirect stdio streams

fbricon:
this change prevents client-server from breaking when something (like lemminx-maven) writes to sdtout

~~However, we lose, for now the server startup output. That still needs to be fixed.~~

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java
@@ -94,6 +94,12 @@ public class XMLLanguageServer
 
 	@Override
 	public CompletableFuture<InitializeResult> initialize(InitializeParams params) {
+		Object initOptions = InitializationOptionsSettings.getSettings(params);
+		Object xmlSettings = AllXMLSettings.getAllXMLSettings(initOptions);
+		XMLGeneralClientSettings settings = XMLGeneralClientSettings.getGeneralXMLSettings(xmlSettings);
+
+		LogHelper.initializeRootLogger(languageClient, settings == null? null : settings.getLogs());
+
 		LOGGER.info("Initializing XML Language server" + System.lineSeparator() + new ServerInfo().details());
 
 		this.parentProcessId = params.getProcessId();
@@ -108,7 +114,7 @@ public class XMLLanguageServer
 		xmlTextDocumentService.updateClientCapabilities(capabilityManager.getClientCapabilities().capabilities,
 				capabilityManager.getClientCapabilities().getExtendedCapabilities());
 
-		updateSettings(InitializationOptionsSettings.getSettings(params));
+		updateSettings(initOptions, false /* already configured logging*/ );
 
 		ServerCapabilities nonDynamicServerCapabilities = ServerCapabilitiesInitializer.getNonDynamicServerCapabilities(
 				capabilityManager.getClientCapabilities(), xmlTextDocumentService.isIncrementalSupport());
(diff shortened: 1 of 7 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/logs/LogHelper.java
```diff
@@ -43,27 +43,30 @@ public static void initializeRootLogger(LanguageClient newLanguageClient, LogsSe
 		logger.setUseParentHandlers(false);// Stops output to console
 		
 		// Configure logging LSP client handler
-		if (settings.getClient()) {
-			try {
-				logger.addHandler(LogHelper.getClientHandler(newLanguageClient));
-			} catch (Exception e) {
-				// TODO: handle exception
-			}
+		try {
```
angelozerr, line 46: I don't understand, if you do that, it means that `xml.logs.client: false` will not work?

fbricon: you're right, I'll fix it

angelozerr: Please  rename `LogsSettings#getClient()` to `LogsSettings#isClient()`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLServerLauncher.java
```diff
@@ -23,6 +26,8 @@
 
 import org.eclipse.lemminx.commons.ParentProcessWatcher;
 import org.eclipse.lemminx.customservice.XMLLanguageClientAPI;
+import org.eclipse.lemminx.io.NoOpInputStream;
```
angelozerr, line 29: NoOpInputStream, NoOpInputStream, NoOpPrintStream are just used for XMLServerLauncher. I think it should be better to avoid exposing those classes which are not an API for contributions.

That's why I suggest you that you define NoOpInputStream, NoOpInputStream, NoOpPrintStream as static private class inside XMLServerLauncher.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java
```diff
@@ -135,20 +141,23 @@ public void initialized(InitializedParams params) {
 	 * 
 	 * @param initializationOptionsSettings the XML settings
 	 */
-	public synchronized void updateSettings(Object initializationOptionsSettings) {
-		if (initializationOptionsSettings == null) {
+	public synchronized void updateSettings(Object initOptions) {
+		updateSettings(initOptions, true);
+	}
+
+	private synchronized void updateSettings(Object initOptions, boolean initLogs) {
```
angelozerr, line 148: Add comments to explain the initLogs parameter.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java
```diff
@@ -135,20 +141,23 @@ public void initialized(InitializedParams params) {
...
+		Object initSettings = AllXMLSettings.getAllXMLSettings(initOptions);
+		XMLGeneralClientSettings xmlClientSettings = XMLGeneralClientSettings.getGeneralXMLSettings(initSettings);
 		if (xmlClientSettings != null) {
-			// Update logs settings
-			LogsSettings logsSettings = xmlClientSettings.getLogs();
-			if (logsSettings != null) {
-				LogHelper.initializeRootLogger(languageClient, logsSettings);
+			if (initLogs) {
+				// Update logs settings	
+				LogHelper.initializeRootLogger(languageClient, xmlClientSettings.getLogs());
```
angelozerr, line 158: I think it worked like this before, but when user change `xml.logs.client` to false, it doesn't remove the LSP handler. In other words the user need to restart vscode.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLServerLauncher.java
fbricon: I'd rather keep them as separate classes. Keeping them public is no big deal, but if you prefer, I can move them to the lemminx package and keep them package visible only

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java
fbricon: no https://github.com/eclipse/lemminx/pull/873/files#diff-8e86b1bafd539aecf46fde1e4f1cdcbaR43 unregisters all Log handlers, every time we change settings

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLServerLauncher.java
angelozerr: > I can move them to the lemminx package and keep them package visible only

Yes please. I would like to avoid that contributor uses those classes.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java
angelozerr: Oops sorry, I have not seen this piece of code. Thanks for the clarification.

Issue #878, opened 2020-09-16
Code Action to close root element closing tag inserts wrong closing tag


datho7561:
Given this XML:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<a attr=""></a
```

There is a code action to close `<a>` with `</a>`. However, the result is:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<a attr=""></a></a
```

angelozerr:
Fixed with https://github.com/eclipse/lemminx/pull/897

PR #881, opened 2020-09-21
Improve ETagRequired error range

### LINKED ISSUE #876: Improve ETagRequired error range
Given this XML file:

```xml
<foo>
  <bar></
</foo>
```

We have that:

image

Although the ETagRequired error range should highlight the start tag element:

angelozerr:
Improve ETagRequired error range

Fixes #876

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
@@ -333,6 +333,44 @@ public class DOMElement extends DOMNode implements org.w3c.dom.Element {
 		return startTagCloseOffset;
 	}
 
+	public int getBestStartTagCloseOffset() {
+		if (isStartTagClosed()) {
+			return getStartTagCloseOffset();
+		}
+		if (hasEndTag()) {
+			// ex : <foo           </foo>
+			if (hasAttributes()) {
+				// ex : <foo attr=""          </foo>
+				// returns end offset of last attributes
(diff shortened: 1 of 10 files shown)


angelozerr:
My code requires to be cleaned and I need to add some tests again, but I think we can start to play with it.

Basicly, this PR improves error range for diagnostic. For instance given this XML 

```xml
<abc>
  <def/>
```

Before `def `was highlighted which is wrong. Today it highlights `abc`. When I fixed that I need to fix code action to close abc. There were a lot of bugs and I fixed (by adding new tests) and I support the eat of characters (ex : it fixes the issue https://github.com/eclipse/lemminx/issues/878)

To test correctly this PR, I suggest you do step by step. Ex : 

write 

```xml
<a
```

see if the error range is correct and code action too.

write 

```xml
<a>
```

write 

```xml
<a><b
```

write 

```xml
<a><b>
```

etc...

datho7561:
```xml
<foo>
  <bar><bar></</bar>
</foo>
```
Apply quickfix -->
```xml
<foo>
  <bar><bar</bar></bar>
</foo>
```
Apply quickfix -->
```xml
<foo>
  <bar><bar</bar>></bar>
</foo>
```
This behaviour is slightly different from what happens on master (on master the quickfixes don't help either), but this should probably be fixed in a different PR.

datho7561:
```xml
<foo>
  <bar><bar></bar>
  </bar></</bar>
</foo>
```
Apply quickfix -->
```xml
<foo</foo></foo>
```

angelozerr:
Abandoned for https://github.com/eclipse/lemminx/pull/897

PR #883, opened 2020-09-22
Use CompletableFuture in CacheResourceManagerTest 

### LINKED ISSUE #753: CacheResourcesManagerTest.testAvailableCache fails sometimes
The Junit test CacheResourcesManagerTest.testAvailableCache fails sometimes in CI build. I suspect that several PR uses the same folder to check file is not downloaded. I think teh fix is to use a temporar file to download XSD, DTD files to test.

datho7561:
Use a CompletableFuture to wait for the download to finish when testing the CacheResourceManager. This should mean different CPU/network conditions don't cause the tests to fail. A generous timeout is given for the downloads.

Fixes #753

### DIFF
--- a/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/DTDHoverExtensionsTest.java
+++ b/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/DTDHoverExtensionsTest.java
@@ -15,6 +15,7 @@ package org.eclipse.lemminx.extensions.contentmodel;
 
 import static org.eclipse.lemminx.XMLAssert.r;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -75,13 +76,14 @@ public class DTDHoverExtensionsTest {
 			Path cachedFilePath = CacheResourcesManager.getResourceCachePath(httpDTDUri);
 			Files.deleteIfExists(cachedFilePath);
 
-			// Download the DTD by waiting 1 sec
 			CacheResourcesManager cacheResourcesManager = new CacheResourcesManager();
 			try {
 				cacheResourcesManager.getResource(httpDTDUri);
-			} catch (CacheResourceDownloadingException ignored) {
+				fail("Expected file to be downloading");
+			} catch (CacheResourceDownloadingException containsFuture) {
+				// Wait for download to finish
+				containsFuture.getFuture().get(30, TimeUnit.SECONDS);
 			}
-			TimeUnit.MILLISECONDS.sleep(1000);
 			assertTrue(Files.exists(cachedFilePath),
 					"'" + cachedFilePath + "' file should be downloaded in the cache.");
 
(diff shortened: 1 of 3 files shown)


datho7561:
Found some other tests that fail when CPU usage is limited, so I am going to try to fix those as well.

datho7561:
I tested this PR by limiting CPU usage to 30% and bandwidth to 100kbps. Tests consistently pass. Settings lower than that caused tests to fail.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaPublishDiagnosticsTest.java
```diff
@@ -15,6 +15,7 @@
 import static org.eclipse.lemminx.XMLAssert.pd;
 import static org.eclipse.lemminx.XMLAssert.r;
 
+import java.util.concurrent.TimeUnit;
```
datho7561, line 18: Oops I'll take this out

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/uriresolver/CacheResourcesManagerTest.java
```diff
@@ -71,11 +71,20 @@ public void testUnavailableCache() throws Exception {
 			fail("cacheResourcesManager should be busy downloading the url");
 		} catch (CacheResourceDownloadingException ignored) {
 		}
-		TimeUnit.MILLISECONDS.sleep(200);
-		// failed to download so returns null
-		assertNull(cacheResourcesManager.getResource(uri));
 
-		TimeUnit.SECONDS.sleep(1);// wait past the cache expiration date
+		boolean downloaded = false;
+		int i = 0;
```
datho7561, line 76: No need to make this a fixed number of attempts

datho7561:
In the latest commit:

Use the CompletableFuture that can be accessed in the exception in order
to wait for the download to finish when testing the CacheResourceManager.
This should mean different CPU/network conditions don't cause the tests to fail.
It also avoids regular checking that the previous PR implemented.

Please let me know if you think this is a good idea.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/DTDHoverExtensionsTest.java
```diff
@@ -79,26 +79,30 @@ public void webXML() throws Exception {
 			CacheResourcesManager cacheResourcesManager = new CacheResourcesManager();
 			try {
 				cacheResourcesManager.getResource(httpDTDUri);
-			} catch (CacheResourceDownloadingException ignored) {
-			}
-			TimeUnit.MILLISECONDS.sleep(1000);
-			assertTrue(Files.exists(cachedFilePath),
-					"'" + cachedFilePath + "' file should be downloaded in the cache.");
+			} catch (CacheResourceDownloadingException containsFuture) {
+				containsFuture.getFuture().thenAccept(path -> {
```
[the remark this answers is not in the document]
datho7561: You are right, it isn't executed. I will rework this.

PR #891, opened 2020-09-25
Fix infinite loop inside LSPMessageFormatter for some cases

### LINKED ISSUE #856: Infinite loop inside LSPMessageFormatter for some cases
In some case https://github.com/eclipse/lemminx/blob/875d5b427f9db918857df9db2b0939282b048a12/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xerces/LSPMessageFormatter.java#L153 don't exit from the loop.

Given this XSD:

```xml
<xs:schema
	targetNamespace="http://foo">
	<xs:element name="foo">
		<xs:complexType>
			<xs:sequence>
				<xs:element name="bar" />
			</xs:sequence>
		</xs:complexType>
	</xs:element>
</xs:schema>
```

and this XML file:

```xml
<foo xmlns="http://foo" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://foo foo.xsd">
	<bar />
</foo>
```

error are crazy:

image

angelozerr:
Fix infinite loop inside LSPMessageFormatter for some cases

Fixes #856

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xerces/LSPMessageFormatter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xerces/LSPMessageFormatter.java
@@ -137,29 +137,28 @@ public class LSPMessageFormatter implements MessageFormatter {
 		}
 	}
 
-	private static String reformatElementNames(boolean hasNamespace, String names) {
-		StringBuilder sb = new StringBuilder();
+	private static String reformatElementNames(String names) {
+		StringBuilder result = new StringBuilder();
 
 		MultiLineStream stream = new MultiLineStream(names, 0);
 
 		while (!stream.eos()) { // }
 			stream.advance(1);// Consume ' ' or '{' if first item
-			if(hasNamespace) {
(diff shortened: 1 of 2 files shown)


angelozerr:
This PR fixes #856, so now you have correct error range and server doesn't stop when you try to validate the foo.xml file too many times (because of infinite loop).

But code action is wrong:

image

After spending time, the correct fix is to add `elementFormDefault="qualified"` in the XSD file:

```xml
<xs:schema
	targetNamespace="http://foo">
	<xs:element name="foo">
		<xs:complexType>
			<xs:sequence>
				<xs:element name="bar" />
			</xs:sequence>
		</xs:complexType>
	</xs:element>
</xs:schema>
```

I think it's an hard task.

datho7561:
> After spending time, the correct fix is to add `elementFormDefault="qualified"` in the XSD file:

I think it makes sense to put this in a separate PR. Could you please open a separate issue for this?

datho7561 (approved):

angelozerr:
> I think it makes sense to put this in a separate PR. Could you please open a separate issue for this?

yes sure https://github.com/eclipse/lemminx/issues/894

PR #896, opened 2020-09-29
XMLExtensionRegistry: capture initial settings if it's absent

yatli:
See: https://github.com/fannheyward/coc-xml/issues/9

Symptom is, `xml.fileAssociations` doesn't work with `coc.nvim`.

Pasting my comment in that issue:

------------------

Observations from my debug run:

The file associations successfully made it into `XMLLanguageServer.updateSettings`. Around line 184,
```java
		ContentModelSettings cmSettings = ContentModelSettings
				.getContentModelXMLSettings(initializationOptionsSettings);
```
I can see the file association rules in cmSettings. However, it's not used to populate the `XMLFileAssociationResolverExtension` settings.

The only path to `XMLFileAssociationResolverExtension.setFileAssociations` is through `XMLTextDocumentService.doSave` ->  `ContentModelPlugin.doSave` -> `ContentModelManager.setFileAssociations`:

```java
	public void doSave(ISaveContext saveContext) {
		if (initialized) {
			extensions.stream().forEach(extension -> extension.doSave(saveContext));
		} else {
			this.initialSaveContext = saveContext;
		}
	}
```

```java
		Object initializationOptionsSettings = saveContext.getSettings();
		cmSettings = ContentModelSettings.getContentModelXMLSettings(initializationOptionsSettings);
```

unfortunately, `coc.nvim` emits `didSave` event too fast, and `XMLTextDocumentService.doSave` is called twice before the extensions are registered... And the actual initial save context is lost.

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/XMLExtensionsRegistry.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/XMLExtensionsRegistry.java
@@ -27,6 +27,7 @@ import org.eclipse.lemminx.services.extensions.codelens.ICodeLensParticipant;
 import org.eclipse.lemminx.services.extensions.diagnostics.IDiagnosticsParticipant;
 import org.eclipse.lemminx.services.extensions.format.IFormatterParticipant;
 import org.eclipse.lemminx.services.extensions.save.ISaveContext;
+import org.eclipse.lemminx.services.extensions.save.ISaveContext.SaveContextType;
 import org.eclipse.lemminx.uriresolver.URIResolverExtensionManager;
 import org.eclipse.lsp4j.InitializeParams;
 
@@ -110,7 +111,10 @@ public class XMLExtensionsRegistry implements IComponentProvider {
 	public void doSave(ISaveContext saveContext) {
 		if (initialized) {
 			extensions.stream().forEach(extension -> extension.doSave(saveContext));
-		} else {
+		} else if (this.initialSaveContext == null || (saveContext != null && saveContext.getType() == SaveContextType.SETTINGS)) {
+			// capture initial configuration iff:
+			// 1. the saveContext is for configuration, not document save
+			// 2. we haven't captured settings before
 			this.initialSaveContext = saveContext;
 		}
 	}
--- a/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/ExtensionRegistryDoSaveTest.java
+++ b/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/ExtensionRegistryDoSaveTest.java
@@ -1,0 +1,101 @@
+/**
+ *  Copyright (c) 2020 Yatao Li.
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v2.0
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *  Yatao Li <yatao.li@live.com> - initial API and implementation
+ */
+package org.eclipse.lemminx.services.extensions;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.function.Predicate;
+
+import org.eclipse.lemminx.XMLAssert.SettingsSaveContext;
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.services.extensions.IXMLExtension;
+import org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry;
+import org.eclipse.lemminx.services.extensions.save.ISaveContext;
+import org.eclipse.lsp4j.InitializeParams;
+import org.junit.jupiter.api.Test;
+
+import com.google.gson.JsonObject;
+
+/**
+ * Ensures that {@code XMLExtensionsRegistry.doSave()} correctly captures the
+ * initial LS configuration, and the configuration is correctly forwarded to
+ * {@code IXMLExtension.doSave()} for registered extensions.
+ */
+public class ExtensionRegistryDoSaveTest {
+	
+	class RegistryTestExtension implements IXMLExtension {
+		
+		private ISaveContext context = null;
+		
+		@Override
+		public void start(InitializeParams params, XMLExtensionsRegistry registry) {
+		}
+
+		@Override
+		public void stop(XMLExtensionsRegistry registry) {
+		}
+		
+		@Override
+		public void doSave(ISaveContext context) {
+			this.context = context;
+		}
+
+		public ISaveContext getContext() {
+			return context;
+		}
+	}
+	
+	private static class SaveFileContext implements ISaveContext {
+		@Override
+		public DOMDocument getDocument(String uri) {
+			return null;
+		}
+
+		@Override
+		public void collectDocumentToValidate(Predicate<DOMDocument> validateDocumentPredicate) {
+		}
+
+		@Override
+		public SaveContextType getType() {
+			return SaveContextType.DOCUMENT;
+		}
+
+		@Override
+		public String getUri() {
+			return null;
+		}
+
+		@Override
+		public Object getSettings() {
+			return null;
+		}
+	}
+
+	@Test
+	public void initialConfiguration() {
+		XMLExtensionsRegistry registry = new XMLExtensionsRegistry();
+		ISaveContext settingsSaveContext = new SettingsSaveContext(new JsonObject());
+		// initial config passed in during server startup.
+		registry.doSave(settingsSaveContext);
+		// then another event comes in from the client
+		registry.doSave(new SaveFileContext());
+		// extension registration is triggered
+		registry.initializeIfNeeded();
+		RegistryTestExtension extension = new RegistryTestExtension();
+		registry.registerExtension(extension);
+		// examine that the extension properly receives the initial config
+		assertEquals(settingsSaveContext, extension.getContext());
+	}
+}
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/generators/FileContentGeneratorPlugin.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/generators/FileContentGeneratorPlugin.java
@@ -25,9 +25,11 @@ public class FileContentGeneratorPlugin implements IXMLExtension {
 
 	@Override
 	public void start(InitializeParams params, XMLExtensionsRegistry registry) {
-		IXMLFullFormatter formatter = (IXMLFullFormatter) registry;
-		FileContentGeneratorManager manager = new FileContentGeneratorManager(formatter);
-		registry.registerComponent(manager);
+		if (registry instanceof IXMLFullFormatter) {
+			IXMLFullFormatter formatter = (IXMLFullFormatter) registry;
+			FileContentGeneratorManager manager = new FileContentGeneratorManager(formatter);
+			registry.registerComponent(manager);
+		}
 	}
 
 	@Override


angelozerr:
> unfortunately, coc.nvim emits didSave event too fast,

Do you think it could be fixed ?

Your PR seems an hack but I'm not against to merge it if you can fix the problem on your side.

yatli:
It's not a hack, it's a bug fix.
The LSP does not restrict a client from sending `didSave` after the server is started.
The problem is that the extensions are loaded too late, and multiple things can trigger `doSave` which results in null initialContext.
Fixing one client does not solve the problem.

angelozerr:
> It's not a hack, it's a bug fix.

Ok thanks for your clarification. Do you think you could create a JUnit test for this usecase? 

Could add a comment which explains why you do that. Thanks.

yatli:
Yeah sure. I think I figure out how to do that :)

yatli:
One thing I'm not sure, about the initial settings capture logic.
Currently this PR is conservative -- just takes the first non-null incoming initial context. All later ones will be ignored.
There could be corner cases where the client sends multiple config messages after server start.
I'm not sure _when_ will the extension registration happen, so eager context capture would result in extensions holding different settings -- so I'm not going with that logic now.

angelozerr:
> Currently this PR is conservative -- just takes the first non-null incoming initial context. All later ones will be ignored.

I wonder if we should store those initial save context in an array list?

> I'm not sure when will the extension registration happen, so eager context capture would result in extensions holding different settings -- so I'm not going with that logic now.

It happens when the first participant (ex : completion) is called.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/XMLExtensionsRegistry.java
```diff
@@ -110,7 +110,10 @@ public void initializeParams(InitializeParams params) {
 	public void doSave(ISaveContext saveContext) {
 		if (initialized) {
 			extensions.stream().forEach(extension -> extension.doSave(saveContext));
-		} else {
+		} else if (this.initialSaveContext == null || (saveContext != null && saveContext.getSettings() != null)) {
```
angelozerr, line 113: I think you should write the condition like this:

`} else if (this.initialSaveContext == null || (saveContext != null && saveContext.getType() == SaveContextType.SETTINGS)) {`

angelozerr:
@yatli is there any chance to have JUnit test please?

yatli:
Yep, I'll do it when I get back to the dev box.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/XMLExtensionsRegistry.java
yatli: Got it.

datho7561:
I tested this PR with coc-xml and confirms it works. Once there is a unit test, I think it will be good to merge.

This might be helpful for writing the unit tests: https://github.com/eclipse/lemminx/commit/7fe05f1319bb3b2ba5893c112e93bc66587e1b87#diff-0043a4b8409438a109b6b8c04c1c178fR34

angelozerr:
@yatli we plan to do a release soon, is there any chance that you could provide tests? Thanks!

yatli:
@angelozerr I've set up the test in https://github.com/eclipse/lemminx/pull/896/commits/6752e98bd2747f453c9ac2a6db96cd40d2a6f2fd but it didn't work because my IXMLExtension did not load and the extensions list is private. I'm not a Java dev so my knowledge with this is very limited. Could you help to determine the best way to implement the best logic?

datho7561:
@yatli  Here's what worked for me to get the extension to load:
1. Move RegistryTestExtension into a separate `.java` file under the tests directory
2. Make a separate testing service registry: https://stackoverflow.com/questions/29204840/unit-test-serviceloader, i.e. copy

`org.eclipse.lemminx/src/main/resources/META-INF/services/org.eclipse.lemminx.services.extensions.IXMLExtension`

to

`org.eclipse.lemminx/src/test/resources/META-INF/services/org.eclipse.lemminx.services.extensions.IXMLExtension`

3. Add RegistryTestExtension to the testing service registry (add a new line with the full name of the RegistryTestExtension class into the copy of the file)

angelozerr:
> Make a separate testing service registry

You can avoid doing that. The idea is to call registerExtension. Please see [ExtensionRegistryDisposeTest](https://github.com/eclipse/lemminx/blob/9f5f757f2e173883ca4e583099dc7a19e547c617/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/ExtensionRegistryDisposeTest.java#L34)

yatli:
I put the test next to the suggested one but the registerExtension method is still not visible to me.

What magic is this 🤔

Screenshot_20201010-112418

angelozerr:
> I put the test next to the suggested one but the registerExtension method is still not visible to me.

> What magic is this 

It's because registerExtension has not modifier (private, public, protected before registerExtension), that's why only classes which are in the same package than XMLExtensionregistry can use it.

If you set your test class in the same package (in src/test/java) than XMLExtensionregistry it should work, but if you want to set your test class in another package, I suggest that you add `protected `before the `registerExtension` method.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/XMLExtensionRegistryTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+		@Override
+		public Object getSettings() {
+			return null;
+		}
+	}
+
+	@Test
+	public void initialConfiguration() {
+		XMLLanguageService registry = new XMLLanguageService();
```
angelozerr, line 71: To fix compilation problem, replace with

`		XMLExtensionsRegistry registry = new XMLExtensionsRegistry();`

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/XMLExtensionRegistryTest.java
```diff
@@ -0,0 +1,83 @@
+package org.eclipse.lemminx.services.extensions;
```
angelozerr, line 1: Add EPL 2.0 license (see other class with sample)

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/XMLExtensionRegistryTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.eclipse.lemminx.services.XMLLanguageService;
+import org.eclipse.lemminx.services.extensions.IXMLExtension;
+import org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry;
+import org.eclipse.lemminx.services.extensions.save.ISaveContext;
+import org.eclipse.lsp4j.InitializeParams;
+import org.junit.jupiter.api.Test;
+
+import com.google.gson.JsonObject;
+
+public class XMLExtensionRegistryTest {
```
angelozerr, line 19: To be consistent with existing test ExtensionRegistryDisposeTest, I suggest to rename this class with ExtensionRegistryDoSaveTest

angelozerr, line 19: Add javadoc which explain the test

yatli:
Thanks @angelozerr , I've cleaned it up a little bit.

wrt `XMLExtensionsRegistry` vs. `XMLLanguageService`, I chose the later because the registry does not properly initialize:

```
java.lang.ClassCastException: class org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry cannot be cast to class org.eclipse.lemminx.services.IXMLFullFormatter (org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry and org.eclipse.lemminx.services.IXMLFullFormatter are in unnamed module of loader 'app')
	at org.eclipse.lemminx.extensions.generators.FileContentGeneratorPlugin.start(FileContentGeneratorPlugin.java:28)
	at org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry.registerExtension(XMLExtensionsRegistry.java:210)
	at org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry.lambda$2(XMLExtensionsRegistry.java:202)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry.initialize(XMLExtensionsRegistry.java:201)
	at org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry.initializeIfNeeded(XMLExtensionsRegistry.java:191)
	at org.eclipse.lemminx.services.extensions.ExtensionRegistryDoSaveTest.initialConfiguration(ExtensionRegistryDoSaveTest.java:92)
```

... but it looks like this is a false alarm, as the test completes successfully.

angelozerr:
> I chose the later because the registry does not properly initialize:

Please update `FileContentGeneratorPlugin#start` like this:

```
	public void start(InitializeParams params, XMLExtensionsRegistry registry) {
		if (registry instanceof IXMLFullFormatter) {
			IXMLFullFormatter formatter = (IXMLFullFormatter) registry;
			FileContentGeneratorManager manager = new FileContentGeneratorManager(formatter);
			registry.registerComponent(manager);
		}
	}
```

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/ExtensionRegistryDoSaveTest.java
```diff
@@ -0,0 +1,98 @@
...
+		// initial config passed in during server startup.
+		registry.doSave(new SettingsSaveContext(new JsonObject()));
+		// then another event comes in from the client
+		registry.doSave(new SaveFileContext());
+		// extension registration is triggered
+		registry.initializeIfNeeded();
+		RegistryTestExtension extension = new RegistryTestExtension();
+		registry.registerExtension(extension);
+		// examine that the extension properly receives the initial config
+		assertTrue(extension.isExtensionDoSaveCalled());
```
angelozerr, line 96: I think the test should check that the ISaveContext which is stored is the save context coming from SETTINGS (and not from DOCUMENT which is done before settings).

yatli: When the context is DOCUMENT it won't make it into the extension DoSave.

angelozerr: I think the correct test is:

```java
package org.eclipse.lemminx.services.extensions;

import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertNotNull;

import java.util.function.Predicate;

import org.eclipse.lemminx.XMLAssert.SettingsSaveContext;
import org.eclipse.lemminx.dom.DOMDocument;
import org.eclipse.lemminx.services.extensions.save.ISaveContext;
import org.eclipse.lsp4j.InitializeParams;
import org.junit.jupiter.api.Test;

import com.google.gson.JsonObject;

public class XMLExtensionRegistryTest {

	class RegistryTestExtension implements IXMLExtension {

		private ISaveContext context;

		@Override
		public void start(InitializeParams params, XMLExtensionsRegistry registry) {
		}

		@Override
		public void stop(XMLExtensionsRegistry registry) {
		}

		@Override
		public void doSave(ISaveContext context) {
			this.context = context;
		}

		public ISaveContext getContext() {
			return context;
		}
	}

	private static class SaveFileContext implements ISaveContext {
		@Override
		public DOMDocument getDocument(String uri) {
			return null;
		}

		@Override
		public void collectDocumentToValidate(Predicate<DOMDocument> validateDocumentPredicate) {
		}

		@Override
		public SaveContextType getType() {
			return SaveContextType.DOCUMENT;
		}

		@Override
		public String getUri() {
			return null;
		}

		@Override
		public Object getSettings() {
			return null;
		}
	}

	@Test
	public void initialConfiguration() {
		XMLExtensionsRegistry registry = new XMLExtensionsRegistry();
		// initial config passed in during server startup.
		ISaveContext settingsSaveContext  = new SettingsSaveContext(new JsonObject());
		registry.doSave(settingsSaveContext );
		// then another event comes in from the client
		ISaveContext documentSaveContext  = new SaveFileContext();
		registry.doSave(documentSaveContext);
		// extension registration is triggered
		registry.initializeIfNeeded();
		RegistryTestExtension extension = new RegistryTestExtension();
		registry.registerExtension(extension);
		// examine that the extension properly receives the initial config
		assertNotNull(extension.getContext());
		assertEquals(settingsSaveContext.getType(), extension.getContext().getType());
	}
}
```

what do you think about that?

yatli: Just got back home, no need to RDP into the dev box now :)
Yes I think this looks better.
In af3dfac I updated the test logic.

angelozerr:
@yatli could you squash your 6 commits in one commit please.

yatli:
Done.

angelozerr:
@yatli thanks so much for your contribution and your patience. As you explained us, Java is not your language but you did a great PR. Thanks again!

yatli:
🎉 thanks @angelozerr @datho7561 for the tips and review!
This will enable us to use vim (`coc-xml`) and other LSP friends to edit msbuild, XAML, svg and a lot more.

angelozerr:
> thanks @angelozerr @datho7561 for the tips and review!

You are welcome!

> This will enable us to use vim (coc-xml) and other LSP friends to edit msbuild, XAML, svg and a lot more.

It's a great news for LemMinx!

angelozerr:
For XAML, I wonder if it should be nice to develop a LemMinx extension to manage xClass attribute and binding between XAML file and CLR:

Given this XML fle:

```xaml
<Window x:Class="DemoApp.Demos.Blog.SimpleWindow"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
    <Grid>
        <Button />
    </Grid>
</Window>
```

You could manage with a XAMLPlugin custom completion, hyperlink, validation etc for x:Class attribute. What do you think about that?

yatli:
hmm very interesting. I believe this will bring in a .NET runtime though -- ideally coreclr.
The Avalonia (a cross platform XAML UI framework) team has done something similar to power their VS extension.

and probably integration into the build system, to listen to assembly update events...

angelozerr:
If you think it could be interesting to support advanced support for XAML like I I suggest you,  I could give you some help to implement that. The current PR https://github.com/eclipse/lemminx/pull/905 will open the door to manage custom service with command.

yatli:
Thanks. I'll see if I can find a time slot to focus on this topic.

angelozerr:
See https://github.com/eclipse/lemminx/blob/master/docs/LemMinX-Extensions.md#xml-language-server-services-available-for-extensions for more documentation about XML custom commands.

yatli:
Any words for the release? ;)

angelozerr:
We should release in the next weeks.

yatli:
Thanks for the update!

PR #899, opened 2020-10-01
Completion and goto definition for XSL Stylesheets

Woocash9:
This PR adds support for completion and goto definition actions for the following objects in XSL stylesheet:

completion:
- name attribute of call-template element
- name attribute of with-param element
- mode attribute of apply-templates element
- variable or param in the scope inside any attribute

goto definition:
- named template - triggered from within call-template element or one of its children
- regular template - triggered from mode attribute of apply templates element
- included stylesheet - triggered from include or import element
- variable or param - triggered from attribute value when on variable reference

I know that generally extensions should be created as external plugins. However this PR is about general support for XSL not for specific XML document therefore in my opinion it would be better to keep it as internal extension.

Disclaimer
This is my first open source PR ever so please bear with me.

Thanks
Woocash


fbricon:
Hi @Woocash9, thanks for your interest and your contribution! 
I'll let @angelozerr review and discuss the merits of the code, and I'll simply focus on the contribution process.

FYI, the LemMinX project is licensed under the EPL-2 which is totally incompatible with the GPL. Unless you relicense your code to EPL-2, we won't be able to touch it.

If you're ok to change the license, before your contribution can be accepted by the project, you need to create and electronically sign an [Eclipse Contributor Agreement (ECA)](http://www.eclipse.org/legal/ecafaq.php):

1. Log in to the [Eclipse foundation website](https://accounts.eclipse.org/user/login/). You will need to
   create an account with the Eclipse Foundation if you have not already done so.
2. Click on "Eclipse ECA", and complete the form.

Be sure to use the same email address in your Eclipse account that you intend to use when you commit to GitHub.
All committers all commits are bound to the [Developer Certificate of Origin.](https://www.eclipse.org/legal/DCO.php)
As such, all parties involved in a contribution must have valid ECAs and commits must include [valid "Signed-off-by" entries.](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git)
Commits can be signed off by including the `-s` attribute in your commit message, for example, `git commit -s -m 'Interesting Commit Message.`

Because this is a rather large PR (> 1000 lines of code) we'll need.to open a [contribution questionnaire](https://www.eclipse.org/projects/handbook/#ip-cq). Nothing for you to do, but the IP validation process might delay merging the PR by a few days.

Woocash9:
Hi @fbricon, Thanks for the feedback. Will take a look at ECA over the weekend.

angelozerr:
@Woocash9 thanks so much for your contribution!

I have seen quickly your PR and we did that for XSD xs:element but we provide too references, codelens, highlight, rename, etc support.

IMHO to be consistent with XSD, you should follow the same idea than XSD extension https://github.com/eclipse/lemminx/tree/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/participants

The basic idea is not to wrap XMLDocument with XSLDocument, but provide a basic search utilities:

 * search target attributes for supporting completion, definition and highlight. See https://github.com/eclipse/lemminx/blob/7fe05f1319bb3b2ba5893c112e93bc66587e1b87/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/utils/XSDUtils.java#L136

 * search origin attributes for supporting codelens, reference, rename and highlight. See https://github.com/eclipse/lemminx/blob/7fe05f1319bb3b2ba5893c112e93bc66587e1b87/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/utils/XSDUtils.java#L255

Once you have implemented this utilities you can use it in a lot of features like completion, definition, references, codelens and the code is very simply.

For instance see

 * XSD completion -> https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/participants/XSDCompletionParticipant.java
 * XSD definition -> https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/participants/XSDDefinitionParticipant.java

It should be fantastic if you could adapt your PR by following this idea (we did that too for DTD with DTDUtils).

Another comments are:

 * please don't use `.*`  in your import packages.
 * please write tests for each support like we have written for XSD https://github.com/eclipse/lemminx/tree/master/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/xsd

If you think to wrap XMLDocument with XSLDocument is a better solution please give us your explanation, I'm not against wrapping XMLDocument but please explain me why you wish to do that. 

Please note that it exists several issues for your features https://github.com/eclipse/lemminx/labels/xsl%20extension

So I suggest you you create a PR per features (don't mix completion and definition in the same PR) and when you create a PR, please add in your comments 

```
Fixes #479

Signed-off-by: ...
```

Hope my comment will not discourage you and I'm very enthousiasm with your features.

angelozerr:
@Woocash9 is there any chance that you work on this issue again?

angelozerr:
@Woocash9 are you interested again to work on this issue?

PR #903, opened 2020-10-07
Indentation settings for CodeAction

datho7561:
When it recieves `textDocument/codeAction`, the server updates the indentation settings with the indentation settings of the current document. It does this by requesting the settings `xml.format.tabSize` and `xml.format.insertSpaces` from the client. It does this right before generating the CodeActions for the document.

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
@@ -28,6 +28,8 @@ import java.util.function.BiFunction;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import com.google.gson.JsonPrimitive;
+
 import org.eclipse.lemminx.client.ExtendedClientCapabilities;
 import org.eclipse.lemminx.client.LimitExceededWarner;
 import org.eclipse.lemminx.client.LimitFeature;
@@ -59,6 +61,8 @@ import org.eclipse.lsp4j.Command;
 import org.eclipse.lsp4j.CompletionItem;
 import org.eclipse.lsp4j.CompletionList;
 import org.eclipse.lsp4j.CompletionParams;
+import org.eclipse.lsp4j.ConfigurationItem;
+import org.eclipse.lsp4j.ConfigurationParams;
 import org.eclipse.lsp4j.DefinitionParams;
 import org.eclipse.lsp4j.DidChangeTextDocumentParams;
 import org.eclipse.lsp4j.DidCloseTextDocumentParams;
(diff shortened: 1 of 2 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
```diff
@@ -385,6 +412,9 @@ public void didClose(DidCloseTextDocumentParams params) {
 	public CompletableFuture<List<Either<Command, CodeAction>>> codeAction(CodeActionParams params) {
 		return computeDOMAsync(params.getTextDocument(), (cancelChecker, xmlDocument) -> {
 			String uri = params.getTextDocument().getUri();
+
+			sharedSettings.getFormattingSettings().merge(getIndentationSettings(uri).join());
```
angelozerr, line 416: Using _join_ can be dangerous (it could block the process), I think it should be better to use `thenApply`

datho7561: Okay. Would it be a good idea to do a `CompletableFuture.allOf(...)` here for the DOMAsync and the settings request, or should I do these as a chain of `.thenApply(...)`? I am fairly new to asynchronous programming.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
```diff
@@ -383,9 +409,14 @@ public void didClose(DidCloseTextDocumentParams params) {
...
-		return computeDOMAsync(params.getTextDocument(), (cancelChecker, xmlDocument) -> {
-			String uri = params.getTextDocument().getUri();
-			return getXMLLanguageService()
+		String uri = params.getTextDocument().getUri();
+		return getIndentationSettings(uri).thenAccept((indentationSettings) -> {
+			sharedSettings.getFormattingSettings().merge(indentationSettings);
+		})
+		.thenCombine(computeDOMAsync(params.getTextDocument(), (cancelChecker, xmlDocument) -> {
+			return xmlDocument;
+		}), (void_, xmlDocument) -> {
```
datho7561, line 418: Is there a naming conversion for `Void` objects?

datho7561:
Working on a fix so that if the client doesn't support this feature, it falls back to the default formatting settings properly.

rgrunber (commented):
Overall looks good and respects indentation/size settings when combined with https://github.com/redhat-developer/vscode-xml/pull/337 .

Should references to 'xml.format.useSpaces' and 'xml.format.indentSize' in comments and the commit log be updated to be 'xml.format.insertSpaces' and 'xml.format.tabSize' ?

fbricon:
Yeah let's keep it consistent

datho7561:
> Should references to 'xml.format.useSpaces' and 'xml.format.indentSize' in comments and the commit log be updated to be 'xml.format.insertSpaces' and 'xml.format.tabSize' ?

Good catch! Thanks!

rgrunber:
Change looks good to me so feel free to merge.

PR #905, opened 2020-10-08
Support for 'xml/executeClientCommand` access to server from extension

BoykoAlex:
Spring XML Extension needs two things from the XML server:
- Support for `xml/executeClientCommand` to execute a command on the client. The client command execution would first try to find a command registered by an LS if not found would attempt to execute a client command (registered command in Eclipse or in VSCode)
- Access to `XMLLanguageServer` instance to register/unregister command and send `xml/executeClientCommand` request to the client. Need to work out nice API around it as it looks like `XMLLanguageServer` or `XMLLanguageClient` was not meant to be accessed by extensions.

PR #906, opened 2020-10-08
Handle schema location URLs others than file, http, ftp protocols

BoykoAlex:
Custom URL for schema location requiring URL Handler registration, i.e custom protocol (not http(s), ftp or file) fails to be loaded if `ContentModelManager` caching flag is on.

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/model/ContentModelManager.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/model/ContentModelManager.java
@@ -310,6 +310,10 @@ public class ContentModelManager {
 				Path file = cacheResolverExtension.getCachedResource(resolvedUri);
 				if (file != null) {
 					cmDocument = modelProvider.createCMDocument(file.toUri().toString());
+				} else {
+					// It's not a file URL, and not a remote URL (ftp, http, https)? It's something else that might have a registered URL Handler.
+					// Therefore just try to load the model from the URL.
+					cmDocument = modelProvider.createCMDocument(resolvedUri);
 				}
 			} catch (CacheResourceDownloadingException e) {
 				// the DTD/XML Schema is downloading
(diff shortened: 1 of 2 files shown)


BoykoAlex:
Factored out from https://github.com/eclipse/lemminx/pull/905

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaCompletionExtensionsTest.java
```diff
@@ -226,6 +233,57 @@ public void noNamespaceSchemaLocationCompletion() throws BadLocationException {
 				c("End with '</Configuration>'", "/Configuration>"),
 				c("End with '</ViewDefinitions>'", "/ViewDefinitions>"), c("End with '</View>'", "/View>"));
 	}
+	
+	@Test
+	public void customProtocolNamespaceSchemaLocationCompletionWhenCachingOn() throws BadLocationException {
+		
+		URL.setURLStreamHandlerFactory(new URLStreamHandlerFactory() {
```
angelozerr, line 240: What about is other test implement too this static method setURLStreamHandlerFactory? I fear that we could have trouble, no?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/model/ContentModelManager.java
```diff
@@ -310,6 +310,10 @@ private CMDocument findCMDocument(String uri, String publicId, String systemId,
 				Path file = cacheResolverExtension.getCachedResource(resolvedUri);
 				if (file != null) {
 					cmDocument = modelProvider.createCMDocument(file.toUri().toString());
+				} else {
+					// It's not a file URL, and not a remote URL (ftp, http, https)? It's something else that might have a registered URL Handler.
+					// Therefore just try to load the model from the URL.
+					cmDocument = modelProvider.createCMDocument(resolvedUri);
```
angelozerr, line 316: I understand now why you did that (thanks for your test!) but I think it's not a clean fix. I think a correct fix should provide the capability to define protocol which benefit with cache or not.

Let me find an idea that I will try to suggest it in order to we can discuss together (if you like it or not).

angelozerr: @BoykoAlex I create a PR  https://github.com/eclipse/lemminx/pull/907 please review it and tell me if it works for you. Thanks!

BoykoAlex: Yes, it would work for me as well, lets go with #907 and I'll close this one

BoykoAlex:
#907 should take care of this

PR #907, opened 2020-10-09
Fix bug with custom protocol which doesn't load content model.

angelozerr:
Fix bug with custom protocol which doesn't load content model.

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/model/ContentModelManager.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/model/ContentModelManager.java
@@ -237,8 +237,7 @@ public class ContentModelManager {
 			boolean downloading = false;
 			Exception cacheError = null;
 			String resolvedUri = resolvedURIInfo.getResolvedURI();
-			boolean isFileResource = URIUtils.isFileResource(resolvedUri);
-			if (!isFileResource && cacheResolverExtension.isUseCache()) {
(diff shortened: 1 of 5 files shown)


BoykoAlex:
Yes, this should work for as well. Thanks!

BoykoAlex:
There is a sudden side-effect with the fix:

```
	@Override
	public XMLInputSource resolveEntity(XMLResourceIdentifier resourceIdentifier) throws XNIException, IOException {
		String url = resourceIdentifier.getExpandedSystemId();
		// Try to get the downloaded resource. In the case where the resource is
		// downloading but takes too long, a CacheResourceDownloadingException is
		// thrown.
		Path file = getCachedResource(url);
		if (file != null) {
			// The resource was downloaded locally, use it.
			return new XMLFileInputSource(resourceIdentifier, file);
		}
		return null;
	}
```

The first call gives me `null` URL then `getCachedResource(url)` call eventually gets to `CacheResourcesManager#isUseCacheFor(String url)` with the `null` url and throws NPE. The result of this NPE is XSD model loaded for my custom protocol URL not loaded (i.e. it becomes `null`)
I suggest just check for `null` somewhere... The change below would work for me (any other null check before it gets here would be just as sufficient for me)

```
	private boolean isUseCacheFor(String url) {
		for (String protocol : protocolsForCahe) {
			if (url != null && url.startsWith(protocol)) {
				return true;
			}
		}
		return false;
	}
```

angelozerr:
@BoykoAlex it should b efixed with https://github.com/eclipse/lemminx/pull/910

PR #908, opened 2020-10-09
Fix error range for cvc-complex-type-2.3

### LINKED ISSUE #885: Error range for cvc-complex-type.2.3
Given the schema
```xml
  <xs:element name="foo">
    <xs:complexType>
      <xs:sequence>
        <xs:element ref="bar" minOccurs="0" maxOccurs="unbounded"/>
        <xs:element ref="foo" minOccurs="0" maxOccurs="unbounded"/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
  <xs:element name="bar">
    <xs:complexType>
      <xs:sequence>
        <xs:element ref="bar" minOccurs="0" maxOccurs="unbounded"/>
        <xs:element ref="foo" minOccurs="0" maxOccurs="unbounded"/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
</xs:schema>
```
...and this document, which is an instance of the schema:
```xml
<foo
  xsi:noNamespaceSchemaLocation="close-tag-type.xsd">
  <bar>/bar></bar>
</foo>

```

The error range for `cvc-complex-type.2.3` is between the pipes: `<bar>/bar></bar|>|`
Instead, it should be between these pipes: `<|bar|>/bar></bar>`

datho7561:
Now handles text nodes that are only take up one line, as well as `<![CDATA[]]>`.

Closes #885

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLPositionUtility.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/XMLPositionUtility.java
@@ -81,7 +81,7 @@ public class XMLPositionUtility {
 
 	/**
 	 * Returns the attribute name range and null otherwise.
-	 * 
+	 *
 	 * @param attr the attribute.
 	 * @return the attribute name range and null otherwise.
 	 */
@@ -100,7 +100,7 @@ public class XMLPositionUtility {
 
 	/**
 	 * Returns the attribute value range and null otherwise.
-	 * 
+	 *
 	 * @param attr the attribute.
 	 * @return the attribute value range and null otherwise.
 	 */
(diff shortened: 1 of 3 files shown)


datho7561:
Sorry, I don't think I quite understand the second part. Is this so that it highlights between the pipes in this case?

```xml
<elementsOnly>
  |text content
  <child />
  more text content|
</elementsOnly>
```

datho7561:
^fixed the first case

angelozerr:
> Sorry, I don't think I quite understand the second part. Is this so that it highlights between the pipes in this case?

I think you will not have cvc-complex-type-2.3 error in this case.

datho7561:
Thanks for the help, Angelo! I avoided using `getData()`, and now the ranges don't include whitespace in CDATA sections.

angelozerr:
Works like a charm. Great job @datho7561 !

PR #912, opened 2020-10-13
No symbols computed if max symbols is zero

### LINKED ISSUE #898: Symbols Max Items Computed doesn't work for 0
If `xml.symbols.maxItemsComputed` is set to 0, then all the symbols are still computed.

datho7561:
Closes #898

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLSymbolSettings.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLSymbolSettings.java
@@ -18,13 +18,15 @@ import java.util.Arrays;
  */
 public class XMLSymbolSettings {
 
+	private static final int DEFAULT_MAX_ITEMS_COMPUTED = 5000;
+
 	private transient XMLExcludedSymbolFile[] excludedFiles;
 
 	private boolean enabled = true;
 
 	private String[] excluded;
 
-	private int maxItemsComputed;
+	private int maxItemsComputed = DEFAULT_MAX_ITEMS_COMPUTED;
 
 	public XMLExcludedSymbolFile[] getExcludedFiles() {
 		return excludedFiles;
@@ -60,9 +62,9 @@ public class XMLSymbolSettings {
 	/**
 	 * Given a file URI, this will check if it matches any of the given
 	 * file patterns.
-	 * 
+	 *
 	 * A uri is 'excluded' if it matches any of the given patterns.
-	 * 
+	 *
 	 * **Important:** Set the excluded file patterns before calling this using 'setExcluded()'.
 	 * @param uri
 	 * @return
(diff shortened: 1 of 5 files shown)


datho7561:
Unit test works, but in practice, this doesn't actually work.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLDocumentSymbolsTest.java
```diff
@@ -143,4 +143,21 @@ public void exceedSymbolLimit() {
...
+				"    <c>\n" + //
+				"      <d />\n" + //
+				"    </c>\n" + //
+				"  </b>" + //
+				"</a>";
+
+		XMLSymbolSettings settings = new XMLSymbolSettings();
+		settings.setMaxItemsComputed(0);
+
+		testDocumentSymbolsFor(xml, "test.xml", settings);
```
fbricon, line 160: what is the assertion here?

datho7561: These are no document symbols when the limit is zero. I will make this test a bit more clear.

datho7561: Please let me know if this clarifies the test

datho7561:
[ ] Note to self: address comment on reference to this PR

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLSymbolSettings.java
```diff
@@ -24,7 +24,7 @@
 
 	private String[] excluded;
 
-	private int maxItemsComputed;
+	private int maxItemsComputed = 5000;
```
angelozerr, line 27: Define a private constant like 

`private static final DEFAULT_MAX_ITEMS_COMPUTED = 5000;`

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/XMLAssert.java
```diff
@@ -766,6 +766,33 @@ public static void testDocumentSymbolsFor(XMLLanguageService xmlLanguageService,
 
 	}
 
+	public static void testDocumentSymbolsNumber(String xml, String fileURI, XMLSymbolSettings symbolSettings,
```
angelozerr, line 769: XMLAssert is a kind of framework to test the XML language services features.

As this XMLAssert class is used for any extension, I would like to provide only services test which are a lot of used (ex : test for completion, hover, validation, etc). 

I'm not sure that a lot extension will do some test with testDocumentSymbolsNumber. So please move those methods in XMLDocumentSymbolsTest.java. 

We could change that according user issues.

angelozerr:
Please write a comment in vscode-xml  / docs / Symbols about this 0 value. Thanks!

PR #926, opened 2020-11-05
Changelog for 0.14.0

datho7561:


angelozerr:
@datho7561 it misses some issues from vscode-xml like https://github.com/redhat-developer/vscode-xml/issues/220

PR #939, opened 2020-12-02
Replace 'ps -p' with 'kill -0' in ParentProcessWatcher

### LINKED ISSUE #936: Use `kill -0` instead of `ps -p` in  ParentProcessWatcher
To prevent some crashes when running in WSL Alpine Linux (see https://github.com/redhat-developer/vscode-java/issues/1711), we need to use `kill -0` instead of `ps -p` in [ParentProcessWatcher](https://github.com/eclipse/lemminx/blob/4d154e4bb6b46aa1c80dad514e41f55973680079/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ParentProcessWatcher.java#L91), like in https://github.com/eclipse/eclipse.jdt.ls/pull/1612/files

fbricon:
Fixes #936


rgrunber:
This works for me. Also behaves accordingly when a SIGKILL is sent to the parent (or a parent of the parent).

PR #942, opened 2020-12-08
xml catalog definition should be built-in

angelozerr:
xml catalog definition should be built-in

Fixes https://github.com/redhat-developer/vscode-xml/issues/375


angelozerr:
This PR embeds catalog.xsd. It doesn't provide tests because there are some existing tests with XML catalog.

To test locally the PR:

 * remove the local lemminx cache
 * disconnect internet (to check catalog.xsd comes from the embed catalog and not from the http url)
 * create a catalog.xml  file without DTD or XSD liek this:

```xml
<?xml version="1.0"?>
<catalog xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog">
</catalog>
```

You should benefit with completion, validation etc.

datho7561:
It doesn't use the embedded version if a `schemaLocation` is specified

angelozerr:
> It doesn't use the embedded version if a schemaLocation is specified. Is this intentional? I can see why this could be helpful.

Yes it's normal. If you use schemaLocation it means that you want to specify a given XSD. the usecase here is to provide a default XSD for `urn:oasis:names:tc:entity:xmlns:xml:catalog` namespace URI.

datho7561 (approved):

PR #950, opened 2020-12-21
Display if server is a native binary in init message

### LINKED ISSUE #949: Include message of if the server is a binary in the startup message
Right now, running the startup message when running a native-image of LemMinX lists the Java version as unknown, and gives no other indication that LemMinX is running as a native-image. Add a clear indication to the startup message that the server is running as a native-image.

datho7561:
Adds a 'Native Image' entry to the initialization information.
If LemMinX is running as a `native-image` binary, it is set to true.

Closes #949

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/ServerInfo.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/ServerInfo.java
@@ -21,6 +21,9 @@ public class ServerInfo {
 
   static final String MASTER = "master";
 
+  // https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java#L97
+  private static final boolean IS_NATIVE_IMAGE = "Substrate VM".equals(System.getProperty("java.vm.name"));
+
   private ResourceBundle rb = ResourceBundle.getBundle("git");
 
   public ServerInfo() {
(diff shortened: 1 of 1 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/ServerInfo.java
```diff
@@ -21,6 +21,9 @@
 
   static final String MASTER = "master";
 
+  // https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java#L97
+  static final boolean IS_NATIVE_IMAGE = "Substrate VM".equals(System.getProperty("java.vm.name"));
```
angelozerr, line 25: private static

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/ServerInfo.java
```diff
@@ -80,14 +83,15 @@ public String toString() {
...
-   * 
+   *
    * @return the formatted server details
    */
   public String details() {
     StringBuilder details = new StringBuilder();
     details.append("LemMinX Server info:");
     append(details, "Version", getVersion());
     append(details, "Java", getJava());
+    append(details, "Native Image", Boolean.toString(IS_NATIVE_IMAGE));
```
fbricon, line 94:
```java
if (IS_NATIVE_IMAGE) {
  append(details, "Native Image", null);
} else {
  append(details, "Java", getJava());
}
```

We might also display graalvm/substratevm version instead

datho7561: I added a "VM Version" entry that will show the `java.vm.version`, which will show the GraalVM version when running as a native-image, or the Java version if using a JDK/JRE.

fbricon:
LGTM

Screenshot 2021-01-14 at 11 02 03
Screenshot 2021-01-14 at 11 05 27

PR #952, opened 2020-12-22
Add new formatting setting `xml.format.splitAttributesIndentSize`

datho7561:
Add a new setting which controls the level of indentation of attributes with respect to their parent element when `xml.format.splitAttributes` is enabled.

Based off of a comment by Cong Wang (@I322871) on the vscode-xml Gitter

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java
@@ -17,7 +17,7 @@ import org.eclipse.lsp4j.FormattingOptions;
 /**
  * This class is the root of all formatting settings. It is necessary to update
  * this class for any new additions.
- * 
+ *
  * All defaults should be set here to eventually be overridden if needed.
  */
 public class XMLFormattingOptions extends FormattingOptions {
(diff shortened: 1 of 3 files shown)


datho7561:
- [ ] Write tests

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java
```diff
@@ -345,6 +348,25 @@ public boolean isPreserveAttrLineBreaks() {
...
+	}
+
+	/**
+	 * Returns the value of numberOfIndentsForSplitAttributes or zero if it was set to a negative value
+	 *
+	 * @return the value of numberOfIndentsForSplitAttributes or zero if it was set to a negative value
+	 */
+	public int getNumberOfIndentsForSplitAttributes() {
+		int numberOfIndentsForSplitAttributes = getNumber(NUMBER_OF_INDENTS_FOR_SPLIT_ATTRIBUTES).intValue();
+		return numberOfIndentsForSplitAttributes < 0 ? 0 : numberOfIndentsForSplitAttributes;
```
datho7561, line 367: this logic should probably go into the set function

datho7561: nope, because then you can set negative indentation through the client

datho7561:
Please test with redhat-developer/vscode-xml#386

datho7561:
need to rename commit...

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java
```diff
@@ -345,6 +348,25 @@ public boolean isPreserveAttrLineBreaks() {
 		}
 	}
 
+	/**
+	 * Sets the value of splitAttributesIndentSize
+	 *
+	 * @param splitAttributesIndentSize
```
datho7561, line 354: oops

PR #968, opened 2021-01-19
LemMinX crashes if one extension class cannot be created

### LINKED ISSUE #967: LemMinX crashes if one extension class cannot be created
I had this issue when I testing lemminx-maven in Eclipse IDE (a problem with Java compiler on my side). I notice when one extension class cannot be instanciated it crashes the XML LS and we loose all XML features (completion, diagnostics, etc).

To test this problem,you can add in the org.eclipse.lemminx.services.extensions.IXMLExtension

BAD_CLASS and you try to start test, everything is broken.

angelozerr:
LemMinX crashes if one extension class cannot be created

Fixes #967


angelozerr:
I could add test by adding the file `src/test/resources/META-INF/serves/org.eclipse.lemminx.services.extensions.IXMLExtension` with the content `BAD-CLASS` and tests should continue to work, but I would like to do that otherwise we will see a lot of logs for tests.

datho7561 (approved):
LGTM

PR #970, opened 2021-01-20
Preserve trailing whitespace of text content

datho7561:
When trimming trailing line whitespace for text content during formatting:
 * preserve trailing whitespace on last line of text-only nodes
 * trim the trailing whitespace on last line of text in mixed content

Closes redhat-developer/vscode-xml#377

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
@@ -754,6 +754,10 @@ class XMLFormatter {
 			// check if format range reaches the end of the document
 			if (this.endOffset == this.textDocument.getText().length()) {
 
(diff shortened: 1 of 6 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
```diff
@@ -81,7 +81,7 @@ public XMLFormatterDocument(TextDocument textDocument, Range range, SharedSettin
 		/**
```
datho7561, line 81: Drop this from the PR

datho7561:
Closing; this PR adds more hacks on top of existing hacks to get this use case to work. As well, there are test failures in this PR. It would be much nicer to do a formatting overhaul like in https://github.com/eclipse/lemminx/pull/1044, or at least rewrite how text content is handled in the formatter.

PR #974, opened 2021-01-29
Text nodes and mixed content formatting overhaul

datho7561:
Try to improve formatting handling of text content and mixed content.

Clsoes redhat-developer/vscode-xml#377


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
```diff
@@ -41,6 +41,7 @@
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.TextEdit;
+import org.jsoup.helper.StringUtil;
```
angelozerr, line 44: Please use StringUtils from lemminx.

datho7561:
It looks like I broke all the new formatting features when I was trying to fix regressions. This will need a lot more work.

datho7561:
FormatOverhaul

This is the case that is a regression: comments after text content are put on a new line instead of staying at the end of the line:

datho7561:
Closing this PR, since it doesn't work and doesn't clean up the code that much. We likely need to do a larger refactoring in order to improve the formatter.

PR #978, opened 2021-02-01
Wrap catalog DocumentLink path resolving in try/catch

### LINKED ISSUE #977: `InvalidPathException` is logged when attempting to generate a document link for a url in catalogs
System: win10 with AdoptOpenJDK 11, couldn't reproduce under Linux

example catalog to reproduce:
```xml
<catalog
  xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog"
  xsi:schemaLocation="urn:oasis:names:tc:entity:xmlns:xml:catalog
    http://www.oasis-open.org/committees/entity/release/1.1/catalog.xsd"
  prefer="public">
  <public publicId="value" uri="https://google.ca" />
</catalog>
```

datho7561:
Closes #977

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/catalog/URICatalogEntry.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/catalog/URICatalogEntry.java
@@ -9,6 +9,7 @@
 *******************************************************************************/
 package org.eclipse.lemminx.extensions.catalog;
 
+import java.nio.file.InvalidPathException;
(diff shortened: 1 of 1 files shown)


datho7561:
From my manual testing, this addresses the issue, but I don't know how to write a unit test for this.

rgrunber (approved):
If this fixes the issue, then I'm fine with it. Only thing I can think of is in cases where the catch handles the problem (and only that problem) it's usually customary to add a comment to the issue.

datho7561:
I'll do an amend for that.

PR #988, opened 2021-02-10
HTML/Markdown -> HTML -> just Markdown

### LINKED ISSUE #962: Formatting documentation strings in hover popups.
The JSON language server supports markdown formatted docstrings, which enables us to write nice looking popups such as:

grafik

Can LemMinX do something like that as well? 

It appears all `<xs:documentation>` contents is streamed as unformatted plain text with all kinds of whitespace and newlines etc. removed. So even plain text formatted docstring as the one from the following example result in poor user experience at the moment.

**Result**
grafik

**Source**
```xsl
 <xs:annotation>
  <xs:documentation>
   See http://www.w3.org/XML/1998/namespace.html and
   http://www.w3.org/TR/REC-xml for information about this namespace.

    This schema document describes the XML namespace, in a form
    suitable for import by other schema documents.  

    Note that local names in this namespace are intended to be defined
    only by the World Wide Web Consortium or its subgroups.  The
    following names are currently defined in this namespace and should
    not be used with conflicting semantics by any Working Group,
    specification, or document instance:

    base (as an attribute name): denotes an attribute whose value
         provides a URI to be used as the base for interpreting any
         relative URIs in the scope of the element on which it
         appears; its value is inherited.  This name is reserved
         by virtue of its definition in the XML Base specification.

    lang (as an attribute name): denotes an attribute whose value
         is a language code for the natural language of the content of
         any element; its value is inherited.  This name is reserved
         by virtue of its definition in the XML specification.
  
    space (as an attribute name): denotes an attribute whose
         value is a keyword indicating what whitespace processing
         discipline is intended for the content of the element; its
         value is inherited.  This name is reserved by virtue of its
         definition in the XML specification.

    Father (in any context at all): denotes Jon Bosak, the chair of 
         the original XML Working Group.  This name is reserved by 
         the following decision of the W3C XML Plenary and 
         XML Coordination groups:

             In appreciation for his vision, leadership and dedication
             the W3C XML Plenary on this 10th day of February, 2000
             reserves for Jon Bosak in perpetuity the XML name
             xml:Father
  </xs:documentation>
 </xs:annotation>
```

What the client receives is:

```
:: --> LemMinX textDocument/hover(583): {'textDocument': {'uri': 'file:///C:/Apps/Sublime%20Text/Data/Packages/Markdown%20Extras/Snippets/Link.sublime-snippet'}, 'position': {'line': 2, 'character': 5}}
:: <<< LemMinX 583: {'contents': {'value': 'See http://www.w3.org/XML/1998/namespace.html and http://www.w3.org/TR/REC-xml for information about this namespace. This schema document describes the XML namespace, in a form suitable for import by other schema documents. Note that local names in this namespace are intended to be defined only by the World Wide Web Consortium or its subgroups. The following names are currently defined in this namespace and should not be used with conflicting semantics by any Working Group, specification, or document instance: base (as an attribute name): denotes an attribute whose value provides a URI to be used as the base for interpreting any relative URIs in the scope of the element on which it appears; its value is inherited. This name is reserved by virtue of its definition in the XML Base specification. lang (as an attribute name): denotes an attribute whose value is a language code for the natural language of the content of any element; its value is inherited. This name is reserved by virtue of its definition in the XML specification. space (as an attribute name): denotes an attribute whose value is a keyword indicating what whitespace processing discipline is intended for the content of the element; its value is inherited. This name is reserved by virtue of its definition in the XML specification. Father (in any context at all): denotes Jon Bosak, the chair of the original XML Working Group. This name is reserved by the following decision of the W3C XML Plenary and XML Coordination groups: In appreciation for his vision, leadership and dedication the W3C XML Plenary on this 10th day of February, 2000 reserves for Jon Bosak in perpetuity the XML name xml:Father\r\n\r\nSource: [sublime-snippet.xsd](file:/C:/Apps/Sublime%20Text/Data/Packages/LSP-lemminx/catalogs/sublime/sublime-snippet.xsd)', 'kind': 'markdown'}, 'range': {'start': {'line': 2, 'character': 3}, 'end': {'line': 2, 'character': 13}}}
```

Note: I injected the content into an XSD called _sublime-snippet.xsd_ which I am working on in order to provide intelligence features to write Sublime Text / TextMate resources.

datho7561:
Treat `xs:documentation` and `xs:appinfo` as mixed Markdown and HTML, with the HTML tags escaped using entities. Unescape the entities, then render the Markdown portions into HTML using the commonmark-java library. Finally, use the existing code to convert this HTML into Markdown that doesn't contain any HTML tags.

Closes #962 

### DIFF
--- a/org.eclipse.lemminx/pom.xml
+++ b/org.eclipse.lemminx/pom.xml
@@ -267,5 +267,10 @@
 			<version>9.4.38.v20210224</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.commonmark</groupId>
+			<artifactId>commonmark</artifactId>
+			<version>0.17.1</version>
+		</dependency>
 	</dependencies>
 </project>
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/contentmodel/XSDAnnotationModel.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/contentmodel/XSDAnnotationModel.java
@@ -54,7 +54,7 @@ class XSDAnnotationModel {
 
 	/**
 	 * Returns content from appinfo element(s)
-	 * 
+	 *
 	 * @return content from appinfo element(s)
 	 */
 	public List<String> getAppInfo() {
@@ -63,7 +63,7 @@ class XSDAnnotationModel {
 
 	/**
 	 * Returns content from documentation elements(s)
-	 * 
+	 *
 	 * @return content from documentation elements(s)
 	 */
 	public List<String> getDocumentation() {
@@ -72,7 +72,7 @@ class XSDAnnotationModel {
 
 	/**
 	 * Returns documentation content from the provided collection of annotations
-	 * 
+	 *
 	 * @param annotations the collection of annotations
 	 * @return documentation content from the provided collection of annotations
 	 */
@@ -82,7 +82,7 @@ class XSDAnnotationModel {
 
 	/**
 	 * Returns documentation content from the provided collection of annotations
-	 * 
+	 *
 	 * @param annotations the collection of attribute value annotations
 	 * @param value       the attribute value to find documentation content for
 	 * @return documentation content from the provided collection of annotations
@@ -115,7 +115,7 @@ class XSDAnnotationModel {
 
 	/**
 	 * Returns appinfo content from the provided collection of annotations
-	 * 
+	 *
 	 * @param annotations the collection of annotations
 	 * @return appinfo content from the provided collection of annotations
 	 */
@@ -125,7 +125,7 @@ class XSDAnnotationModel {
 
 	/**
 	 * Returns appinfo content from the provided collection of annotations
-	 * 
+	 *
 	 * @param annotations the collection of attribute value annotations
 	 * @param value       the attribute value to find appinfo content for
 	 * @return appinfo content from the provided collection of annotations
@@ -147,7 +147,7 @@ class XSDAnnotationModel {
 
 	/**
 	 * Returns the prefix (ie. xs) from the provided collection of annotations
-	 * 
+	 *
 	 * @param annotations the collection of annotations
 	 * @return the prefix (ie. xs) from the provided collection of annotations
 	 */
@@ -157,10 +157,10 @@ class XSDAnnotationModel {
 
 	/**
 	 * Returns the prefix (ie. xs) from the provided collection of annotations
-	 * 
+	 *
 	 * Prerequisite: <code>value</code> should be provided if <code>annotations</code>
 	 * is a collection of attribute value annotations
-	 * 
+	 *
 	 * @param annotations the collection of annotations
 	 * @param value       the attribute value
 	 * @return the prefix (ie. xs) from the provided collection of annotations
@@ -192,20 +192,20 @@ class XSDAnnotationModel {
 			return null;
 		}
 	}
-	
+
 	/**
-	 * Returns the <code>XSAnnotation</code> instance for the 
+	 * Returns the <code>XSAnnotation</code> instance for the
 	 * provided <code>annotation</code>
-	 * 
+	 *
 	 * If <code>value</code> is provided, the <code>XSAnnotation</code> for
 	 * an attribute value will be searched for
-	 * 
+	 *
 	 * If not provided, the <code>XSAnnotation</code> for
 	 * an attribute or element will be searched for
-	 * 
-	 * @param annotation the annotation object 
+	 *
+	 * @param annotation the annotation object
 	 * @param value      the attribute value
-	 * @return the <code>XSAnnotation</code> instance for the 
+	 * @return the <code>XSAnnotation</code> instance for the
 	 * provided <code>annotation</code>
 	 */
 	private static XSAnnotation getXSAnnotation(XSObject annotation, String value) {
@@ -258,9 +258,9 @@ class XSDAnnotationModel {
 			super.endElement(uri, localName, qName);
 			if (current != null) {
 				if (qName.endsWith(APPINFO_ELEMENT)) {
-					addIfNonEmptyString(model.appInfo, normalizeSpace(current.toString()));
+					addIfNonEmptyString(model.appInfo, current.toString().trim());
 				} else if (qName.endsWith(DOCUMENTATION_ELEMENT)) {
-					addIfNonEmptyString(model.documentation, normalizeSpace(current.toString()));
+					addIfNonEmptyString(model.documentation, current.toString().trim());
 				}
 				current = null;
 			}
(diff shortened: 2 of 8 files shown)


datho7561:
[ ] More thorough testing
[ ] Post different in JAR and binary size due to commonmark-java dependency

angelozerr:
@datho7561 could you please attach in this PR:

 * a sample text (HTML + markdown) covered by this PR
 * a screenshot before your PR
 * a screenshot after your PR just to see quickly your improvement (which will be helpful too for history).

Thanks!

datho7561:
Requiring `commonmark-java` adds 0.1 Mb to the uber jar, and less than 1 Mb to the Linux binary.

Here is a example schema I made to play with. I'll attach the photos soon.

```xml
  <xs:element name="root">
    <xs:annotation>
      <xs:documentation>
        # root

        This is the root element of the document.

        Use it like this:

        ```
        &amp;lt;root>
          ...
        &amp;lt;/root>
        ```

        *Since 0.1.0*
      </xs:documentation>
    </xs:annotation>
    <xs:complexType>
      <xs:sequence>
        <xs:element name="container" maxOccurs="unbounded">
          <xs:annotation>
            <xs:documentation>
              ## container

              Groups &lt;em>several&lt;/em> `content` together in a logical structure.

              `&lt;em> Doesn't display. lt still needs to be double escaped. &lt;/em>`

              ![whale image](https://openclipart.org/image/400px/3243)
            </xs:documentation>
          </xs:annotation>
          <xs:complexType>
            <xs:sequence>
              <xs:element name="content" maxOccurs="unbounded" type="xs:string">
                <xs:annotation>
                  <xs:documentation>
                    ## content
                    &lt;p> This node represents *text* content. &lt;/p>
                    &lt;p> This is the second [paragraph](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p) in the documentation. &lt;/p>
                  </xs:documentation>
                </xs:annotation>
              </xs:element>
            </xs:sequence>
          </xs:complexType>
        </xs:element>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
</xs:schema>
```

datho7561:
| Before                             | After                                |
| rootBefore
 | withMarkdownRoot |
| containerBefore | withMarkdownContainer |
| contentBefore | withMarkdownContent |

angelozerr:
It looks very promising! 

@fbricon @rgrunber  what do you think about this PR? I

f the result better perhaps vscode-java (markdown converter is a copy/paste from JDT LS) should adopt this PR too?

rgrunber:
Seems pretty small at 160kb, and no other runtime deps. Upstream seem sot have one maine contributor for a while, but they seem fairly active. Definitely looks better.

How does this compare to https://github.com/kotcrab/remark-java/ though ?

datho7561:
commonmark-java is for Markdown -> Java, whereas remark-java is for HTML -> Markdown without any HTML tags.

org.eclipse.lemminx/pom.xml
```diff
@@ -267,5 +267,10 @@
 			<version>9.4.38.v20210224</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.commonmark</groupId>
+			<artifactId>commonmark</artifactId>
+			<version>0.17.1</version>
```
rgrunber, line 273: I see a CQ of https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22263 for 0.15.1 (before the groupId changed to org.commonmark), but we'll need to file a new one for this version.

datho7561: I will take a look into filing the CQ for it

datho7561: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23347

datho7561: From what I can tell, the CQ for commonmark is approved.

rien333:
> Closing this PR, since Markdown documentation in XSD files is just a hack and not something that we intend to support

In what sense is being able to render/use markdown considered a hack? The [W3C recommendation](https://www.w3.org/TR/xmlschema11-1/#Annotation)  states: 

> ### 2.2.6 Annotation Components
>
> An annotation is information for human and/or mechanical consumers. The interpretation of such information is not defined in this specification.

From this, I gather that we are not 1) prescribed a particular documentation format, 2) that human readability is of the essence. And indeed, whereas my less technical colleagues would definitely be confused by this,

```xml
<xsd:annotation>
	<xsd:documentation>
	<p>Indication of the type of fruit.</p>
	<p>Possible values: </p> 
        <ul>
		<li><code>banana</code>: foo bar</li>
		<li><code>apple</code>: foo bar</li>
		<li><code>kiwi</code>: foo bar</li>
	</ul>
	<xsd:annotation>
<xsd:documentation>		
``` 

they will have little trouble parsing the following, due to markdown's adherence to existing plain text formatting conventions:

```xml
<xsd:annotation>
	<xsd:documentation>
	Indication of the type of fruit.
	
	Possible values:
	    * `banana`: foo bar
	    * `apple`: foo bar
	    * `kiwi`: foo bar
	<xsd:annotation>
<xsd:documentation>		
``` 

Coupled with the fact that, these days, a lot of documentation is markdown formatted (and hence, conventionally undestood), I would argue that markdown better alligns with the W3C recommendations than HTML. 

> Closing this PR, since Markdown documentation in XSD files is ... not something that we intend to support

Finally, I would add that many LSPs do have markdown documentation support. Why, then, would it be out of scope of this project?

datho7561:
Currently, we consider supporting HTML more important than supporting Markdown, since many existing XSDs use HTML.

> In what sense is being able to render/use markdown considered a hack?

Currently, lemminx doesn't support using Markdown in documentation for XSDs properly. However, sometimes it seems like it works. The reasoning behind this is complex:

VS Code only supports displaying Markdown, not HTML, in the hover documentation. As a result, in order to display the HTML documentation in VS Code, we need to "decompile" the HTML into Markdown (the opposite direction than you normally do). However, we currently don't try to escape any existing Markdown special characters in the HTML content during the decompilation process. As a result, sometimes Markdown content is conserved during this conversion and renders properly in VS Code. However, there are some edge cases where the conversion causes things to render in ways you wouldn't expect (see the before/after table above). The fix I worked on in this PR would involve rendering the Markdown first, then decompiling the HTML back into Markdown. This approach isn't flawless though, since some documentation with only HTML content contains characters that are considered formatting in Markdown, so this conversion will break rendering HTML properly.

**TLDR**: it may seem like lemminx supports rendering markdown, but this is an unintended consequence of the implementation

> Why, then, would it be out of scope of this project?

The main reasons we believe it's not worthwhile to do this work:
- We don't have much time to work on lemminx, and other issues are more pressing
- Adding Markdown support involves adding a Markdown --> HTML compiler as a dependency (currently we only have a HTML --> Markdown decompiler as a dependency). It is hard to find a Markdown --> HTML compiler that supports Java 8, is actively maintained, and we are sure will continue to be maintained
- Adding Markdown support might break rendering for some HTML documentation, which we consider more important
- lemminx is an LSP for writing XML. We don't have the resources to build a full IDE for writing XSD. As such, we assume that the developer is working with an existing XSD. Many existing XSDs use HTML documentation, since Markdown was not as ubiquitous as it is now when those XSDs were developed.

PR #990, opened 2021-02-17
Rename not completing with binary

angelozerr:
Rename not completing with binary

Fixes https://github.com/redhat-developer/vscode-xml/issues/424

### DIFF
--- a/org.eclipse.lemminx/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/org.eclipse.lemminx/src/main/resources/META-INF/native-image/reflect-config.json
@@ -1178,6 +1178,14 @@
 		}]
 	},
 	{
+		"name": "org.eclipse.lsp4j.RenameParams",
+		"allDeclaredFields": true,
+		"methods": [{
+			"name": "<init>",
+			"parameterTypes": []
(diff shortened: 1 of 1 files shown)


rgrunber:
Was able to reproduce, and this change fixes the issue. :+1: 

FWIW, you can reproduce by downloading a JVM containing GraalVM from https://github.com/graalvm/graalvm-ce-builds/releases and placing it locally : `$HOME/.local/bin/openjdk1.8.0_272-jvmci-21.0-b03` . You can then run `JAVA_HOME=$HOME/.local/bin/openjdk1.8.0_272-jvmci-21.0-b03/ mvn clean verify -Pnative -DskipTests` . It should generate the binary in the target folder. You can then `xml.server.binary.path` and `xml.server.preferBinary` to use it.

angelozerr:
Thanks a lot @rgrunber for your review and your great explanation :)

JVimes:
The win32 binary from the GitHub artifacts fixed it for me. 👍

angelozerr:
> The win32 binary from the GitHub artifacts fixed it for me. 

Great! Thanks for your feedback.

PR #994, opened 2021-02-24
Use User-Agent "LemMinX" when downloading schemas

datho7561:
Closes redhat-developer/vscode-xml#429


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/CacheResourcesManager.java
```diff
@@ -167,13 +167,15 @@ public Path getResource(final String resourceURI) throws IOException {
 				String actualURI = resourceURI;
 				URL url = new URL(actualURI);
 				conn = url.openConnection();
+				conn.setRequestProperty("User-Agent", "LemMinX");
```
angelozerr, line 170: Use constants like USER_AGENT_KEY = "User-Agent" and USER_AGENT_VALUE = "LemMinX"

PR #996, opened 2021-02-25
Add eclipse/dash-licenses for tracking IP metadata.

rgrunber:
- Enable with 'verify-iplog' profile under the 'verify' lifecycle
- Summary generated under ${project.build.directory}/dash/summary

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,13 @@
 			<id>cbi-release</id>
 			<url>https://repo.eclipse.org/content/repositories/cbi-releases/</url>
 		</pluginRepository>
+		<pluginRepository>
+			<id>dash-licenses-snapshots</id>
+			<url>https://repo.eclipse.org/content/repositories/dash-licenses-snapshots/</url>
+			<snapshots>
+			    <enabled>true</enabled>
+			</snapshots>
+		</pluginRepository>
 	</pluginRepositories>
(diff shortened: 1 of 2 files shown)


datho7561:
From my testing, this doesn't seem to work under Java 8.

rgrunber:
Yup. https://github.com/eclipse/dash-licenses/blob/master/pom.xml#L29 . I believe they're using the `var` keyword. This actually prevents me from running it on some other Jenkins instances at Eclipse until I upgrade them to Java 11.

datho7561:
Hmm. Is the point of this plugin to run it in CI/CD?

rgrunber:
While it would be nice to run it as part of the build, another case is to have a commiter/project lead run it on their local machine and submit the generated information as part of the IPlog review.

pom.xml
```diff
@@ -296,5 +303,26 @@
 				</pluginManagement>
 			</build>
 		</profile>
+		<profile>
+			<id>verify-iplog</id>
```
datho7561, line 307: Maybe add a doc for how to generate the report, and mention that generating the report requires Java 11.

rgrunber: I've made a proposed addition to the README.

datho7561 (approved):

Issue #995, opened 2021-02-25
Do an Eclipse release review


mickaelistria:
LemMinX did not participate in a release review, while Eclipse Development Process mandates at least 1 release review per year. This is becoming a concern, see https://www.eclipse.org/lists/tools-pmc/msg05151.html .
Please plan a release review soon to show that LemMinX is playing by the rules of Eclipse Development Process.

rgrunber:
We should create/update the release plan for the upcoming 0.15.1 as described in https://www.eclipse.org/projects/handbook/#releases-plan .

We could also double-check the items in https://www.eclipse.org/projects/handbook/#checklist .

rgrunber:
So we've got the [0.16.0 Review Documentation](https://projects.eclipse.org/projects/technology.lemminx/releases/0.16.0) , the [IPLog](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23070) , so we'll just need to request PMC Approval (which goes to the pmc mailing list) and then we can start the review.

datho7561:
Before requesting PMC approval, should we push in a PR that includes `<name>` and `<organization>`?

rgrunber:
> Before requesting PMC approval, should we push in a PR that includes `<name>` and `<organization>`?

Yes, please do. `<name>` makes sense. If you want to do `<organization>` you can do that as well, but I have seen projects that don't.

datho7561:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=572192

datho7561:
https://github.com/eclipse/lemminx/releases/tag/0.16.0

PR #1006, opened 2021-03-12
Add project name to pom.xml

datho7561:
A part of #995


rgrunber (approved):
Initially I was thinking `XML Language Server (LemMinX)`, but I guess technically it's LemMinX, and the Eclipse to indicate organization. The language server part is more of a description.

datho7561:
Okay. I'll look into merging once https://www.githubstatus.com/ is doing better (since the binary GitHub Action will likely fail otherwise)

PR #1008, opened 2021-03-23
Use related information for unclosed elements

### LINKED ISSUE #963: Linting makes editing hard
grafik

The screenshot shows a simple file which only misses the end tag `</xs:schema>`. This little issue results in the whole document being marked invalid, which doesn't help much. The file becomes harder to read while editing and doesn't tell much about the origin of the error.

I'd expect only the opening tag name and/or the end of the document being marked invalid. The content in between is very correct.

datho7561:
Revert `MarkupEntityMismatch` and `ETagRequired` error ranges so they only cover the start tag element name.

Keep existing `CodeAction` behaviour intact.

If the client supports `DiagnosticRelatedInformation`, add the expected location of the close tag as related info to `MarkupEntityMismatch` and `ETagRequired` errors.

Closes #963

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
@@ -49,7 +49,7 @@ public class DOMElement extends DOMNode implements org.w3c.dom.Element {
 
 	/*
 	 * (non-Javadoc)
-	 * 
+	 *
 	 * @see org.w3c.dom.Node#getNodeType()
 	 */
 	@Override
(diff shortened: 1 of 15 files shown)


angelozerr:
@datho7561 could you add a demo please.

datho7561:
related-info-element-with-attributes
related-info-element-with-child
related-info-root-element

angelozerr:
It looks very promising!

Given XML and with related information : 

```xml
<root>

</root
```

It should higlight the root of end tag instead of start tag.

datho7561:
Okay, it should handle this case properly now.

datho7561:
Okay, I moved the static function to `DOMElement`

angelozerr:
Very good job @datho7561 !

PR #1009, opened 2021-03-24
New setting to prevent auto self close destroying content

datho7561:
Adds the setting `xml.completion.autoCloseRemovesContent`, which defaults to `true` (to preserve existing behaviour). When the setting is set to `false`, and the auto close tags funcitonality is enabled, `<bbb />` and `</aaa>` won't be deleted when an `/` is inserted at the pipe:

```xml
<aaa |
  <bbb />
</aaa>
```

Please test with redhat-developer/vscode-xml#441, which registers the new setting.

Closes redhat-developer/vscode-xml#440


rgrunber (approved):
Looks good to me. the preference is behaving as expected and removes the interior content provided it isn't text. I would merge this.

PR #1014, opened 2021-04-08
Update Xerces to 2.12.1

### LINKED ISSUE #1013: Update Xerces to 2.12.1
https://github.com/eclipse/lemminx/blob/81298daf1d6aab4a5fd06fab3675affbf8fbff5d/org.eclipse.lemminx/pom.xml#L226

Xerces 2.12.1 was released on 12/01/2020. It includes a bunch of fixes (many related to XML 1.1). See http://xerces.apache.org/xerces2-j/releases.html

fbricon:
I only checked it compiles and tests pass locally.
Haven't tried binary generation or actual smoke tests

Fixes #1013

### DIFF
--- a/org.eclipse.lemminx/pom.xml
+++ b/org.eclipse.lemminx/pom.xml
@@ -223,7 +223,7 @@
 		<dependency>
 			<groupId>xerces</groupId>
 			<artifactId>xercesImpl</artifactId>
-			<version>2.12.0</version>
+			<version>2.12.1</version>
 		</dependency>
 		<dependency>
 			<groupId>xml-apis</groupId>


rgrunber:
[diff.txt](https://github.com/eclipse/lemminx/files/6327784/diff.txt)

Looks to be pretty minor.

datho7561:
It doesn't look like this update will provide an easy route to supporting XSD 1.1, since XSD 1.1 is not available through the SAX parser, just for JAXP [(see "How do I validate against XML schema?")](https://xerces.apache.org/xerces2-j/faq-xs.html). I'll continue to investigate if there are any regressions.

datho7561 (approved):

PR #1018, opened 2021-04-19
Code cleanup

fbricon:
Apply several refactorings from https://github.com/JnRouvignac/AutoRefactor/

- Use String.replace instead of replaceAll when possible
- Proper usage of StringBuilders
- Use lazy logical statement rather than eager
- Use diamond operator
- Use String.contains instead of indexOf
- All in one method rather than for loop
- Add brackets to control statements

Do not squash this PR!


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ListLineTracker.java
```diff
@@ -216,21 +228,24 @@ public int getOffsetAt(Position position) throws BadLocationException {
...
 		return offset;
 	}
 
 	@Override
 	public final Line getLineInformationOfOffset(int position) throws BadLocationException {
-		if (position > fTextLength)
-			throw new BadLocationException("Offset > length: " + position + " > " + fTextLength);  //$NON-NLS-1$//$NON-NLS-2$
+		if (position > fTextLength) {
+			throw new BadLocationException("Offset > length: " + position + " > " + fTextLength);
+		}  //$NON-NLS-1$//$NON-NLS-2$
```
datho7561, line 242: Need to manually move the comment in this case

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
```diff
@@ -481,7 +481,7 @@ private void formatElement(DOMElement element) throws BadLocationException {
 						for (DOMNode child : element.getChildren()) {
 							boolean textElement = !child.isText();
 
-							hasElements = hasElements | textElement;
+							hasElements = hasElements || textElement;
```
datho7561, line 484:
```java
hasElements = hasElements || !child.isText();
```

datho7561 (changes requested):

datho7561 (approved):

PR #1020, opened 2021-04-20
Print errors to standard error if flag set

### LINKED ISSUE #1019: Output error stream when developing language server
Set up the default error stream to be output somewhere when developing the server. This allows the developer to access the Exception traces that happen before the language client connects to the server.

In particular, this would be very helpful when trying to debug the native image of the server. Crashes can occur during the initialization request, since all the classes that form the initialization request must be registered for reflection in order to successfully send the request.

datho7561:
If the environment variable `LEMMINX_DEBUG` is set, then Java's errors will not be disconnected from standard error.

Closes #1019


rgrunber:
This change works for me. Setting LEMMINX_DEBUG (eg. in the vscode launch as an example) worked for me and printed some simulated errors.

PR #1024, opened 2021-05-06
Fix ability to disable "no grammar" message

### LINKED ISSUE #467: Error range for ct-props-correct.3

datho7561:
Fixes redhat-developer/vscode-xml#467

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationSettings.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationSettings.java
@@ -182,6 +182,7 @@ public class XMLValidationSettings {
 			this.namespaces = settings.namespaces;
 			this.schema = settings.schema;
 			this.enabled = settings.enabled;
+			this.noGrammar = settings.noGrammar;
 			this.disallowDocTypeDecl = settings.disallowDocTypeDecl;
 			this.resolveExternalEntities = settings.resolveExternalEntities;
 		}
--- a/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/settings/SettingsMergeTest.java
+++ b/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/settings/SettingsMergeTest.java
@@ -1,0 +1,38 @@
+/*******************************************************************************
+* Copyright (c) 2021 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.settings;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+import java.lang.reflect.Field;
+
+import org.eclipse.lemminx.extensions.contentmodel.settings.XMLValidationSettings;
+import org.eclipse.lsp4j.PublishDiagnosticsCapabilities;
+import org.junit.jupiter.api.Test;
+
+public class SettingsMergeTest {
+
+	@Test
+	public void testValidationSettingsMerge() throws IllegalArgumentException, IllegalAccessException {
+		XMLValidationSettings settings = new XMLValidationSettings();
+		settings.setEnabled(false);
+		settings.setDisallowDocTypeDecl(true);
+		settings.setNoGrammar("ignore");
+		settings.setResolveExternalEntities(true);
+		settings.setCapabilities(new PublishDiagnosticsCapabilities());
+		settings.setNamespaces(null);
+		settings.setSchema(null);
+		XMLValidationSettings mergeTargetSettings = new XMLValidationSettings();
+		mergeTargetSettings.merge(settings);
+		assertEquals(settings, mergeTargetSettings);
+	}
+
+}


fbricon:
Can we have unit tests that check there won't be another regression in the future pretty please?

datho7561:
I don't know how to effectively unit test this scenario, but I attached an attempt

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationSettings.java
```diff
@@ -12,6 +12,8 @@
  */
 package org.eclipse.lemminx.extensions.contentmodel.settings;
 
+import java.lang.reflect.Field;
```
fbricon, line 15: nope

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/settings/SettingsMergeTest.java
```diff
@@ -0,0 +1,49 @@
...
+public class SettingsMergeTest {
+
+	@Test
+	public void testValidationSettingsMerge() throws IllegalArgumentException, IllegalAccessException {
+		XMLValidationSettings settings = new XMLValidationSettings();
+		for (Field field : XMLValidationSettings.class.getFields()) {
+			Type type = field.getGenericType();
+			switch (type.getTypeName()) {
+			case "boolean":
+				field.set(settings, false);
```
fbricon, line 29: set to true? as false is the default value

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/settings/SettingsMergeTest.java
```diff
@@ -0,0 +1,49 @@
...
+				break;
+			case "int":
+			case "short":
+			case "long":
+			case "char":
+			case "byte":
+			case "float":
+			case "double":
+				field.set(field, 0);
```
fbricon, line 38: set to 1, as 0 is the default value

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationSettings.java
fbricon: still no

datho7561: I don't know of any other way to test for the regression where we add a new field but don't manage it in the merge function. Do you have any ideas?

fbricon: It's fine in the test, useless in XMLValidationSettings.java

datho7561: Okay. Thank you, missed that

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/settings/SettingsMergeTest.java
```diff
@@ -0,0 +1,56 @@
...
+		XMLValidationSettings mergeTargetSettings = new XMLValidationSettings();
+		mergeTargetSettings.merge(settings);
+		assertEquals(settings, mergeTargetSettings);
+	}
+
+	/**
+	 * Mutates all the values of all the fields, under the assumption that all Object fields are non-null
+	 */
+	public static void mutateValues(Object obj) throws IllegalArgumentException, IllegalAccessException {
+		for (Field field : XMLValidationSettings.class.getFields()) {
```
rgrunber, line 35: `XMLValidationSettings.class.getFields()` should return all publicly accessible fields, but they're all private, so it's empty for me. I think to do it this way you'd have to use `XMLValidationSettings.class.getDeclaredFields()` in combination with setAccessible(true) which is basically a no-go since Java 9 so I'd probably avoid this.

Alternatively, could you not just use the setters and confirm whether merge does the right thing ?

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/settings/SettingsMergeTest.java
```diff
@@ -0,0 +1,56 @@
...
+			case "int":
+			case "short":
+			case "long":
+			case "char":
+			case "byte":
+			case "float":
+			case "double":
+				field.set(field, (Integer)field.get(obj) == 0 ? 1 : 0);
+				break;
+			default:
```
rgrunber, line 50: noGrammar being a String would probably fall under this case, but because it's already null, I don't think it would catch the case of it not being properly set.

rgrunber (changes requested):
The change definitely fixes things, but just a question of fixing the testcase.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/settings/SettingsMergeTest.java
datho7561: Good catch. I took this route because if we add a new field in the future, then we could end up with the same bug. I'll switch back to using the getters and setters

rgrunber:
I would just restore it to how it was before with the manual setting of methods and calling merge. No need for setAccessible(..). IMO, I think it's fine to merge.

Anyone adding to validation settings should add to that testcase and they would catch any issues with merge.

PR #1025, opened 2021-05-06
Bump Guava to 30.1.1

datho7561:
Replaces #1011

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
 			<dependency>
 				<groupId>com.google.guava</groupId>
 				<artifactId>guava</artifactId>
-				<version>27.0.1-jre</version>
+				<version>30.1.1-jre</version>
 				<exclusions>
 					<exclusion>
 						<groupId>com.google.guava</groupId>


pom.xml
```diff
@@ -56,7 +56,7 @@
 			<dependency>
 				<groupId>com.google.guava</groupId>
 				<artifactId>guava</artifactId>
-				<version>27.0.1-jre</version>
+				<version>30.0-jre</version>
```
fbricon, line 59: Why not [30.1.1-jre](https://search.maven.org/artifact/com.google.guava/guava/30.1.1-jre/bundle)?

datho7561: I went with the minimum version that listed the vulnerability as fixed by CRDA. I'll test 30.1.1 to see if it works, and use it instead if it does.

rgrunber (approved):
Looks like it's approved ( maven/mavencentral/com.google.guava/guava/30.1.1-jre, Apache-2.0 AND CC0-1.0 AND LicenseRef-Public-Domain, approved, CQ23244 ) and tests are all passing so I think this is fine to merge.

PR #1036, opened 2021-05-17
Fix range formatting in binary server

### LINKED ISSUE #1035: Range formatting does not work when using binary
When attemping to range format when using the binary version of the server, the following exception is thrown:

```
java.lang.RuntimeException: Unable to invoke no-args constructor for class org.eclipse.lsp4j.DocumentRangeFormattingParams. Registering an InstanceCreator with Gson for this type may fix this problem.
```

datho7561:
Closes #1035


rgrunber (approved):
This fixes the issue for me.

Issue #1041, opened 2021-05-19
Preserve invalid content while formatting


angelozerr:
When formatting is done with invalid content theinvalid content is lost.

For instance given this XML content:

```xml
<a=
```

After formatting we lose the invalid content `=`

```xml
<a
```

There are several issues with this problem:

 * https://github.com/eclipse/lemminx/issues/1034
 * https://github.com/eclipse/lemminx/issues/679
 * https://github.com/eclipse/lemminx/issues/675
 * https://github.com/redhat-developer/vscode-xml/issues/624
 * 
and a PR which tries to fix the problem with invalid content inside  tag element https://github.com/eclipse/lemminx/pull/766

Today the formatting process load the XML content to format in a DOM Document and the DOM document is visited and we generate the XML content according DOM node type. At the end of the process we return ONE TextEdit which contains the generates XML string.

I think it's not the right way. Indeed after studying the formatting support in the WTP XML Editor (which preserve invalid content), the right way is to create several TextEdits (instead of one textedit) when a format must be done (ex : we create a TextEdit to format a start tag element to indent it, create N TextEdit  to format N start tag element to indent them).

See https://github.com/kingargyle/wtp-sourceediting/blob/master/plugins/org.eclipse.wst.xml.core/src/org/eclipse/wst/xml/core/internal/formatter/DefaultXMLPartitionFormatter.java from WTP which uses this strategy.

It's a lot of work, but IMHO I think we should do that to have a clean XML formatter.

clementnuss:
fully agree with this issue, this is a "killer" for me at the moment, as I have to disable xml formatting altogether for the (many) documents where I use macros that insert credentials in xml fields.
these "macros" are not recognized by lemminx and replaced with `null` fields instead of being ignored.
commenting here to show the interest and to get notified of future improvements :)

angelozerr:
This issue should be fixed by the experimental formatter https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md#xmlformatexperimental

PR #1044, opened 2021-05-20
Provide basic experimental formatter which supports invalid XML

### LINKED ISSUE #1195: Provide basic experimental formatter which supports invalid XML
This issue is used to initialized the new formatter which uses a new strategy to format XML. Instead of rewriting the XML with proper indentation (which causes problem with invalid XML which looses some content) with one TextEdit, the main idea is to generate several TextEdit whith only spaces and line sperator (remove or add spaces before the start tag / end tag element).

This issue will provide a new format stetting "experimental : true" to use this new formatter.

This issue will provide a basic formatter and all existing settings will not take care of. We will improve it step by step by taking care of existing format settings.

angelozerr:
Provide basic experimental formatter which supports invalid XML

Fixes #1195


angelozerr:
This PR requires https://github.com/redhat-developer/vscode-xml/pull/703

To play with the new formatter you must sélect thé experimental settings.

The formatter is not perfect but it starts working. It has the same behavior than oxygen when text node is formatted.

I need to clean code and write some tests again but you can start planning with it.

Please note that range formatting is not supported for the moment and some settings are not supported again.

rgrunber:
I tried a bunch of snippets referenced from the issues in https://github.com/eclipse/lemminx/issues/1041#issue-895608360 , and they all seem to be fixed now. XML content/tags are not lost on a file that doesn't validate.

Is there a way to preserve spaces for all tags that only have text data and nothing else ? I'm thinking of  `<name>` & `<description>` tags of a pom file.

angelozerr:
> I tried a bunch of snippets referenced from the issues in https://github.com/eclipse/lemminx/issues/1041#issue-895608360 , and they all seem to be fixed now 🚀 . XML content/tags are not lost on a file that is not completely valid.

Great!

>Is there a way to preserve spaces for all tags that only have text data and nothing else (no children) ? I'm thinking of <name> & <description> tags of a pom file.

In this case, several spaces are replaced with one space. Is it annoying?

angelozerr:
> Is there a way to preserve spaces for all tags that only have text data and nothing else (no children) ? I'm thinking of & tags of a pom file.

Ok I tried with Oxygen, and it preserves spaces for description, name of pom.xml, because it uses the XSD information which have xs:string. When element is a xs;string, it means that spaces are preserved. 

I need to support that too.

angelozerr:
> Is there a way to preserve spaces for all tags that only have text data and nothing else (no children) ? I'm thinking of <name> & <description> tags of a pom file.

Now it should work (because description, version of pom.xml are xs:string). I introduce a new settings:

image

When element is bound with xs:string,  all spaces are preserved. You can try it and if you disable the grammar aware settings, preserve spaces should not work for description,version of pom.xml.

rgrunber:
It definitely works now. Initially it also didn't because I had `xml.downloadExternalResources.enabled` set to `false` and so the pom file could't download. It worked after change to `true`. However, what about a case where an XML file doesn't have an associated grammar whose type is `xs:string` ? Is their only option to add the element to `xml.preferences.preserveSpace` ?

angelozerr:
> xml.preferences.preserveSpace ?

Yes or you can write xml:space="preserved"

angelozerr:
This experimental formatter should start working and should have the same result than Oxygen. See https://www.oxygenxml.com/doc/versions/24.0/ug-editorEclipse/topics/format-and-indent-xml.html explanation with format element category.

This experimental formatter doesn't support all settings described in
https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md 

It supports those settings: 

 * https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md#xmlformatsplitattributes
 * https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md#xmlformatspacebeforeemptyclosetag
 * https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md#filesinsertfinalnewline
 * https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md#filestrimfinalnewlines
 * https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md#xmlformatsplitattributesindentsize

The main idea is to improve step by step this experimental formatter with the missing settings in another PRs.

datho7561:
Playing around, and it looks pretty good so far!

I found a case where content is still destroyed:

```xml
<asdf ""`=asdf />
```

Also, the result of formatting this is a bit odd, since it leaves a space within the `bbb` element:

```xml
<aaa>
  <bbb>
  </bbb>
<aaa>
```

However, I think that these are okay given the fact we are including this as an experimental formatter.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/TextEditUtils.java
```diff
@@ -0,0 +1,186 @@
...
+			}
+		} else {
+			return false;
+		}
+		return true;
+	}
+
+	public static String applyEdits(TextDocument document, List<? extends TextEdit> edits) throws BadLocationException {
+		String text = document.getText();
+		List<? extends TextEdit> sortedEdits = mergeSort(edits /* .map(getWellformedEdit) */, (a, b) -> {
```
datho7561, line 126: Is there any reason for using the custom merge sort instead of a built-in sorting function?

datho7561 (changes requested):

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/TextEditUtils.java
angelozerr: I use now Comparator, but please not ethis method is used only for tests.

angelozerr:
> Also, the result of formatting this is a bit odd, since it leaves a space within the bbb element:

It is the expexted behavior, see explanation at  https://github.com/redhat-developer/vscode-xml/pull/703/files#diff-db69e2b1fc0e8a5db9a104ac14f4102a40e9483a3f4df4987ca56f1fb95cb06aR43

datho7561:
When running the binary, I get "There is no formatter for xml installed", I think this might be related to restructuring some of the settings?

angelozerr:
@datho7561 I created my `own org.eclipse.lemminx.settings.LSPFormattingOptions` because the LSP4J can support String array, and I need it for preserveSpace which is an array of String.

Have you tried to add `org.eclipse.lemminx.settings.LSPFormattingOptions` in reflect-config.json?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/XMLFormatterDocumentNew.java
```diff
@@ -0,0 +1,559 @@
...
+		try {
+			return textDocument.lineDelimiter(0);
+		} catch (BadLocationException e) {
+			LOGGER.log(Level.SEVERE, e.getMessage(), e);
+		}
+		return System.lineSeparator();
+	}
+
+	/**
+	 * Returns a List containing a single TextEdit, containing the newly formatted
```
datho7561, line 112: It looks like sometimes multiple `TextEdit` are returned. Update the documentation to reflect this.

datho7561 (changes requested):
Once this comment is addressed I think it should be good to merge

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/XMLFormatterDocumentNew.java
angelozerr: good catch, fixed.

PR #1048, opened 2021-06-03
Add reflection config for CodeLens and References

### LINKED ISSUE #1046: CodeLens does not work in binary
Example to replicate:

```xml

  <!-- elements -->

  <xs:element name="root">
    <xs:complexType>
      <xs:sequence>
        <xs:element ref="container"/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>

  <xs:element name="container" type="asdf" />

  <xs:element name="content" type="asdf" />

  <!-- types -->

  <xs:complexType name="asdf">
    <xs:sequence>
      <xs:element ref="content"/>
    </xs:sequence>
  </xs:complexType>

</xs:schema>
```

There should be Code Lens before the `container` and `content` element declarations, and the `asdf` type declaration.

Instead, the response that's returned is:

```
[Trace - 4:43:33 PM] Received response 'textDocument/codeLens - (35)' in 6ms.
Result: [
    {},
    {},
    {}
]
```

### LINKED ISSUE #1047: Go To References does not work in binary
Here is an example DTD file

```dtd
<!ELEMENT aaa (bbb+)>
<!ELEMENT bb|b EMPTY>
```

If you are using the Java server, when you put your cursor where the pipe `|` is, then do `textDocument/references`, it should place your cursor on `bbb` from the previous line.

If you try this in the binary server, it freezes.

datho7561:
Fixes #1046, Fixes #1047

### DIFF
--- a/org.eclipse.lemminx/src/main/resources/META-INF/native-image/reflect-config.json
+++ b/org.eclipse.lemminx/src/main/resources/META-INF/native-image/reflect-config.json
@@ -485,6 +485,14 @@
 		}]
 	},
 	{
+		"name": "org.eclipse.lsp4j.CodeLens",
+		"allDeclaredFields": true,
+		"methods": [{
+			"name": "<init>",
+			"parameterTypes": []
+		}]
+	},
+	{
 		"name": "org.eclipse.lsp4j.CodeLensCapabilities",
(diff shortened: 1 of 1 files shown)


angelozerr:
Thanks @datho7561 !

PR #1059, opened 2021-06-17
Fix Go To References in binary

datho7561:
Closes redhat-develoepr/vscode-xml#513


rgrunber (approved):
Fixes the issue for me. Also noted that selecting the codelens entry itself did work before.

PR #1070, opened 2021-06-25
Add support for exiting immediately from a shutdown() request.

rgrunber:
LemMinX (at least for those adopting vscode-languageclient 7.x) will not exit immediately after the client shuts down, and relies on the parent process watcher, which can take some time.

- In languageclient 7.x, the client fails to send the necessary exit()
  once a shutdown() response is received from the language server
- When client defines shouldLanguageServerExitOnShutdown as true, the
  language server will exit immediately after the shutdown request

The delay ensures the shutdown() response will be received by the client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## [0.17.1](https://github.com/eclipse/lemminx/milestone/25?closed=1) (June 25, 2021)
+
+### Bug Fixes
+
+ * Add support for exiting immediately from a shutdown() request. See [#1070](https://github.com/eclipse/lemminx/pull/1070).
+
 ## [0.17.0](https://github.com/eclipse/lemminx/milestone/23?closed=1) (June 22, 2021)
 
 ### Enhancements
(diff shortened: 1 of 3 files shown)


rgrunber:
@datho7561

datho7561 (changes requested):
Add a changelog, since we will need to do a 0.17.1 release for this

datho7561 (approved):

angelozerr:
I have not tested but after debugging vscode languageclient 7.0.0 , the problem occured in shutdown process: It called:

 - connection.exit()
 - connection.end() without waiting for exit process.

I have not tested, but I think this commit https://github.com/microsoft/vscode-languageserver-node/commit/6eef6d21ec50399a67d80442c624344870f683c8#diff-1442f78812d67be25f18d3b5d82e8b4191ae2f939a900eac32d4360222c5d845R3349 should fix the issue.

IMHO, once https://github.com/microsoft/vscode-languageserver-node/commit/6eef6d21ec50399a67d80442c624344870f683c8#diff-1442f78812d67be25f18d3b5d82e8b4191ae2f939a900eac32d4360222c5d845R3349 will be released, we should test it and remove the `shouldLanguageServerExitOnShutdown ` capability

Anyway thanks @rgrunber for your patch!

PR #1074, opened 2021-06-30
fix stackoverflow and ressource leak in folding

GitMensch:
The old code used `ArrayList.subList()` which is only an adjusted view on the original list:
* likely to StackOverflow when called multitude of times in this recursive subList->subList->subList case
* resource hungry because:
   * the old ArrayList is still there and the view has a reference on it
   * the stackElement items (Tags) are kept until the loop is left because they are still referenced there

The new code removes the trailing elements directly, so the references are gone and the garbage collector is free to do its work; and most important: there's only one List, not one for each close tag/comment.

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
@@ -99,7 +99,11 @@ class XMLFoldings {
 					}
 					if (i >= 0) {
 						TagInfo stackElement = stack.get(i);
-						stack = stack.subList(0, i); // stack.length = i;
+						// remove obsolete entries ()
+						int j = stack.size() - 1;
+						while (j >= i) {
+							stack.remove(j--);
+						}
 						int line = document.positionAt(scanner.getTokenOffset()).getLine();
 						int startLine = stackElement.startLine;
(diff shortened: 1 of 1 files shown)


angelozerr:
It works like a charm, thanks so much @GitMensch ! The main reason why I used subList and it was because I translated the HTML foldings typescript file into Java (in ts they used array.slice) quickly.

Your fix is very nice. thanks!

GitMensch:
I'm lucky to be of help :-) especially as I use lemminx via the redhat XML plugin.
Just for a brought understanding: while it is part of the 0.18.0 milestone it looks like a near 0.17.2 is likely, correct?

Do you have any experience how long it takes for the redhat XML plugin to do a release with an updated lemminx included (or is that maybe updated from within the installed extensions)?

angelozerr:
> I'm lucky to be of help :-) especially as I use lemminx via the redhat XML plugin.

That's great! And your patch is very relevant and I hope it will fix some memory trouble.

> Do you have any experience how long it takes for the redhat XML plugin to do a release with an updated lemminx included

We do a release when we have a lot of interesting features or if we find an important bug. For the moment the release is not planned.

> (or is that maybe updated from within the installed extensions)?

Please read https://github.com/redhat-developer/vscode-xml#contributing you will find a link to download the last vsix built by last commit of lemminx or vscode-xml.

PR #1082, opened 2021-07-11
follow org.w3c.dom.Element API in DOMElement

### LINKED ISSUE #1079: DOMElement seems to not fully match the interface org.w3c.dom.Element
`getAttributeNS()` should either return a value or `""` if the value is not set
`getElementsByTagName()` and `getElementsByTagNameNS` should always return a `NodeList`, if nothing matches an empty one (the later one is explicit that it must create a _new_ `Nodelist`).

The namespace variants specify that they may throw a  `new DOMException(NOT_SUPPORTED_ERR, msg)`.

Question: Should I PR

```diff
-		return null;
+		throw new DOMException(DOMException.NOT_SUPPORTED_ERR, "no namespace support");
```

for those (possibly only if a namespace is set) and possibly return a `new Nodelist()` for `getElementsByTagName()` or something else - or just add a comment "we don't care for the interface definition here, this is never called" (and add a RuntimeException there)?

GitMensch:
fixes #1079

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMElement.java
@@ -34,7 +34,7 @@ public class DOMElement extends DOMNode implements org.w3c.dom.Element {
 
 	String tag;
 	boolean selfClosed;
-
+	
 	// DomElement.start == startTagOpenOffset
 	int startTagOpenOffset = NULL_VALUE; // |<root>
 	int startTagCloseOffset = NULL_VALUE; // <root |>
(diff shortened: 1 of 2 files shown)


GitMensch:
The current failing build (JRE 1.8 compat) is in an area still open (see questions in related issue), so I'll leave it as-is for now until @angelozerr defined what we want to do. That implementation will then use JRE 1.8 compat (I've just toggled the local project settings to ensure I won't use JRE 11 here).

angelozerr:
The PR looks promising. 

It should be really nice if you could write some tests to check your new implementations will continue to work if we do some refactoring. I suggest that you see https://github.com/eclipse/lemminx/blob/0007ad45e1d89480bdb7052b75c9d79ef2ce9358/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/dom/DOMDocumentTest.java#L86 and add  new tests in this file. Thanks!

PR #1083, opened 2021-07-12
Do not attempt to auto close tags in attribute values

datho7561:
Fixes redhat-developer/vscode-xml#547


rgrunber (approved):
Works for me. Feel free to merge.

Issue #1069, opened 2021-06-24
Incorrect error range for `src-import.3.2`


datho7561:
Given a schema:

```xml
<xs:schema
    xmlns:xs="http://www.w3.org/2001/XMLSchema"
    targetNamespace="http://example.org/my-example"
    xmlns:NS="http://example.org/my-example">
  <xs:import schemaLocation="./otherSchema.xsd"/>
</xs:schema>
```

and given that the schema file `otherSchema.xsd` exists and contains:

```xml
<xs:schema
    xmlns:xs="http://www.w3.org/2001/XMLSchema"
    targetNamespace="http://foo.bar"
>
  <xs:element name="root-element"></xs:element>
</xs:schema>
```

an error with the code `xsd(src-import.3.2)` is generated somewhere in the the starting tag of `<xs:schema>` in the first file. It would be nice if this error range instead covered the name of the `<xs:import>` element, or maybe even the value of the `schemaLocation` attribute of the `<xs:import>` element.

AlexXuChen:
Hello

PR #1086, opened 2021-07-15
fixed `src-import.3.1` xsd error code highlighting

### LINKED ISSUE #1075: src-import.3.1 Highlighting for Namespace Attribute
Given the following snippet, assuming `foo.xsd` exists in the same directory:
```xsd
<xs:schema xmlns:xs='http://www.w3.org/2001/XMLSchema'>
	<xs:import namespace='BAD_NAMESPACE' schemaLocation='foo.xsd' />
	<xs:element name='foo'></xs:element>
</xs:schema>
```
the following behaviour will occur:
Screenshot from 2021-07-02 09-35-27
which throws `src-import.3.1: The namespace attribute, 'BAD_NAMESPACE', of an <import> element information item must be identical to the targetNamespace attribute, 'null', of the imported document.`

It would be better if the `namespace` attribute value was highlighted instead of the current default.

AlexXuChen:
Fixed `src-import.3.1` XSD error code highlighting to highlight `namespace` attribute value.

Closes #1075 

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/participants/XSDErrorCode.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/participants/XSDErrorCode.java
@@ -34,6 +34,9 @@ import org.eclipse.lsp4j.Range;
  *
  * @see https://wiki.xmldation.com/Support/Validator
  *
+ * All error code types and messages can be found in the Xerces library
+ * https://github.com/apache/xerces2-j/blob/trunk/src/org/apache/xerces/impl/msg/XMLSchemaMessages.properties
+ *
  */
 public enum XSDErrorCode implements IXMLErrorCode {
 
@@ -56,7 +59,8 @@ public enum XSDErrorCode implements IXMLErrorCode {
 	src_element_3("src-element.3"),
 	src_resolve_4_2("src-resolve.4.2"), //
 	src_resolve("src-resolve"), src_element_2_1("src-element.2.1"),
-	EmptyTargetNamespace("EmptyTargetNamespace");
+	EmptyTargetNamespace("EmptyTargetNamespace"),
+	src_import_3_1("src-import.3.1");
 
 	private final String code;
 
@@ -163,7 +167,18 @@ public enum XSDErrorCode implements IXMLErrorCode {
 		}
 		case EmptyTargetNamespace :
 			return XMLPositionUtility.selectAttributeValueAt(XSDUtils.TARGET_NAMESPACE_ATTR, offset, document);
+		case src_import_3_1: {
+			// If the imported file of `schemaLocation` contains at least a doctype and 'xs:schema' with at least one `xs:element`,
+			// then the `xs:import` line will be highlighted, otherwise the `xs:schema` line will be highlighted
+			DOMNode elementHighlighted = document.findNodeAt(offset);
+			if (elementHighlighted.getNodeName().equals(XSDUtils.XS_SCHEMA_TAG)) { // `xs:schema` line is highlighted
+				return XMLPositionUtility.selectChildNodeAttributeValueFromGivenNameAt(XSDUtils.XS_IMPORT_TAG, XSDUtils.NAMESPACE_ATTR, offset, document);
+			} else { // `xs:import` line is highlighted
+				return XMLPositionUtility.selectAttributeValueAt(XSDUtils.NAMESPACE_ATTR, offset, document);
+			}
 		}
+		}
+
 		return null;
 	}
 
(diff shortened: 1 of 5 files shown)


datho7561:
If I use your example, and add content into `foo.xsd`:

```xml
	<xs:element name="root-element">
		
	</xs:element>
</xs:schema>
```

Then `"foo.xsd"` is highlighted in the `xs:import` instead of `BAD_NAMESPACE`

AlexXuChen:
> Then `"foo.xsd"` is highlighted in the `xs:import` instead of `BAD_NAMESPACE`

I believe I ran into this originally (the example in #1086), and the initial fix was using `selectAttributeValueAt`. However, with an empty `foo.xsd`, the highlight occurs in the tag at the top.

I believe the fix is to separate into subcases

rgrunber (approved):
Change looks good to me, and fixes the  issue where the referenced schema is empty.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/participants/XSDErrorCode.java
```diff
@@ -163,7 +164,18 @@ public static Range toLSPRange(XMLLocator location, XSDErrorCode code, Object[]
 		}
 		case EmptyTargetNamespace :
 			return XMLPositionUtility.selectAttributeValueAt(XSDUtils.TARGET_NAMESPACE_ATTR, offset, document);
+		case src_import_3_1: { // src-import.3.1: The namespace attribute, _, of an <import> element information item must be identical to the targetNamespace attribute, _, of the imported document.
```
rgrunber, line 167: Would it be easier to reference something like http://svn.apache.org/viewvc/xerces/java/trunk/src/org/apache/xerces/impl/msg/XMLSchemaMessages.properties?view=log as opposed to copying the full error message ?

Is there a way to reference that file ?

AlexXuChen: I added the comment just so we know what error is thrown, but I guess it's inconsistent with the other cases, so I can remove it.

However, I feel it is helpful to have somewhere that lists the message(s) thrown so that (in the future) a developer knows where to look specifically when handling this error.

rgrunber: I would maybe  add http://svn.apache.org/viewvc/xerces/java/trunk/src/org/apache/xerces/impl/msg/XMLSchemaMessages.properties?revision=1869198&view to the top of the file. I think that's the upstream resource.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/xsd/XSDValidationExtensionsTest.java
```diff
@@ -379,6 +379,18 @@ public void schema_reference_4_IncludeErrorRange() throws BadLocationException {
 		testDiagnosticsFor(xsd, d);
 	}
 
+	@Test
+	public void src_import_3_1_BadNamespace() throws BadLocationException {
```
datho7561, line 383: Add another test that covers the case you described in XSDErrorCode.java

datho7561 (changes requested):
Looks pretty good.

This will always select the first `<xs:import>` if there are multiple `<xs:import>, even if the first one isn't the one generating the error. However, I don't think there is a way around this given the Xerces bug. I am good with this implementation of the fix.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/participants/XSDErrorCode.java
```diff
@@ -34,6 +34,9 @@
  *
  * @see https://wiki.xmldation.com/Support/Validator
  *
+ * All error code types and messages can be found in the Xerces library
+ * http://svn.apache.org/viewvc/xerces/java/trunk/src/org/apache/xerces/impl/msg/XMLSchemaMessages.properties?revision=1869198&view
```
angelozerr, line 38: I think it's better to use github link https://github.com/apache/xerces2-j/blob/trunk/src/org/apache/xerces/impl/msg/XMLSchemaMessages.properties because you can nanvigate easily for another sources.

rgrunber: Agreed, though I noticed the GitHub link doesn't seem as up to date as the SVN one. I'd double check that just to be sure updates still flow through there.

datho7561 (approved):
LGTM

AlexXuChen:
Related to #1069

PR #1087, opened 2021-07-20
fixed `src-import.3.2` XSD error highlighting support

### LINKED ISSUE #1069: Incorrect error range for `src-import.3.2`

AlexXuChen:
Fixed the `src-import.3.2` XSD error highlighting support to highlight the `xs:import` `schemaLocation` attribute.

Closes #1069 


datho7561 (approved):
LGTM

PR #1088, opened 2021-07-22
added CodeAction for opening XML binding wizard

AlexXuChen:
Added CodeAction for opening XML binding wizard to bind document to existing grammar/schema file.
(edit) made binding wizard CodeAction dependent on client support (reference: https://github.com/eclipse/lemminx/pull/1088#pullrequestreview-713931073)

Closes redhat-developer/vscode-xml/issues/515

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/CodeActionFactory.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/CodeActionFactory.java
@@ -20,6 +20,7 @@ import java.util.List;
 
 import org.eclipse.lsp4j.CodeAction;
 import org.eclipse.lsp4j.CodeActionKind;
+import org.eclipse.lsp4j.Command;
 import org.eclipse.lsp4j.CreateFile;
 import org.eclipse.lsp4j.CreateFileOptions;
 import org.eclipse.lsp4j.Diagnostic;
@@ -41,7 +42,7 @@ public class CodeActionFactory {
 
 	/**
 	 * Create a CodeAction to remove the content from the given range.
-	 * 
+	 *
 	 * @param title
 	 * @param range
 	 * @param document
@@ -54,13 +55,13 @@ public class CodeActionFactory {
 
 	/**
 	 * Create a CodeAction to insert a new content at the end of the given range.
-	 * 
+	 *
 	 * @param title
 	 * @param range
 	 * @param insertText
 	 * @param document
 	 * @param diagnostic
-	 * 
+	 *
 	 * @return the CodeAction to insert a new content at the end of the given range.
 	 */
 	public static CodeAction insert(String title, Position position, String insertText, TextDocumentItem document,
@@ -77,11 +78,11 @@ public class CodeActionFactory {
 
 	/**
 	 * Returns the text edit to insert a new content at the end of the given range.
-	 * 
+	 *
 	 * @param insertText text to insert.
 	 * @param position   the position.
 	 * @param document   the text document.
-	 * 
+	 *
 	 * @return the text edit to insert a new content at the end of the given range.
 	 */
 	public static TextDocumentEdit insertEdit(String insertText, Position position, TextDocumentItem document) {
@@ -142,7 +143,7 @@ public class CodeActionFactory {
 
 	/**
 	 * Makes a CodeAction to create a file and add content to the file.
-	 * 
+	 *
 	 * @param title      The displayed name of the CodeAction
 	 * @param docURI     The file to create
 	 * @param content    The text to put into the newly created document.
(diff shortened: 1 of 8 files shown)


datho7561:
Take a look into the test failures; I think they shouldn't be too hard to resolve

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/CodeActionFactory.java
```diff
@@ -169,4 +170,22 @@ public static CodeAction createFile(String title, String docURI, String content,
...
+
+	/**
+	 * Makes a CodeAction to call a command from the available server commands.
+	 *
+	 * @param title       The displayed name of the CodeAction
+	 * @param documentURI The document URI of the document the command is called on
+	 * @param commandId   The id of the given command to add as CodeAction
+	 * @param diagnostic  The diagnostic that this CodeAction will fix
+	 */
+	public static CodeAction createCommand(String title, String documentURI, String commandId, Diagnostic diagnostic) {
```
datho7561, line 182: I think it would be better to make this method more generic, so that it could be reused in the future if need to make another CodeAction that calls a command. For instance, I think this should:
 * Accept an array or a `List` of `Object`s to pass to the command instead of `documentURI`
 * Use the title for the title of the Command as well, instead of a fixed string.
 * Name the `Command` object something less specific

datho7561 (changes requested):

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/CodeActionFactory.java
angelozerr: The maini idea with this fracture is that it should be generic and not to have custom code.
Indeed i would like to provide a jar library with this  class and any lsp language server  could use it.

datho7561 (changes requested):
The changes that you have made look good, and it works for me.

I think we may want to add a new "extended client capability" for the binding wizard, and hide the `CodeAction` when the binding wizard is not supported by the client. That way, clients that have implemented a binding wizard (such as vscode-xml) can send this information to the server, and the `CodeAction` is listed. However, if the client hasn't implemented a binding wizard yet (i.e. Eclipse, vim, ...), then the `CodeAction` doesn't appear. This should prevent end users from getting confusing messages when trying to use the `CodeAction` when it's not supported.

Take a look at `ExtendedClientCapabilities.java`, and for an example of how to pass the setting to the language server in the client, see `xmlClient.ts:122` in vscode-xml.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/CodeActionFactory.java
```diff
@@ -169,4 +170,22 @@ public static CodeAction createFile(String title, String docURI, String content,
...
+
+	/**
+	 * Makes a CodeAction to call a command from the available server commands.
+	 *
+	 * @param title         The displayed name of the CodeAction
+	 * @param commandParams The document URI of the document the command is called on
+	 * @param commandId     The id of the given command to add as CodeAction
+	 * @param diagnostic    The diagnostic that this CodeAction will fix
+	 */
+	public static CodeAction createCommand(String title, List<Object> commandParams, String commandId, Diagnostic diagnostic) {
```
angelozerr, line 182: To be consistent with Command constructor, move `String commandId` before `List<Object> commandParams`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/SharedSettings.java
```diff
@@ -131,11 +133,29 @@ public boolean isOpenSettingsCommandSupport() {
...
 	public void setOpenSettingsCommandSupport(boolean openSettingsCommandSupport) {
 		this.openSettingsCommandSupport = openSettingsCommandSupport;
 	}
 
+	/**
+	 * Returns true if the client supports the `xml.open.binding.wizard` command using dropdown and false otherwise
+	 *
+	 * @return bindingWizardSupport
+	 */
+	public boolean getBindingWizardSupport() {
```
angelozerr, line 148: Rename `getBindingWizardSupport` with `isBindingWizardSupport`

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/XMLAssert.java
```diff
@@ -636,6 +635,19 @@ public static CodeAction ca(Diagnostic d, TextEdit... te) {
 		return codeAction;
 	}
 
+	/**
+	 * Mock code action for creating a command code action
+	 */
+	public static CodeAction cac(Diagnostic d, Command c) {
```
angelozerr, line 641: rename to `ca` to be consistent with other `ca` methods.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/NoGrammarConstraintsCodeAction.java
```diff
@@ -101,17 +103,26 @@ public void doCodeAction(Diagnostic diagnostic, Range range, DOMDocument documen
 					diagnostic);
 			codeActions.add(dtdWithXmlModelAction);
 
+			// ---------- Open Binding Wizard
+			if (sharedSettings.getBindingWizardSupport()) {
+				String documentURI = document.getDocumentURI();
+				String title = "Open binding wizard to bind existing grammar/schema";
```
angelozerr, line 109: For title I will use the same than Codelens that's to say `Bind to grammar/schema...`

AlexXuChen: Do you think the "existing" part is more descriptive?

angelozerr: yes

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/NoGrammarConstraintsCodeAction.java
```diff
@@ -101,17 +103,26 @@ public void doCodeAction(Diagnostic diagnostic, Range range, DOMDocument documen
 					diagnostic);
 			codeActions.add(dtdWithXmlModelAction);
 
+			// ---------- Open Binding Wizard
+			if (sharedSettings.isBindingWizardSupport()) {
+				String documentURI = document.getDocumentURI();
+				String title = "Bind to existing grammar/schema";
+				List<Object> commandParams = Arrays.asList(new Object[]{documentURI});
```
angelozerr, line 110: You should write:

List<Object> commandParams = Arrays.asList(documentURI);

angelozerr:
Good job @AlexXuChen

PR #1093, opened 2021-07-29
Codelens for referenced grammars

### LINKED ISSUE #1092: Codelens for referenced grammars
When a XML is bound with grammars, it should be niceto display a Codelens per referenced grammar. It can be helpfull when catalog is used.

This idea is taken from Yaml Language Server.

To implement this idea the create of Codelens must be done in existing ContentModelCodeLensParticipant and ReferencedGrammarInfo must be used to know the referenced grammars.

angelozerr:
Codelens for referenced grammars

Fixes #1092

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCodeLensParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCodeLensParticipant.java
@@ -13,14 +13,20 @@
 package org.eclipse.lemminx.extensions.contentmodel.participants;
 
 import static org.eclipse.lemminx.client.ClientCommands.OPEN_BINDING_WIZARD;
+import static org.eclipse.lemminx.client.ClientCommands.OPEN_URI;
+import static org.eclipse.lemminx.extensions.contentmodel.commands.CheckBoundGrammarCommand.canBindWithGrammar;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Set;
 
 import org.eclipse.lemminx.client.CodeLensKind;
 import org.eclipse.lemminx.dom.DOMDocument;
-import org.eclipse.lemminx.dom.DOMElement;
+import org.eclipse.lemminx.dom.DOMRange;
 import org.eclipse.lemminx.extensions.contentmodel.commands.AssociateGrammarCommand;
+import org.eclipse.lemminx.extensions.contentmodel.model.ContentModelManager;
+import org.eclipse.lemminx.extensions.contentmodel.model.ContentModelProvider.Identifier;
+import org.eclipse.lemminx.extensions.contentmodel.model.ReferencedGrammarInfo;
 import org.eclipse.lemminx.services.extensions.codelens.ICodeLensParticipant;
 import org.eclipse.lemminx.services.extensions.codelens.ICodeLensRequest;
 import org.eclipse.lemminx.utils.DOMUtils;
@@ -35,13 +41,13 @@ import org.eclipse.lsp4j.jsonrpc.CancelChecker;
  * At first this participant is enabled only when LSP client can support the
  * client command "xml.open.binding.wizard" to open the wizard to bind a XML to
  * a grammar/schema.
- * 
+ *
  * In this case, when XML file is not associated to a grammar/schema (DTD, XSD),
  * this class generates [Bind to grammar/schema...] CodeLens on the root of the
  * DOM Document:
- * 
+ *
  * On client side, click on this Codelens should open a wizard:
- * 
+ *
  * <ul>
  * <li>page1 : display a combo to select the binding type ("standard",
  * "xml-model").</li>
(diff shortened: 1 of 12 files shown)


angelozerr:
This PR requires https://github.com/redhat-developer/vscode-xml/pull/569

angelozerr:
@datho7561 you can start playing with my PR.

Here the result:

image

I nned to clean my code and write tests.

angelozerr:
> Works pretty good so far! Some NPEs propagate to the logs, but you'll likely catch that in testing.

Have you some trace log with NPE?

> I noticed when opening a schema that uses a URL, the URL is opened in my browser. Do you think it's worth opening the locally cached version of the remote schema (maybe in read only mode)? If it makes sense, but makes the PR way more complex, we could always do this in the future.

I wanted to do that but to be honnest with you I did the PR quickly, just to have some feedback, let's me try to fix that.

I wonder if those codelens should be displayed on the top of the document(before for instance <?xml processing instruction to be more visible. What do you think about that?

datho7561:
> Have you some trace log with NPE?

Here is a stacktrace:

```
Message: Error while processing code lens for the participant 'org.eclipse.lemminx.extensions.contentmodel.participants.ContentModelCodeLensParticipant'.
java.lang.NullPointerException
	at org.eclipse.lemminx.extensions.contentmodel.participants.ContentModelCodeLensParticipant.createReferencedGrammarLens(ContentModelCodeLensParticipant.java:96)
	at org.eclipse.lemminx.extensions.contentmodel.participants.ContentModelCodeLensParticipant.createReferencedGrammarLenses(ContentModelCodeLensParticipant.java:91)
	at org.eclipse.lemminx.extensions.contentmodel.participants.ContentModelCodeLensParticipant.doCodeLens(ContentModelCodeLensParticipant.java:78)
	at org.eclipse.lemminx.services.XMLCodeLens.getCodelens(XMLCodeLens.java:48)
	at org.eclipse.lemminx.services.XMLLanguageService.getCodeLens(XMLLanguageService.java:264)
	at org.eclipse.lemminx.XMLTextDocumentService.lambda$codeLens$18(XMLTextDocumentService.java:429)
	at java.base/java.util.concurrent.CompletableFuture.biApply(CompletableFuture.java:1236)
	at java.base/java.util.concurrent.CompletableFuture$BiApply.tryFire(CompletableFuture.java:1205)
	at java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:479)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
```

> I wonder if those codelens should be displayed on the top of the document(before for instance <?xml processing instruction to be more visible. What do you think about that?

I think this is a good idea, because I guess you could do:

```xml
<!DOCTYPE root SYSTEM "dtd-npe-example.dtd">
```
... as a document, and then you can't put the code lens on the root element since there is none.

angelozerr:
> Here is a stacktrace:

It should be fixed, please give me feedback.

> I think this is a good idea, because I guess you could do:

done, the codelens should appears on the top of the document

image

> Do you think it's worth opening the locally cached version of the remote schema (maybe in read only mode)?

Should work now. I don't know how to manage readonly mode yet.

Please tell me if it works for you thanks!

Tomorrow I will clean my code and write tests.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCodeLensParticipant.java
```diff
@@ -62,9 +67,60 @@
...
 	public void doCodeLens(ICodeLensRequest request, List<CodeLens> lenses, CancelChecker cancelChecker) {
-		if (!canSupport(request)) {
+		// List of referenced schema/grammar
+		createReferencedGrammarLenses(request, lenses);
+		// "Bind to grammar/schema..."
+		createBindToGrammarSchemaLenses(request, lenses);
+	}
+
+	private void createReferencedGrammarLenses(ICodeLensRequest request, List<CodeLens> lenses) {
+		if (!canSupportOpenUri(request)) {
```
datho7561, line 85: It might be interesting to create a CodeLens with the path to the schema that does nothing when clicked if `OpenUri` is not supported. What do you think?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCodeLensParticipant.java
```diff
@@ -62,9 +67,60 @@
...
+			return;
+		}
+		
+		DOMDocument document = request.getDocument();
+		if (!document.hasGrammar()) {
+			return;
+		}
+		Range range = XMLPositionUtility.createRange(document);
```
datho7561, line 93: Seems like the range for a CodeLens is expected to only span one line (See the definition of the `CodeLens` interface): https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeLens

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCodeLensParticipant.java
```diff
@@ -62,9 +67,60 @@
...
+		Identifier identifier = info.getIdentifier();
+		String publicId = identifier.getPublicId();
+		String systemId = identifier.getSystemId();
+		String uri = publicId != null ? publicId : systemId;
+		String title = uri + " (" + identifier.getKind() + ")";
+		String grammarURI = info.isInCache() ? info.getGrammarCacheInfo().getCachedResolvedUri() : null;
+		if (grammarURI == null) {
+			grammarURI = info.getResolvedURIInfo().getResolvedURI();
+		}
+		Command command = new Command(title, OPEN_URI, Arrays.asList(grammarURI));
```
datho7561, line 118: It might be nice to pass a argument to the command to tell the client whether the document should be opened in read-only mode. If the document is in the cache, if would be nice to prevent the user from writing to it. If this gets too complex, feel free to ignore.

datho7561 (changes requested):

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCodeLensParticipant.java
angelozerr: Yes you are right (ex : on WWD from Eclipse IDE). supported now.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCodeLensParticipant.java
angelozerr: To be consistent with hyperlink (ex : from  xsi:noNamepaceSchemaLaocation value) I would like to avoid opening the xsd in read-only mode.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCodeLensParticipant.java
angelozerr: Fixed, now the start and the end position have the same value.

datho7561:
I get:

```
Message: Error while processing code lens for the participant 'org.eclipse.lemminx.extensions.contentmodel.participants.ContentModelCodeLensParticipant'.
java.lang.NullPointerException
	at org.eclipse.lemminx.extensions.contentmodel.participants.ContentModelCodeLensParticipant.createReferencedGrammarLens(ContentModelCodeLensParticipant.java:111)
	at org.eclipse.lemminx.extensions.contentmodel.participants.ContentModelCodeLensParticipant.createReferencedGrammarLenses(ContentModelCodeLensParticipant.java:97)
	at org.eclipse.lemminx.extensions.contentmodel.participants.ContentModelCodeLensParticipant.doCodeLens(ContentModelCodeLensParticipant.java:80)
	at org.eclipse.lemminx.services.XMLCodeLens.getCodelens(XMLCodeLens.java:48)
	at org.eclipse.lemminx.services.XMLLanguageService.getCodeLens(XMLLanguageService.java:264)
	at org.eclipse.lemminx.XMLTextDocumentService.lambda$codeLens$19(XMLTextDocumentService.java:466)
	at java.base/java.util.concurrent.CompletableFuture.biApply(CompletableFuture.java:1236)
	at java.base/java.util.concurrent.CompletableFuture$BiApply.tryFire(CompletableFuture.java:1205)
	at java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:479)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
```

To reproduce the error, I open the following XSD:

```xml
<xs:schema
    >
  <xs:element name="root">
    <xs:complexType>
      <xs:sequence>
        <xs:element ref="child" minOccurs="0" maxOccurs="unbounded" />
      </xs:sequence>
    </xs:complexType>
  </xs:element>
  <xs:element name="child">
    <xs:complexType>
      <xs:attribute name="attr"></xs:attribute>
    </xs:complexType>
  </xs:element>
</xs:schema>
```

I think its related to https://github.com/eclipse/lemminx/pull/1093/files#diff-48bfd6def2521586629b68d4a472a8a61220079f472dfad06dee03c05e500680R106; I think it will try to generate a code lens in XSD/DTD

angelozerr:
> To reproduce the error, I open the following XSD:

It should be fixed. Now if you open a XSD you should see the same referenced grammar that you see in the outline.

datho7561 (approved):

datho7561:
Wait; tests are failing

Issue #1066, opened 2021-06-23
Send telemetry event when opening a file


fbricon:
When a file is opened, a telemetry event should be sent with
- file extension (.xml, .xsd, .classpath, ...)
- associated grammar:
      * none/xsd/dtd
      * local/remote
      * schema/grammar URL, if remote

we should be able to understand the popular content. If some schemas are used often that would give us some guidance on priorities.

rgrunber:
Entry point would probably be https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java#L531 . From there, one can access the DOMDocument of the document being opened. It already has helper methods

angelozerr:
@mauri-sfdc why are you confused? As @fbricon said, our goal is to give us some guidance on priorities. For instance if we see that :

 * there are a lot of xsl file which is opened, we could improve our basic XSL support.
 * there are a lot of dtd file which is opened, we could improve syntax coloration for DTD and improve DTD support again.
 * there are a lot of XML bound with RelaxNG files, we could start to implement RelaxNG support.

The telemetry will give us those relevant informations.

angelozerr:
To know the referenced grammars, we should get list of [ReferencedGrammarInfo](https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/model/ReferencedGrammarInfo.java) from the opened, closed, etc DOM document.

```java
Set<ReferencedGrammarInfo> referencedGrammarInfos = contentModelManager.getReferencedGrammarInfos(document);
```

See sample at https://github.com/eclipse/lemminx/blob/81298daf1d6aab4a5fd06fab3675affbf8fbff5d/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelSymbolsProviderParticipant.java#L73

You can see content of list of ReferencedGrammarInfo in Outline. With list of ReferencedGrammarInfo you will have know the list of bound XSD, DTD and for each binding you will know the binding type (catalog, files associations, xsi, doctype, etc) I think binding type is a relevant information too.

[...]

PR #1094, opened 2021-07-30
Report telemetry events for text document services

rgrunber:
- ~This change is currently based off of #1091 so once that gets merged, we'll need to rebase against that.~
- ~Need to update USAGE_DATA.md~
- ~Need to review the naming of certain events. Eg. "extension" is a pretty overloaded term.~
- ~Currently only records document file extension, but grammar/schema can also be done.~

### DIFF
--- a/USAGE_DATA.md
+++ b/USAGE_DATA.md
@@ -13,6 +13,11 @@ When telemetry events are enabled, the following information is emitted when the
     * The free, total, and max VM memory
  * Version information:
     * The server version number
+ * Text Document Information
+   * When a document is opened :
+      * The file extension (eg. `xml`, `xsd`, `dtd`)
+      * The associated grammar types (eg. `none`, `doctype`, `xml-model`, `xsi:schemaLocation`, `xsi:noNamespaceSchemaLocation`)
+      * The resolver used to resolve the grammar identifier (eg. `catalog`, `file association`, `embedded catalog.xsd`, `embedded xml.xsd`, `embedded xslt.xsd`)
  * Note: Does NOT include the `JAVA_HOME` environment variable for privacy reasons
 
 Currently, the startup event is the only telemetry event that is emitted.
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/telemetry/DocumentTelemetryInfo.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/telemetry/DocumentTelemetryInfo.java
@@ -1,0 +1,77 @@
+/*******************************************************************************
+* Copyright (c) 2021 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.telemetry;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.extensions.contentmodel.model.ContentModelManager;
+import org.eclipse.lemminx.extensions.contentmodel.model.ReferencedGrammarInfo;
+
+public class DocumentTelemetryInfo {
+
+	private static final String DOC_PROP_EXT = "file.extension";
+	private static final String DOC_PROP_RESOLVER = "file.resolver";
+	private static final String DOC_PROP_GRAMMAR_NONE = "file.grammar.none";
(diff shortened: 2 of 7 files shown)


angelozerr:
The PR looks promising. You can rebase your PR.

> but grammar/schema can also be done.

please see https://github.com/eclipse/lemminx/issues/1066#issuecomment-888953831

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/telemetry/DocumentTelemetryInfo.java
```diff
@@ -0,0 +1,54 @@
...
+		HashMap<String, Object> props = new HashMap<>();
+		props.put(DOC_PROP_EXT, fileExtension);
+
+		if (referencedGrammarInfos.isEmpty()) {
+			props.put(DOC_PROP_GRAMMAR, "none");
+		} else {
+			List<String> values = new ArrayList<String>();
+			for (ReferencedGrammarInfo info : referencedGrammarInfos) {
+				String kind = info.getIdentifier().getKind();
+				if (kind == null) {
```
rgrunber, line 44: For some files (eg. lemminx/org.eclipse.lemminx/src/test/resources/xsd/camel-spring-2.22.0.xsd), the kind is null, but the public id can be something like "http://www.w3.org/2001/XMLSchema". Does using this make sense ?

angelozerr:
@rgrunber please review my PR https://github.com/eclipse/lemminx/pull/1098 and use it in your PT. After that you could get 3 informations:

 * info.getIdentifierURI() : the public or system identifier URI
 * info.getBindingKind() : the declared binding kinf in the xml (xml-model,xsi:schemaLocation, doctype, or null)
 * info.getResolvedBy() : the mechanism which resolves the identifier URI (ex : catalog,file association, embedded xml.xsd, etc)

The telemetry should report those 3 relevant informations. For instance if we see that info.getResolvedBy() returns a lot of "catalog", it means that we have a lot of user which uses XML catalog.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/telemetry/DocumentTelemetryInfo.java
angelozerr: Indeed it was not very clear, that's why I created the PR https://github.com/eclipse/lemminx/pull/1098 to simplify the API.

> For some files (eg. lemminx/org.eclipse.lemminx/src/test/resources/xsd/camel-spring-2.22.0.xsd or lemminx/org.eclipse.lemminx/src/test/resources/catalogs/catalog.xml), the kind is null

Yes it's normal, the kind is null when you don't declare the binding in the XML(when you don't use xsi:schemaLocation, xml-model, etc)

> but the public id can be something like "http://www.w3.org/2001/XMLSchema". Does using this make sense ?

This usecase is for XSD. In the case of XSD schema file, the XSD declares the `http://www.w3.org/2001/XMLSchema` namespace (which becomes the public identifier URI while resolving step). The XSD is associated with the xml.xsd file (to manage completion and allowed element,attributes validation for XSD (ex : xs:element is allowed, xs:bad is forbidden). 
The xml.xsd is retrived by the uri resolver extension https://github.com/eclipse/lemminx/blob/cd73f31a4081744b376df65ea5bb2e26464c6ea5/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/XSDURIResolverExtension.java#L52 (the info.getResolvedBy() should return "embedded xml.xsd")

rgrunber: Thanks. I've rebased against your new contribution.

rgrunber:
- For the public/system ID, and corresponding resolver, I've kept those as lists. I'm still not sure this is the nicest approach.
- For the binding kind, I set a different property for each value, that gets set to true if it exists. I think this is cleaner for this case because we have a pretty limited set of values.

USAGE_DATA.md
```diff
@@ -13,6 +13,12 @@ When telemetry events are enabled, the following information is emitted when the
     * The free, total, and max VM memory
  * Version information:
     * The server version number
+ * Text Document Information
+   * When a document is opened :
+      * The file extension (eg. `xml`, `xsd`, `dtd`)
+      * The associated grammar types (eg. `none`, `doctype`, `xml-model`, `xsi:schemaLocation`, `xsi:noNamespaceSchemaLocation`)
+      * The system/public identifier used for the grammar, if any
```
angelozerr, line 20: After though i wonder if its a relevant information? Its more a data user for me.  I will remove it.

rgrunber: Sounds fine. We can remove the system/public identifier for now. Could this not be used to determine if the schema is a local file or remote ?

rgrunber: Done. I've removed the identifier uri.

angelozerr:
Thanks @rgrunber !

PR #1106, opened 2021-08-23
Report grammar identifier to didOpen telemetry events.

### LINKED ISSUE #1105: Report schema identifier of XML document through telemetry event
As discussed in #1066 , one of the things that language clients might care to know, is what XML schemas are most popular. This might help focus efforts towards relevant support. For example, if a lot of people are using `http://maven.apache.org/xsd/maven-4.0.0.xsd`, then maybe more focus on lemminx-maven should be given.

As with the previous issue, this can be done through https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/model/ReferencedGrammarInfo.java

rgrunber:
- Fixes #1105
- When exactly 1 grammar reference exists per file, simply send the
  relevant values (not a list of size 1). Otherwise, send the values as
  a list


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/telemetry/DocumentTelemetryInfo.java
```diff
@@ -67,10 +70,25 @@
 						break;
 					}
 				}
+				if (identifier != null) {
```
rgrunber, line 73: We could probably filter out identifiers that are local. I wonder if we could just exclude anything with a file URI.

rgrunber: Done.

PR #1109, opened 2021-08-31
Cache ~/.m2/ to speed up GH actions

fbricon:


fbricon:
Before this change (in master)
Screenshot 2021-09-03 at 16 23 30

After this change (building this PR)
Screenshot 2021-09-03 at 16 23 53

.github/workflows/native-image.yaml
```diff
@@ -1,42 +1,65 @@
...
+        with:
+          path: |
+            /opt/hostedtoolcache/GraalVM
+            ~/hostedtoolcache/GraalVM
+            ~/.m2/repository
+            ~/.m2/wrapper
+            !~/.m2/repository/org/eclipse/lemminx
+          key: ${{ runner.os }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-node
```
rgrunber, line 28: Just curious about this value. Looking at https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key, this would match any cache key starting with ${{ runner.os }}-node and the most recent if there are multiple , but there aren't any that contain the "node" in the key declaration. Is it some pre-defined key somewhere else ?

rgrunber: @fbricon , I think this should just be `${{ runner.os }}- according to the docs. Let me know if that makes sense. I wouldn't mind trying this out on on JDT-LS as well.

fbricon:
as per https://github.com/actions/cache/blob/main/examples.md#java---maven, it should actually be `${{ runner.os }}-maven`

rgrunber:
> as per https://github.com/actions/cache/blob/main/examples.md#java---maven, it should actually be `${{ runner.os }}-maven`

That's fine. So the key should be `${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}` to match.

fbricon:
Actually, we should split the cache in 2, for Maven and Graalvm, but I don't know what to use for graalvm restore key

rgrunber (approved):
Looks fine to me. We can hopefully take a similar approach in JDT-LS.

PR #1112, opened 2021-09-03
Register a command "xml.check.file.pattern"

rgrunber:
- Create class CheckFilePattern
- "xml.check.file.pattern" validates that some file in the project
  folder will match against a given pattern

This permits https://github.com/redhat-developer/vscode-xml/issues/586 to be resolved on the server side.

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/commands/CheckFilePatternCommand.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/commands/CheckFilePatternCommand.java
@@ -1,0 +1,39 @@
+/*******************************************************************************
+* Copyright (c) 2021 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.extensions.contentmodel.commands;
+
+import org.eclipse.lemminx.services.extensions.commands.ArgumentsUtils;
+import org.eclipse.lemminx.services.extensions.commands.IXMLCommandService.IDelegateCommandHandler;
+import org.eclipse.lemminx.settings.PathPatternMatcher;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lsp4j.ExecuteCommandParams;
+import org.eclipse.lsp4j.jsonrpc.CancelChecker;
+
+/**
+ * XML Command "xml.check.file.pattern" whose first argument is a file pattern,
+ * and whose second argument is some file URI. The command returns true when
+ * the given pattern matches against the given file URI and false otherwise.
+ */
+public class CheckFilePatternCommand implements IDelegateCommandHandler {
+
+	public static final String COMMAND_ID = "xml.check.file.pattern";
+
+	@Override
+	public Object executeCommand(ExecuteCommandParams params, SharedSettings sharedSettings,
+			CancelChecker cancelChecker) throws Exception {
+		String pattern = ArgumentsUtils.getArgAt(params, 0, String.class);
+		String currentUri = ArgumentsUtils.getArgAt(params, 1, String.class);
+		PathPatternMatcher matcher = new PathPatternMatcher();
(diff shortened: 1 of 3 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/commands/CheckFilePatternCommand.java
```diff
@@ -0,0 +1,41 @@
...
+import java.nio.file.Paths;
+
+import org.eclipse.lemminx.services.IXMLDocumentProvider;
+import org.eclipse.lemminx.services.extensions.commands.ArgumentsUtils;
+import org.eclipse.lemminx.services.extensions.commands.IXMLCommandService.IDelegateCommandHandler;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lsp4j.ExecuteCommandParams;
+import org.eclipse.lsp4j.jsonrpc.CancelChecker;
+
+public class CheckFilePatternCommand implements IDelegateCommandHandler {
```
rgrunber, line 25: A description would be good.

rgrunber: And some tests.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/commands/CheckFilePatternCommand.java
```diff
@@ -0,0 +1,41 @@
...
+	public CheckFilePatternCommand(IXMLDocumentProvider documentProvider) {
+		this.documentProvider = documentProvider;
+	}
+
+	@Override
+	public Object executeCommand(ExecuteCommandParams params, SharedSettings sharedSettings,
+			CancelChecker cancelChecker) throws Exception {
+		String pattern = ArgumentsUtils.getArgAt(params, 0, String.class);
+		PathMatcher matcher = FileSystems.getDefault().getPathMatcher("glob:" + pattern);
+		return documentProvider.getAllDocuments().stream().anyMatch(doc -> matcher.matches(Paths.get(doc.getDocumentURI())));
```
rgrunber, line 39: I was hoping to get all files within the workspace folder(s), but it seems like this isn't possible without sending their URIs over in the data, which doesn't seem right. Seems like we can only compare against opened files.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/commands/CheckFilePatternCommand.java
```diff
@@ -0,0 +1,41 @@
...
+	private IXMLDocumentProvider documentProvider;
+
+	public CheckFilePatternCommand(IXMLDocumentProvider documentProvider) {
+		this.documentProvider = documentProvider;
+	}
+
+	@Override
+	public Object executeCommand(ExecuteCommandParams params, SharedSettings sharedSettings,
+			CancelChecker cancelChecker) throws Exception {
+		String pattern = ArgumentsUtils.getArgAt(params, 0, String.class);
```
angelozerr, line 37: We should have an uri too as argument. The idea is to validate that a given xml file matches à given pattern.

angelozerr: Use https://github.com/eclipse/lemminx/blob/a3058b72a344c89c854e44104c87faa3142b2a66/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/PathPatternMatcher.java#L42 to check that given uri matches the given pattern.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/commands/CheckFilePatternCommand.java
angelozerr: for the test, see https://github.com/eclipse/lemminx/tree/master/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/commands

rgrunber: Done.

angelozerr:
Thanks @rgrunber !

Issue #1045, opened 2021-05-20
Update lsp4j to 0.12.0


datho7561:
There are changes to the lsp4j interface that require changes to the code base. There are also new parameters that will need to be registered for reflection so that the binary server works.

datho7561:
After I have investigated more, I think the only changes needed are:
 * lsp4j 0.11.0 --> 0.12.0 in the pom.xml
 * gson 2.8.2 --> 2.8.6 in the pom.xml, to match the version required by lsp4j

I managed to get the binary server to launch without modifying the reflection configuration.

pablocabrera85:
Hi. Are you planning on including this change in the next release? Is there a date for it?

angelozerr:
Why do you need 0.12.0?

pablocabrera85:
Thanks for the fast reply :-).  We are currently running a composition of Language Servers in the same JVM. All of them have been updated from 0.9.0 to use 0.12.0 in order to consume the latest feature and avoid staying behind LSP4J API changes. 

Inside this ecosystem, we are also using lemminx 0.18.0. 

The features we had were not initially affected by the API signature changes until recently.

We have some level of classloader isolation between components, but the LSP4J API is not included in the list of dependencies that can have multiple versions running in our process. This is something we were thinking of tackling but if it was in lemminx plan to support 0.12.0, we could delay it.

angelozerr:
> . This is something we were thinking of tackling but if it was in lemminx plan to support 0.12.0, we could delay it.

As we would like to do a release soon, it's too dangerous to switch to it for now. And we have no plan for the moment to switch to it. I'm sorry, but if you wish you can create a PR.

mickaelistria:
> As we would like to do a release soon, it's too dangerous to switch to it for now

IMO, switching to newer LSP4J shouldn't happen for a bugfix release (such as 0.18.1), if this is to happen, it should be in a new "functional" release (eg 0.19.0) to clarify it may break compatibility with extensions.

pablocabrera85:
Thanks @angelozerr and @mickaelistria for your quick replies. 

Given that LSP4J API usually breaks compatibility with every new minor version I completely agree that such a change should not be included in a bugfix release (not even a minor if following semver strictly speaking). 

I'll take a look at your contribution guidelines to see if I can open a PR with the changes 👍.

angelozerr:
I close this issue since today LemMinx uses LSP4J 0.14.0

PR #1124, opened 2021-10-14
Update dependencies

akurtakov:


rgrunber:
Looking at https://github.com/eclipse/lemminx/issues/1045, we may need some additional work to support lsp4j 0.12.0.

PR #1135, opened 2021-12-08
Fixed tag rename for XML tags containing colons

AlexXuChen:
Fixed tag rename for XML tags containing colons, fixed duplicate rename for namespace tags.

Fixes redhat-developer/vscode-xml/issues/621

Behaviour:
Peek 2021-12-08 16-48

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLRename.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLRename.java
@@ -180,6 +180,9 @@ public class XMLRename {
 		String fullNodeName = element.getNodeName();
 		int indexOfColon = fullNodeName.indexOf(":");
 		if (indexOfColon > 0) {
+			// Take the prefix and local name of the namespace tag
+			String prefix = element.getPrefix();
+
 			Position startTagStartPosition = startTagRange.getStart();
 			Position startTagPrefixPosition = new Position(startTagStartPosition.getLine(),
 					startTagStartPosition.getCharacter() + indexOfColon);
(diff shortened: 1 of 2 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLRename.java
```diff
@@ -191,23 +194,35 @@ private boolean isRenameTagName(DOMDocument document, DOMElement element, Positi
...
+
+			if (newText.contains(":")) {
+				String[] tagParts = newText.split(":", 2);
+				List<TextEdit> tagTextEdits = new ArrayList<TextEdit>();
+				// Tag prefix rename
+				if (!prefix.equals(tagParts[0])) {
+					tagTextEdits.addAll(renameElementNamespace(xmlDocument, element, prefix.length(), tagParts[0]));
+				}
+				// Tag suffix/local name rename
+				if (!prefix.equals(tagParts[1])) {
```
rgrunber, line 218: Should this be `!suffixName.equals(tagParts[1])`.

Take the case of `<ns:nos>` and attempt to rename to `<ns:ns>`. Nothing happens.

rgrunber (commented):
Looks good to me overall. @angelozerr , let me know if it's fine with you.

angelozerr:
@AlexXuChen it should be nice to have some tests with your PR.

AlexXuChen:
> @AlexXuChen it should be nice to have 2 tests with your PR:
> 
> * test which rename ns
> * test which rename local name

These tests exist in https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLRenameTest.java, although not explicitly called `ns:...`, however the behaviour is the same. I made the bug fix such that no tests would break.

angelozerr:
> These tests exist in https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLRenameTest.java, although not explicitly called ns:..., however the behaviour is the same. I made the bug fix such that no tests would break.

The main idea is to copy/paste a test with ns rename and change the newText with `ns:BBB` instea dof `BBB` and do the same thing for local name rename.

angelozerr:
It works like a charm, good job @AlexXuChen !

PR #1136, opened 2021-12-10
Improved accuracy of attribute value quick fix suggestions

AlexXuChen:
Improved accuracy of attribute value Quick Fix suggestions.

Fixes redhat-developer/vscode-xml/issues/618

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_attribute_3CodeAction.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_attribute_3CodeAction.java
@@ -14,6 +14,8 @@ package org.eclipse.lemminx.extensions.contentmodel.participants.codeactions;
 
 import java.util.Collection;
 import java.util.List;
+import java.text.Collator;
+import java.util.TreeSet;
 
 import org.eclipse.lemminx.commons.CodeActionFactory;
 import org.eclipse.lemminx.dom.DOMAttr;
@@ -29,6 +31,8 @@ import org.eclipse.lsp4j.CodeAction;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
+
+import static org.eclipse.lemminx.utils.StringUtils.isSimilar;
 
 /**
  * Code action to fix cvc-attribute-3 error.
@@ -48,6 +52,7 @@ public class cvc_attribute_3CodeAction implements ICodeActionParticipant {
 				String attributeName = attr.getName();
 				ContentModelManager contentModelManager = componentProvider.getComponent(ContentModelManager.class);
 				Collection<CMDocument> cmDocuments = contentModelManager.findCMDocument(element);
+				String attributeValue = attr.getValue();
 				for (CMDocument cmDocument : cmDocuments) {
 					CMAttributeDeclaration cmAttribute = cmDocument.findCMAttribute(element, attributeName);
 					if (cmAttribute != null) {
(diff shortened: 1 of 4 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_attribute_3CodeAction.java
```diff
@@ -51,19 +53,23 @@ public void doCodeAction(Diagnostic diagnostic, Range range, DOMDocument documen
 				for (CMDocument cmDocument : cmDocuments) {
 					CMAttributeDeclaration cmAttribute = cmDocument.findCMAttribute(element, attributeName);
 					if (cmAttribute != null) {
+						String attributeValue = attr.getValue();
```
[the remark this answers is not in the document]
AlexXuChen: If there are no suggestions retrieved based on similarity, should we suggest all possible values?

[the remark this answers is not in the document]
rgrunber: Yes, I think that makes sense for now. A narrow set of results when we have good matches, and all results when no particular matches can be found.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_attribute_3CodeAction.java
```diff
@@ -14,6 +14,9 @@
 
 import java.util.Collection;
 import java.util.List;
+import java.text.Collator;
+import java.util.ArrayList;
```
rgrunber, line 18: ArrayList not used anywhere, so it can be removed.

rgrunber (approved):
LGTM and works as expected. I don't see any similar tests for code actions, so I guess this should be fine.

@angelozerr , let me know if this can be merged.

angelozerr:
It should be nice to write tests for this feature. We have already test for similar with local name at https://github.com/eclipse/lemminx/blob/e41b18cfc6541109907d6a035da2972b34b5ca6c/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaDiagnosticsTest.java#L556

I suggest you write a test with similar for attribute (see  tests which defines attribute enumeration by searching cvc_attribute_3 in https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaDiagnosticsTest.java

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
```diff
@@ -658,6 +657,20 @@ public void fuzzyElementNamesWithPrefixAndNoMatch() throws Exception {
...
+	@Test
+	public void fuzzyElementMemberValueCodeActionTest() throws Exception {
+		String xml = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\r\n" + //
+				"<dress \r\n" +
+				"xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"\r\n" +
+				"xsi:noNamespaceSchemaLocation=\"src/test/resources/xsd/dressSize.xsd\"\r\n" +
+				"size=\"larg\"/>";
+		Diagnostic diagnostic = d(4, 5, 4, 11, XMLSchemaErrorCode.cvc_attribute_3,
+				"cvc-attribute.3: The value 'larg' of attribute 'size' on element 'dress' is not valid with respect to its type, 'SizeType'.");
+		testCodeActionsFor(xml, diagnostic,
```
AlexXuChen, line 669: There's currently an issue where testCodeActionsFor, where it takes all the diagnostics for the 'actual' rather then just the ones that match the diagnostic.

AlexXuChen: This problem likely comes from `cvc-attribute.3` errors also having the `cvc_datatype_valid_1_2_3` error code

angelozerr: You need to add this error in your diagnostic

AlexXuChen: If I add the diagnostic, the same issue occurs. Additionally, testCodeActionsFor only accepts one diagnostic

angelozerr:
Thanks @AlexXuChen !

PR #1146, opened 2022-01-04
'No definition found' when using 'Go to Definition' for types defined in imported XSD

angelozerr:
'No definition found' when using 'Go to Definition' for types defined in imported XSD

Fixes [#632](https://github.com/redhat-developer/vscode-xml/issues/632)


fbricon:
Thanks @angelozerr

PR #1153, opened 2022-01-14
End Tag completion should be given higher priority than <!-- and <!CDATA[

### LINKED ISSUE #1150: End Tag completion should be given higher priority than <!-- and <!CDATA[
Currently when typing an End-Tag a pop up is asking for "Comment or End-Tag".

As comments in the text content of a node:
`<version.maven-compiler-plugin>3.8.1<!-- some comment --></version.maven-compiler-plugin>`
the continuous typing should just add the end tag as only option.

resulting in
`<version.maven-compiler-plugin>3.8.1</version.maven-compiler-plugin>`

This would enhance the typing of xml files.

(maybe when end tag already there type an "<" may suggest a comment)

See https://github.com/eclipse-m2e/m2e-core/issues/490#issuecomment-1012000085
image

angelozerr:
End Tag completion should be given higher priority than <!-- and <!CDATA[

Fixes #1150

### DIFF
--- a/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/comment-snippets.json
+++ b/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/comment-snippets.json
@@ -7,6 +7,7 @@
     "body": [
       "<!-- $1-->"
     ],
-    "description": "Insert Comment"
+    "description": "Insert Comment",
+    "sortText": "ZZa"
   }
  }
--- a/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/cdata-snippets.json
+++ b/org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/cdata-snippets.json
@@ -7,6 +7,7 @@
     "body": [
       "<![CDATA[ $1]]>"
     ],
-    "description": "Insert CDATA"
+    "description": "Insert CDATA",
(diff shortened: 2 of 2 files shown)


org.eclipse.lemminx/src/main/resources/org/eclipse/lemminx/services/snippets/comment-snippets.json
```diff
@@ -7,6 +7,7 @@
     "body": [
       "<!-- $1-->"
     ],
-    "description": "Insert Comment"
+    "description": "Insert Comment",
+    "sortText": "ZZa"
```
mickaelistria, line 11: sinze 'a'(== 97) > `A`(==65), should the sortext be `zzA` instaed to not make it lower than eg `Hello` ?

angelozerr: Your usecase Hello works with my current PR:

image

mickaelistria: Maybe vscode does lozercase the sorttext or something. I'll try with LSP4E soon.

mickaelistria:
OK, I verified with LSP4E and this would be supposed to work (although some other bug in LSP4E is also in play here and needs to be fixed to get expected results), so it seems good to merge from LemMinX perspective.

PR #1161, opened 2022-01-31
Update gson to 2.8.9

fbricon:
gson 2.8.2 is flagged as insecure:

gson 2.8.2 is flagged as insecure

 upgrading to recommended 2.8.9


angelozerr:
Thanks @fbricon !

PR #1180, opened 2022-02-28
Added setting to disable files in completion settings

AlexXuChen:
Added setting to disable files in completion settings.

Part of redhat-developer/vscode-xml/issues/670

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLCompletionSettings.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLCompletionSettings.java
@@ -26,13 +26,16 @@ public class XMLCompletionSettings {
 
 	private boolean autoCloseRemovesContent;
 
-	public XMLCompletionSettings(boolean autoCloseTags, boolean autoCloseRemovesContent) {
+	private boolean fileCompletion;
+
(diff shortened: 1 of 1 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLCompletionSettings.java
```diff
@@ -26,13 +26,16 @@
 
 	private boolean autoCloseRemovesContent;
 
-	public XMLCompletionSettings(boolean autoCloseTags, boolean autoCloseRemovesContent) {
+	private boolean fileCompletion;
```
angelozerr, line 29: This settings belongs to the file path extension, so please don't add that in this class. I suggest that you see sample like https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/ContentModelSettings.java

In other words, you should create a FilePathSettings in https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/settings

Please rename general to filepath.

angelozerr:
Fixed with https://github.com/eclipse/lemminx/pull/1293

PR #1159, opened 2022-01-27
Updated GraalVM version to 22.0.0.2

AlexXuChen:
Updated GraalVM version used in binary verification builds to 22.0.0.2
- Update native-maven-plugin to 0.9.10
- Use setup-graalvm 5.0 GitHub Action
- Replace deprecated "graalvm-version" with "graalvm" & "java"
- Use msvc-dev-cmd 1.10.0 GitHub Action

Fixes #1158 

Some changes adopted from https://github.com/eclipse/lemminx/pull/1173

Signed-off-by: Alexander Chen <alchen@redhat.com>

PR #1184, opened 2022-03-04
Remove native-image GitHub action from repository

rgrunber:
Create an entry under GH Releases for built artifacts.
    
    - Triggered only when event is not a "pull_request"
    - Create a tag "latest" marked as pre-release and attach latest binary
      artifacts from the source branch
    - Use actions/download-artifact no-args to fetch all uploaded artifacts
      into their own folder
    
(this builds on top of https://github.com/eclipse/lemminx/pull/1159)

Remaining items :

- [x] Package binary checksum data into release
- [x] Binaries should be zipped
- [x] Rename binaries `lemminx-${short-sha}-${platform}` -> `lemminx-${platform}`


rgrunber:
We can't run this under eclipse/lemminx. It needs to be done under redhat-developer/vscode-xml.

So disbale binary uploading and pre-release generation.

rgrunber:
Re-located to https://github.com/redhat-developer/vscode-xml/blob/master/.github/workflows/native-image.yaml so this can be merged now.

Issue #1187, opened 2022-03-21
Adopt linkedEditingRanges wordPattern property


rgrunber:
@vrubezhny mentioned that LemMinX doesn't use the `wordPattern` property to define when the linked editing range should exit the editing mode. I believe he ran into a case where the client (Eclipse?) had a different set of word patterns and there was some unexpected behaviour. I'll let him provide more details if needed.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#linkedEditingRanges

`wordPattern` value for some languages that may be associated with LemMinX
https://github.com/microsoft/vscode/blob/main/extensions/html/language-configuration.json#L35
https://github.com/microsoft/vscode/blob/main/extensions/xml/xml.language-configuration.json#L37

angelozerr:
Fixed with https://github.com/eclipse/lemminx/issues/1187

Thanks @@vrubezhny !

PR #1188, opened 2022-03-25
Adopt linkedEditingRanges wordPattern property #1187

vrubezhny:
Signed-off-by: Victor Rubezhny <vrubezhny@redhat.com>

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLLinkedEditing.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLLinkedEditing.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2021 Red Hat Inc. and others.
+* Copyright (c) 2021, 2022 Red Hat Inc. and others.
 * All rights reserved. This program and the accompanying materials
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/legal/epl-v20.html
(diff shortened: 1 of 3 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLLinkedEditing.java
```diff
@@ -34,6 +34,12 @@
 
 	private static Logger LOGGER = Logger.getLogger(XMLLinkedEditing.class.getName());
 
+	public static final String XML_ELEMENT_WORD_PART_PATTERN =
```
angelozerr, line 37: I'm not an expert inregexp, but are you sure that it's the same regexp than https://github.com/microsoft/vscode/blob/1cc6ee1c711858b4f4af204d0044c62b70828ab8/extensions/xml/xml.language-configuration.json#L37 ?

mickaelistria: It's indeed a different pattern. But testing with node CLI gives the impression that those are equivalant (`'\u00C0' === '\u{C0}'` returns true), and this has the advantage of being parsed with the Java regexp as well as with the JS regexp.

vrubezhny: There are at least two different things between the XML element name pattern I used inlemminx and the one used in VSCode:
- The VSCode pattern is a full pattern describing a whole word (contains starting character definition and the following character definitions) - it's to be used when searching for a "word match", while in lemminx we're verifying the only last typed character (or a single character at some position), so the pattern is really a wordpart-like pattern (doesn't define a first word character).  I think it's still possible to make lemminx to search for the word by a linkedEditingRange region and use a whole word pattern (VSCode-like) to verify the word found, but imho it will add some possibility to fail for the validation (all this is a post processing, so trying to peek up a word by a linkedEditingRegion may be erroneous as the document may be change at  the moment of the event processing). However imho it's still possible.
- A minor difference - my pattern doesn't allow`\\u{10000}-\\u{EFFFF}` character codes, I'll modify the PR in order to enable these characters.

vrubezhny:
I have updated the PR - now the wordPattern used for XML Element Name is full Java equivalent of the wordPattern used in VSCode. Also, it's a full pattern for a word-validation (not char-by-char validation) as it's done in VSCode - so, if properly used (f.i. in LSP4E) everything should work the same way as VSCode does.

mickaelistria:
Sorry, I didn't mean to merge this particular PR (I was thinking about LSP4E). However, it seems safe enough to keep it for the moment and will ease further testing; so I don't think it's worth reverting; but if anyone has reasons to prefer a revert, we can do it.

PR #1198, opened 2022-04-27
Send (aggregated) server.document.open events at fixed periods.

rgrunber:
- Use an executor with fixed period (1 day) to send server.document.open
  events (if any) to the client


rgrunber:
I've tried this out with a delay of 15min and it seems to behave as expected.

rgrunber:
CC @fbricon , let me know if we should include this in the release.

datho7561 (approved):
This change looks good to me!

PR #1202, opened 2022-05-19
Wait a bit after change before sending diagnostics

### LINKED ISSUE #1162: Wait a bit after change before sending diagnostics
In LSP4E, linkedEdit is implemented so that it sends 2 document events and so 2 didChange events (I know this is not perfect, but it's how Eclipse Platform currently does that and not something I can change immediately). As a consequence, typing one char in a tag sends 1 change, LemMinX immediately returns an error diagnostic for non-matching end tag, then LSP4E/Eclipse does fix the other tag and sends the event to LemMinX, that will then remove the error diagnostic. This cause the diagnostic to show and disappear immediately, almost blinking above code, it doesn't look nice.
Although it is something that could be fixed on Eclipse side, I'm wondering whether LemMinX should be more tolerant of fast changes. For example, someone typing really fast may not want to see the error as they're typing, but would like some delay (~50ms) before the diagnostics are reported. This would cause less "blinking diagnostics" and could also reduce the amount of parsing done by LemMinX: someone typing 10chars at 20ms/char would then see the error after ~250ms, 50ms after typing the last char, and LemMinX would have parsed the file only once instead of 10 times. I think the gain in UX is worth it, especially since it would be an opportunity to gain CPU cycles on server side.

angelozerr:
Wait a bit after change before sending diagnostics

Fixes #1162

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ModelValidatorDelayer.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ModelValidatorDelayer.java
@@ -1,0 +1,82 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.commons;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+/**
+ * Validate a given model document with delay.
+ * 
+ * @author Angelo ZERR
+ *
+ * @param <T>
+ */
+public class ModelValidatorDelayer<T> {
+
+	private static final long DEFAULT_VALIDATION_DELAY_MS = 500;
+
+	private final ScheduledExecutorService executorService;
+
+	private final Consumer<T> validator;
+
+	private final Map<String, Future<?>> pendingValidationRequests;
+
+	private final long validationDelayMs;
+
+	public ModelValidatorDelayer(Consumer<T> validator) {
+		this(Executors.newScheduledThreadPool(2), validator, DEFAULT_VALIDATION_DELAY_MS);
+	}
+
+	public ModelValidatorDelayer(ScheduledExecutorService executorService, Consumer<T> validator,
+			long validationDelayMs) {
+		this.executorService = executorService;
+		this.validator = validator;
+		this.pendingValidationRequests = new HashMap<>();
+		this.validationDelayMs = validationDelayMs;
+	}
+
+	/**
+	 * Validate the given model <code>document</code> identified by the given
+	 * <code>uri</code> with a delay.
+	 * 
+	 * @param uri      the document URI.
+	 * @param document the document model to validate.
+	 */
+	public void validateWithDelay(String uri, T document) {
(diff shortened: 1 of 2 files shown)


angelozerr:
This PR publish diagnostics with 200ms as delay.

Here a demo with vscode-xml and linked edting range:

PublishDiagnosticsWithDelayDemo.

Instead of reporting a diagnostics per typed character, it generate few diagnotics.

angelozerr:
[@vrubezhny](https://github.com/vrubezhny) could you review too this PR and tell me if it improves performance for Eclipse IDE with WWD please.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ModelValidatorDelayer.java
```diff
@@ -0,0 +1,82 @@
...
+	 * <code>uri</code> with a delay.
+	 * 
+	 * @param uri      the document URI.
+	 * @param document the document model to validate.
+	 */
+	public void validateWithDelay(String uri, T document) {
+		cleanPendingValidation(uri);
+		Future<?> request = executorService.schedule(() -> {
+			synchronized (pendingValidationRequests) {
+				pendingValidationRequests.remove(uri);
```
fbricon, line 64: shouldn't you cancel the pendingValidationRequests too?

angelozerr: I think we should not do that because it will cancel the current future which should process the validation.

fbricon: ok

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ModelValidatorDelayer.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Validate a given model document with delay.
+ * 
+ * @author Angelo ZERR
+ *
+ * @param <T>
+ */
+public class ModelValidatorDelayer<T> {
+
+	private static final long DEFAULT_VALIDATION_DELAY_MS = 200;
```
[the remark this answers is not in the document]
angelozerr: Yes I know, but I think it's perhaps too long, but if you wish I can update to 500ms

AlexXuChen (commented):
The delay is working as described, however, perhaps we can optimize diagnostic reporting further by only sending the affected diagnostic in the response (i.e. if there are 2 diagnostics in 2 separate files in the workspace, and one diagnostic is changed, the other still sends a `publishDiagnostics` response)

angelozerr:
When you say affected diagnostic you mean the document which is updating?

If it that the report of the diagnostic is just for the updated document.

But if you save the document in this case the server will report diagnostics for all opened document. Perhaps it is tour usecase

AlexXuChen:
> But if you save the document in this case the server will report diagnostics for all opened document. Perhaps it is tour usecase

Yes, I update the diagnostic by saving, so all diagnostics are reported again.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ModelValidatorDelayer.java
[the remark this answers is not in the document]
angelozerr: To be consistent with HTML LS, I updated with 500ms.

angelozerr:
> Yes, I update the diagnostic by saving, so all diagnostics are reported again.

It is the expected behvior.  This PR provides delay for reporting diagnostic when document changed. When document it is saved, there is no delay and I think it is not annoying to report diagnostics for all opened files just on save.

The problem is when you change document, you must try to report few diagnotics to improve : 

 * performance on XML languagse server side. if user type N characters, in master it validate N times and report N diagnostics. With this PR if user type N characters, it should validate one time and report one diagnotic
 * performance on LSP client side.

angelozerr:
Thanks @fbricon @AlexXuChen for your review!

PR #1209, opened 2022-05-27
added folding setting to keep displaying the closing tag after folding

JessicaJHee:
Requires redhat-developer/vscode-xml/pull/714

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
@@ -48,10 +48,6 @@ class XMLFoldings {
 	public XMLFoldings(XMLExtensionsRegistry extensionsRegistry) {
 		this.extensionsRegistry = extensionsRegistry;
 	}
-
-	public boolean isIncludeClosingTagInFold(XMLFoldingSettings context) {
-		return context.isIncludeClosingTagInFold();
-	}
 	
 	class TagInfo {
 
@@ -188,6 +184,10 @@ class XMLFoldings {
 			LOGGER.log(Level.SEVERE, "Foldings received a StackOverflowError while scanning the document", e);
 		}
 		return ranges;
+	}
+
+	private static boolean isIncludeClosingTagInFold(XMLFoldingSettings settings) {
+		return settings.isIncludeClosingTagInFold();
 	}
 
 	private static int addRange(FoldingRange range, List<FoldingRange> ranges) {
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
@@ -141,6 +141,11 @@ class XMLFoldings {
 									stack.remove(j--);
 								}
 								int endLine = startLine;
+
+								if (!isIncludeClosingTagInFold(context)){
+									endLine -= 1;
+								}
+
 								startLine = stackElement.startLine;
 								if (endLine > startLine && prevStart != startLine) {
 									FoldingRange range = new FoldingRange(startLine, endLine);
@@ -152,6 +157,11 @@ class XMLFoldings {
 					} else {
 						int endLine = document.positionAt(scanner.getTokenOffset() + scanner.getTokenLength())
 								.getLine();
+
+						if (!isIncludeClosingTagInFold(context)){
+							endLine -= 1;
+						}
+						
 						if (startLine < endLine) {
 							FoldingRange range = new FoldingRange(startLine, endLine);
 							range.setKind(FoldingRangeKind.Comment);
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
@@ -49,6 +49,10 @@ class XMLFoldings {
 		this.extensionsRegistry = extensionsRegistry;
 	}
 
+	public boolean isIncludeClosingTagInFold(XMLFoldingSettings context) {
+		return context.isIncludeClosingTagInFold();
+	}
+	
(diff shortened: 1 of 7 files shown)


AlexXuChen (changes requested):
Features working as intended:
Peek 2022-05-27 16-57
A couple of changes, and add/change tests in https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFoldingsTest.java with the setting enabled/disabled.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFoldingSettings.java
```diff
@@ -32,4 +32,34 @@ public FoldingRangeCapabilities getCapabilities() {
 	public Integer getRangeLimit() {
 		return capabilities != null ? capabilities.getRangeLimit() : null;
 	}
+
+	public static final boolean DEFAULT_INCLUDE_CLOSING_TAG_IN_FOLD = true;
+
+	private boolean includeClosingTagInFold;
+
+	public XMLFoldingSettings() {
+		this.includeClosingTagInFold = DEFAULT_INCLUDE_CLOSING_TAG_IN_FOLD;
```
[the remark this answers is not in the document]
angelozerr: Indeed it is a good practive to have just one constructor which initialize all fields.

angelozerr:
It shoud be nice to have test with this new settings.

I suggest that you: define a new static method in
https://github.com/eclipse/lemminx/blob/3ed05cb913f471c582f6de498211a5fc104dbca9/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFoldingsTest.java#L236

```java

private static void assertRanges(String[] lines, ExpectedIndentRange[] expected) {
	assertRanges(lines, expected, "", null, new XMLFoldingSettings());
}

private static void assertRanges(String[] lines, ExpectedIndentRange[] expected, XMLFoldingSettings settings) {
	assertRanges(lines, expected, "", null, settings);
}

private static void assertRanges(String[] lines, ExpectedIndentRange[] expected, String message, Integer nRanges) {
    assertRanges(lines, expected, message, nRanges, new XMLFoldingSettings());
}

private static void assertRanges(String[] lines, ExpectedIndentRange[] expected, String message, Integer nRanges, XMLFoldingSettings settings)
TextDocument document = new TextDocument(String.join("\n", lines), "test://foo/bar.json");
		DOMDocument xmlDocument = DOMParser.getInstance().parse(document, null);
		XMLLanguageService languageService = new XMLLanguageService();

		FoldingRangeCapabilities context = new FoldingRangeCapabilities();
		context.setRangeLimit(nRanges);
		settings.setCapabilities(context);
		List<FoldingRange> actual = languageService.getFoldingRanges(xmlDocument, settings);

		List<ExpectedIndentRange> actualRanges = new ArrayList<>();
		for (FoldingRange f : actual) {
			actualRanges.add(r(f.getStartLine(), f.getEndLine(), f.getKind()));
		}
		Collections.sort(actualRanges, (r1, r2) -> r1.startLine - r2.startLine);
		assertArrayEquals(expected, actualRanges.toArray(), message);
}
```

after that you can duplicate test and set your new settings includeClosingTagInFold:

```
@Test
	public void testFoldOneLevelWithincludeClosingTagInFoldAsFalse () {
		String[] input = new String[] {
			/*0*/"<html>",
			/*1*/"Hello",
			/*2*/"</html>"
		};
                XMLFoldingSettings settings = new XMLFoldingSettings ();
                settings.setIncludeClosingTagInFold(false)
		assertRanges(input,  new ExpectedIndentRange[] {r(0, 2)}, );
	}
```

Please not the test will fail, please adjust the proper ranges.

And you can do the same thing for existing tests.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
```diff
@@ -49,6 +49,10 @@ public XMLFoldings(XMLExtensionsRegistry extensionsRegistry) {
 		this.extensionsRegistry = extensionsRegistry;
 	}
 
+	public boolean isIncludeClosingTagInFold(XMLFoldingSettings context) {
```
angelozerr, line 52: As it is not an important method, I will move it after `getFoldingRanges` which is main entry method. More this method is used only in this class, so you can set to private. At the end, this method don't need some fields of the XMLFoldings, so I will add static, in other words : 

`private static isIncludeClosingTagInFold(XMLFoldingSettings foldingSettings)`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFoldingSettings.java
```diff
@@ -32,4 +32,35 @@ public FoldingRangeCapabilities getCapabilities() {
 	public Integer getRangeLimit() {
 		return capabilities != null ? capabilities.getRangeLimit() : null;
 	}
+
+	public static final boolean DEFAULT_INCLUDE_CLOSING_TAG_IN_FOLD = false;
```
angelozerr, line 36: this static filed is used only in this class, set it to `private`.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFoldingsTest.java
```diff
@@ -218,32 +218,206 @@ public void	testLimit()  {
...
+			/*1*/"Hello",
+			/*2*/"</html>"
+		};
+		XMLFoldingSettings settings = new XMLFoldingSettings ();
+        settings.setIncludeClosingTagInFold(true);
+		assertRanges(input,  new ExpectedIndentRange[] {r(0, 2)}, settings);
+	}
+
+	@Test
+	public void	testFoldTwoLevelWithincludeClosingTagInFoldAsTrue ()  {
```
angelozerr, line 247: remove tabs with space (`public void test...`). Do the samething for other test methods

angelozerr:
Great job @JessicaJHee !

PR #1211, opened 2022-05-31
Improve DOM parser memory

angelozerr:
Improve DOM parser memory


angelozerr:
If you start https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/performance/DOMParserPerformance.java in master:

image

You can see that problem comes byte[] because there are for each attribute name and value a substring:

image

with this PR:

image

You can see byte[] uses les memory:

image

rgrunber:
I had to set my xml vmargs to about `1G` to test content.xml (defaul tis 64MB). It went from a heapsize of ~450MB after opening, to around ~320MB, so that seems about right. I think this is fine to merge.

angelozerr:
@rgrunber the real benefit is when you update the editor.

angelozerr:
Thanks @rgrunber for you review!

PR #1212, opened 2022-06-01
DOCTYPE entities interfering with hover annotation display of tags / attributes

angelozerr:
DOCTYPE entities interfering with hover annotation display of tags / attributes.

Fixes https://github.com/redhat-developer/vscode-xml/issues/716


angelozerr:
I need to write test but it should fix the issue https://github.com/redhat-developer/vscode-xml/issues/716. The problem with this issue is when XSD is bound with file association and there are DTD internal, it loose the file association link. The PR should fix that and fixes too:

 * completion based on XSD:

image

 * go to type definition

image

angelozerr:
Tests are written now.Thanks @AlexXuChen for your review.

PR #1216, opened 2022-06-02
Parse of DOM document should be not done in a Thread

### LINKED ISSUE #1215: Parse of DOM document should be not done in a Thread
By playing with the large file https://github.com/eclipse/lemminx/blob/master/org.eclipse.lemminx/src/test/resources/xml/content.xml I notice that we have too many Thread and even if we don't update the editor. The feature which higlight this problem is the start of  linked edit range feature which higlight the start/end tag, which is very slowly although we don't update the editor.

Here a demo which shows the problem:

LinkedEditorWithLargeFile1

As you can see the linked editing range (which colorize as red the tag) is so many slow. It seems that it is the call of CodeAction which get the indent information from client which uses several Thread https://github.com/eclipse/lemminx/blob/3ed3054de4144d3981d84a50630d70d7b40ab4dc/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java#L488

Today the DOM document is parsed in a Thread with CompletableFuture but we don't need that since parse is already done in a Thread (completion, hover, etc features are already executed in a Thread, so we don't need a CompletbaleFuture to parse the DOMDocument).

If we don't parse the DOM document in a CompletableFuture, here the result:

LinkedEditorWithLargeFile2

angelozerr:
Parse of DOM document should be not done in a Thread

Fixes #1215

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLLanguageServer.java
@@ -290,7 +290,7 @@ public class XMLLanguageServer implements ProcessLanguageServer, XMLLanguageServ
 
 	@Override
 	public CompletableFuture<AutoCloseTagResponse> closeTag(TextDocumentPositionParams params) {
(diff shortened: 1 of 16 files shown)


rgrunber:
This definitely improves linked editing range responsiveness for me.

PR #1219, opened 2022-06-03
Added support for folding <!DOCTYPE

### LINKED ISSUE #1213: Support folding for <!DOCTYPE
Given this xml file which defines DOCTYPE with a lof ENTITY:

```xml
<!DOCTYPE opt [
  <!ENTITY size "short">			
  <!ENTITY ResolutionX "1920"> 		
  <!ENTITY ResolutionY "1080">		
  <!ENTITY LevelMeterX "752">				
  <!ENTITY LevelMeterY "1036">			<!-- 44 -->				
  <!ENTITY PipDisplayX "845">				
  <!ENTITY PipDisplayY "1058">			<!-- 22 -->				
  <!ENTITY MainMenuButtonX "1191">
  <!ENTITY MainMenuButtonY "1011">		<!-- 69 -->
  <!ENTITY ToolbarButtonLeftX "707">
  <!ENTITY ToolbarButtonRightX "1194">
  <!ENTITY ToolbarButtonY1 "912">		<!--168 -->
  <!ENTITY ToolbarButtonY2 "947">		<!-- 133 -->
  <!ENTITY ToolbarButtonY3 "982">		<!-- 98 -->
  <!ENTITY QuickslotX "750">
  <!ENTITY QuickslotY "991">			<!-- 89 -->
  <!ENTITY QuickslotPageX "706">
  <!ENTITY QuickslotPageY "999">		<!-- 81 -->
  <!ENTITY LoadingScreenBarY "908">
  <!ENTITY LoadingScreenHintX "235">	
  <!ENTITY LoadingScreenHintY "752">
  <!ENTITY LoadingScreenWindowX "238">
  <!ENTITY LoadingScreenWindowY "240">
  <!ENTITY LoadingScreenWindowWidth "1024">
  <!ENTITY LoadingScreenWindowHeight "512">
  <!ENTITY LoadingScreenTextWidth "1000">
  <!ENTITY LoadingScreenAnimBox "100">
  <!ENTITY LoadingScreenAnimX "916">
  <!ENTITY LoadingScreenAnimY "-200">
]>
<opt>
  
</opt>
```

It should be nice to fold the DOCTYPE. 

This sample comes from https://github.com/redhat-developer/vscode-xml/issues/716

JessicaJHee:
DoctypeFoldingDemo

Fixes #1213 

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
@@ -75,6 +75,7 @@ class XMLFoldings {
 			while (token != TokenType.EOS) {
 				cancelChecker.checkCanceled();
 				switch (token) {
+				case DTDStartDoctypeTag:
 				case StartTag: {
 					String tagName = scanner.getTokenText();
 					int startLine = document.positionAt(scanner.getTokenOffset()).getLine();
@@ -92,6 +93,7 @@ class XMLFoldings {
 					}
 					// fallthrough
 				case EndTagClose:
+				case DTDEndDoctypeTag:
 				case StartTagSelfClose: {
 					int i = stack.size() - 1;
 					while (i >= 0 && !stack.get(i).tagName.equals(lastTagName)) {
--- a/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFoldingsTest.java
+++ b/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/XMLFoldingsTest.java
@@ -232,6 +232,17 @@ public class XMLFoldingsTest {
 	}
 
 	@Test
+	public void testFoldDoctype()  {
+		String[] input = new String[] {
+			/*0*/"<!DOCTYPE opt [",
+			/*1*/"  <!ENTITY size \"short\">",
+			/*2*/"  <!ENTITY ResolutionX \"1920\"> 	",
+			/*3*/"]>",
+		};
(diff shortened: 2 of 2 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFoldings.java
```diff
@@ -75,6 +75,7 @@ public List<FoldingRange> getFoldingRanges(TextDocument document, XMLFoldingSett
 			while (token != TokenType.EOS) {
 				cancelChecker.checkCanceled();
 				switch (token) {
+				case DTDStartDoctypeTag:
 				case StartTag: {
 					String tagName = scanner.getTokenText();
```
angelozerr, line 80: Do you need tag name for the case of DOCTYPE?

JessicaJHee: I believe so, its used to add the start tag to the stack with `stack.add(new TagInfo(startLine, tagName));` so we can keep track of the startline. I could also set a default value for doctype when creating a new TagInfo instead of using `scanner.getTokenText();`, but I'm not sure if that's necessary.

angelozerr: > I could also set a default value for doctype when creating a new TagInfo instead of using scanner.getTokenText();, but I'm not sure if that's necessary.

I agree but it is just for the understand. It is strange to have tagName for DOCTYPE. But we can let like this.

angelozerr:
Great job @JessicaJHee

PR #1224, opened 2022-06-07
Improve CodeAction performance with CodeAction#data & resolveCodeAction

### LINKED ISSUE #941: Improve CodeAction perfromance with CodeAction#data & resolveCodeAction
LSP 3.16 provides 2 new CodeAction support that LemMinx should use for their code action:

 * resolveCodeAction -> https://github.com/eclipse/lsp4j/blob/c0b921895b4576feeb355eea7a053a0d56daba7b/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/services/TextDocumentService.java#L263

For instance, LemMinx provides several code action to generate XSD, DTD from the given XML file which is not bound to a grammar. This process should be done in the resolveCodeAction instead of doing that every time that XML root element is hovered.

 * use CodeAction#data -> https://github.com/eclipse/lsp4j/blob/985a92bc0a0f543aa379edecbf0ddd56bed4df6f/org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/CodeAction.java#L104 

For instance, when a XML doesn't respect a XSD, for an element, we have a CodeAction which suggest the proper element. This CodeAction recompute those available elements although the validation provided them. The idea is to generate diagnostics with data which contains allowed element (coming from the Xerces XSD validator) https://github.com/eclipse/lsp4j/blob/985a92bc0a0f543aa379edecbf0ddd56bed4df6f/org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/Diagnostic.java#L93 and code action will be able to reuse this data instead of recomputing allowed elements.

angelozerr:
Improve CodeAction performance with CodeAction#data & resolveCodeAction

Fixes #941


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCodeActions.java
```diff
@@ -68,4 +68,23 @@ public List<CodeAction> doCodeActions(CodeActionContext context, Range range, DO
...
 	}
+
+	public CodeAction resolveCodeAction(CodeAction unresolved, DOMDocument document, SharedSettings sharedSettings,
+			CancelChecker cancelChecker) {
+		for (ICodeActionParticipant codeActionParticipant : extensionsRegistry.getCodeActionsParticipants()) {
+			try {
+				cancelChecker.checkCanceled();
+				if (codeActionParticipant.resolveCodeAction(unresolved, document, sharedSettings, extensionsRegistry,
+						cancelChecker)) {
+					return unresolved;
```
fbricon, line 79: it resolved but you're returning "unresolved". The naming sucks.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/data/DataEntryField.java
```diff
@@ -0,0 +1,20 @@
+package org.eclipse.lemminx.services.data;
```
fbricon, line 1: copyright header

datho7561:
So far, this looks pretty good. The API is pretty good, and VS Code seems a bit more responsive when looking at the schema generation options for a large XML (`content.xml`) without actually choosing one. However, it's not as fast as I would have expected; maybe one of the code actions still tries to read the whole document?

As of now, there is a compile error, it's pretty easy to resolve though.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCodeActions.java
angelozerr: I changed the API, this issue is fixed.

angelozerr:
> So far, this looks pretty good. The API is pretty good, 

Glad you like the API, compeltion resolve should follow the same logic and use some class (ex : DataEntryField class).

> and VS Code seems a bit more responsive when looking at the schema generation options for a large XML (content.xml) without actually choosing one. 

Nice!

> However, it's not as fast as I would have expected; maybe one of the code actions still tries to read the whole document?

You mean when you click on a quickfix? If it that I have not improve something. The real improvement is when code action must be shown, it is very fast and it avoid generating XSD and DTD schema (it avoid the generation of 2 XSD / DTD which takes around 10s * 2 = 20s).

Click on a codeaction takes around 10s to resolve the quickfix. I don't know if we can improve that,we could do that in a separate PR, but generating a XSD/DTD for a large XML file should be not a common usecase.

>As of now, there is a compile error, it's pretty easy to resolve though.

Ooops I forget to commit the file. It should be fixed now.

Thanks @datho7561 for your feedback!

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/BaseCodeActionRequest.java
```diff
@@ -0,0 +1,38 @@
+package org.eclipse.lemminx.services;
```
datho7561, line 1: license header

datho7561 (changes requested):
Just caught a few missing copyright headers/ headers with the wrong year.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLCodeActionSettings.java
```diff
@@ -0,0 +1,36 @@
+/*******************************************************************************
+* Copyright (c) 2019 Red Hat Inc. and others.
```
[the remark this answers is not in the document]
angelozerr: fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/BaseCodeActionRequest.java
angelozerr: fixed

angelozerr:
Thanks @datho7561 for your review!

PR #1225, opened 2022-06-09
Added code action to fix cvc-elt.1.a

JessicaJHee:
issue599

Fixes redhat-developer/vscode-xml#599 

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_elt_1_aCodeAction.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_elt_1_aCodeAction.java
@@ -1,0 +1,109 @@
+/**
+ * Copyright (c) 2022 Red Hat Inc. and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ * Red Hat Inc. - initial API and implementation
+ */
+package org.eclipse.lemminx.extensions.contentmodel.participants.codeactions;
+
+import java.util.List;
+import java.util.Collection;
+import java.util.TreeSet;
+import java.text.Collator;
+import java.util.ArrayList;
+
+import org.eclipse.lemminx.commons.BadLocationException;
+import org.eclipse.lemminx.commons.CodeActionFactory;
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.dom.DOMElement;
+import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.dom.LineIndentInfo;
+import org.eclipse.lemminx.services.extensions.ICodeActionParticipant;
+import org.eclipse.lemminx.services.extensions.IComponentProvider;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lemminx.utils.XMLPositionUtility;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.jsonrpc.CancelChecker;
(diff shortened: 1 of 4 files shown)


JessicaJHee:
At the moment, it takes another code action to match the closing tag with the new element name. Would it be better to do that in one step or leave them as separate?

angelozerr:
> At the moment, it takes another code action to match the closing tag with the new element name. Would it be better to do that in one step or leave them as separate?

Goal of  quickfix (code action) is to fix error. If your quickfix generate code which causes new error, it is little annoying. In other words your code action should update the start tag and end tag  both and not only the start tag.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_elt_1_aCodeAction.java
```diff
@@ -0,0 +1,92 @@
...
+import org.eclipse.lemminx.extensions.contentmodel.model.CMElementDeclaration;
+import org.eclipse.lemminx.extensions.contentmodel.model.ContentModelManager;
+import org.eclipse.lemminx.extensions.contentmodel.model.CMDocument;
+import org.eclipse.lemminx.utils.StringUtils;
+import org.eclipse.lemminx.services.extensions.IPositionRequest;
+
+/**
+ * Code action to find the expected element name defined in the given xsd and
+ * replace it
+ *
```
angelozerr, line 44: Please add a sample.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_elt_1_aCodeAction.java
```diff
@@ -0,0 +1,92 @@
...
+
+		try {
+			DOMNode unexpectedElement = document.getDocumentElement();
+			if (unexpectedElement == null) {
+				return;
+			}
+			String unexpectedElementText = unexpectedElement.getNodeName();
+
+			// get range of text that needs to be replaced
+			int startOffset;
```
angelozerr, line 60: Replace with:

```java
Range diagnosticRange = diagnostic.getRange();
int startOffset = document.offsetAt(diagnosticRange.getStart()) + 1;
DOMNode node = document.findNodeAt(startOffset);
if (node == null || !node.isElement()) {
	return;
}
Range replaceRange = XMLPositionUtility.selectStartTagName(node);
```

 - if I remember we need to use diagnostic range instead of code action range.
 - int startrOffset must be assigneddirectly.
 - replaceRange is computed only if needed.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLModelDiagnosticsTest.java
```diff
@@ -80,6 +83,36 @@ public void xmlModelWithXSDAndNamespace() throws Exception {
...
+				"    </authors>\r\n" + //
+				"</test>";
+		Diagnostic d = d(0, 1, 0, 5, XMLSchemaErrorCode.cvc_elt_1_a);
+		testDiagnosticsFor(xml, d);
+		testCodeActionsFor(xml, d, ca(d, te(0, 1, 0, 5, "root"),
+				te(5, 2, 5, 6, "root")),
+				ca(d, te(0, 1, 0, 5, "root2"),
+						te(5, 2, 5, 6, "root2")));
+	}
+
```
angelozerr, line 115: Could you write a test with xml which have no end tag. By reading your code it seems you take care about that but we need test this usecase.

JessicaJHee: My code currently doesn't add a new end tag if there is no end tag. Should I extend the code action to also insert the correct end tag if it doesn't exist?

angelozerr: No your behavior is perfect, but a test wth this usecase will prevent from NPE or other bug if we change your code. We will sure that we cover all usecases.

It is difficult to create an end tag, because the question is where you create the end tag?

angelozerr: Thanks @JessicaJHee for the test.

PR #1226, opened 2022-06-13
Added code action support to fix error cvc_complex_type_2_4_b

### LINKED ISSUE #1218: CodeAction for cvc-complex-type.2.4.b
Given this XML file:

```xml
<web-app xmlns="http://java.sun.com/xml/ns/j2ee"
    xsi:schemaLocation="http://java.sun.com/xml/ns/j2ee http://java.sun.com/xml/ns/j2ee/web-app_2_4.xsd"
    version="2.4">
    <servlet></servlet>
</web-app>
```

There is an error on servlet:

image

because servlet expects some content. It should be nice to provide a code action which generate the expected content.

JessicaJHee:
issue1218

Fixes #1218 

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
@@ -1,0 +1,120 @@
+/**
+ * Copyright (c) 2022 Red Hat Inc. and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ * Red Hat Inc. - initial API and implementation
+ */
+package org.eclipse.lemminx.extensions.contentmodel.participants.codeactions;
+
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
+
+import org.eclipse.lemminx.commons.BadLocationException;
+import org.eclipse.lemminx.commons.CodeActionFactory;
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.dom.DOMElement;
+import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.extensions.contentmodel.model.CMDocument;
+import org.eclipse.lemminx.extensions.contentmodel.model.CMElementDeclaration;
+import org.eclipse.lemminx.extensions.contentmodel.model.ContentModelManager;
+import org.eclipse.lemminx.extensions.contentmodel.utils.XMLGenerator;
+import org.eclipse.lemminx.services.extensions.codeaction.ICodeActionParticipant;
+import org.eclipse.lemminx.services.extensions.codeaction.ICodeActionRequest;
+import org.eclipse.lemminx.utils.StringUtils;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.Position;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.jsonrpc.CancelChecker;
+
+/**
+ * CodeAction to insert expected child elements within an element
+ * 
+ * Given this XML where the expected child elements are not present as defined
+ * in the XSD:
+ * 
+ * <servlet></servlet>
+ * Error:
+ * Child elements are missing from element:
+ * - servlet
+ *
+ * The following elements are expected:
+ * - description
+ * - display-name
+ * - icon
+ * - servlet-name
+ * 
+ * To fix the error, the code action will suggest inserting the expected
+ * elements inside the parent tag
+ * 
+ */
+
+public class cvc_complex_type_2_4_bCodeAction implements ICodeActionParticipant {
+
+	@Override
+	public void doCodeAction(ICodeActionRequest request, List<CodeAction> codeActions, CancelChecker cancelChecker) {
+		DOMDocument document = request.getDocument();
+		Diagnostic diagnostic = request.getDiagnostic();
+		try {
+			Range diagnosticRange = diagnostic.getRange();
+			int startOffset = document.offsetAt(diagnosticRange.getStart()) + 1;
+			DOMNode node = document.findNodeAt(startOffset);
+
+			if (node == null || !node.isElement()) {
+				return;
+			}
+
+			DOMElement element = (DOMElement) node;
+			int elementOffset = element.getStartTagCloseOffset() + 1;
+			Position childElementPosition = document.positionAt(elementOffset);
+
+			StringBuilder insertText = new StringBuilder();
+
+			XMLGenerator generator = request.getXMLGenerator();
+			ContentModelManager contentModelManager = request.getComponent(ContentModelManager.class);
+
+			// get list of all existing elements
+			Set<String> existingElements = new TreeSet<String>();
+			List<DOMNode> children = element.getChildren();
+			for (DOMNode child : children) {
+				if (child.isElement()) {
+					existingElements.add(child.getNodeName());
+				}
+			}
+
+			for (CMDocument cmDocument : contentModelManager.findCMDocument(element)) {
+				CMElementDeclaration matchesCMElement = cmDocument.findCMElement(element);
+				if (matchesCMElement != null) {
+					insertText.append(generator.generate(matchesCMElement, null, true, true, existingElements));
+				}
+			}
+
+			// When the parent start and end tag is on the same line: add new line, and
+			// indent if needed
+			if (document.positionAt(element.getStartTagCloseOffset()).getLine() == document
+					.positionAt(element.getEndTagOpenOffset()).getLine()) {
+				int lineNum = document.positionAt(element.getStartTagCloseOffset()).getLine();
+				insertText.append(document.getLineIndentInfo(lineNum).getLineDelimiter());
+				insertText.append(StringUtils.getStartWhitespaces(document.lineText(lineNum)));
+			}
+
+			String insertStr = insertText.toString();
+
+			CodeAction insertExpectedElement = CodeActionFactory.insert(
+					"Insert all expected elements", childElementPosition, insertStr,
+					document.getTextDocument(), diagnostic);
+
+			codeActions.add(insertExpectedElement);
+
+		} catch (BadLocationException e) {
+			// do nothing
+		}
+	}
+}
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/XMLGenerator.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/XMLGenerator.java
@@ -65,6 +65,7 @@ public class XMLGenerator {
 		this.whitespacesIndent = whitespacesIndent;
 		this.lineDelimiter = lineDelimiter;
 		this.canSupportSnippets = canSupportSnippets;
+		this.maxLevel = maxLevel;
 	}
 
 	/**
@@ -76,16 +77,53 @@ public class XMLGenerator {
 	 */
 	public String generate(CMElementDeclaration elementDeclaration, String prefix, boolean generateEndTag) {
 		XMLBuilder xml = new XMLBuilder(sharedSettings, whitespacesIndent, lineDelimiter);
-		generate(elementDeclaration, prefix, generateEndTag, 0, 0, xml, new ArrayList<CMElementDeclaration>());
+		generate(elementDeclaration, prefix, generateEndTag, false, 0, 0, xml, new ArrayList<CMElementDeclaration>(),
+				new ArrayList<String>());
 		if (canSupportSnippets) {
 			xml.addContent(SnippetsBuilder.tabstops(0)); // "$0"
 		}
 		return xml.toString();
 	}
 
-	private int generate(CMElementDeclaration elementDeclaration, String prefix, boolean generateEndTag, int level,
-			int snippetIndex, XMLBuilder xml, List<CMElementDeclaration> generatedElements) {
-		if (generatedElements.contains(elementDeclaration)) {
+	/**
+	 * Returns the XML generated from the given element declaration given option to
+	 * generate only children.
+	 * 
+	 * @param elementDeclaration
+	 * @param prefix
+	 * @param generateEndTag
+	 * @param generateOnlyChildren true if only is children of the given element is
+	 *                             to be generated
+	 * @param existingElementNames a collection of strings of element names that
+	 *                             already exist in the XML
+	 * @return the XML generated from the given element declaration given option to
+	 *         generate only children.
+	 */
+	public String generate(CMElementDeclaration elementDeclaration, String prefix, boolean generateEndTag,
+			boolean generateOnlyChildren, Collection<String> existingElementNames) {
+		XMLBuilder xml = new XMLBuilder(sharedSettings, whitespacesIndent, lineDelimiter);
+		generate(elementDeclaration, prefix, generateEndTag, generateOnlyChildren, 0, 0, xml,
+				new ArrayList<CMElementDeclaration>(), existingElementNames);
+		if (canSupportSnippets) {
+			xml.addContent(SnippetsBuilder.tabstops(0)); // "$0"
+		}
+		return xml.toString();
+	}
+
+	private int generate(CMElementDeclaration elementDeclaration, String prefix, boolean generateEndTag,
+			boolean generateOnlyChildren, int level,
+			int snippetIndex, XMLBuilder xml, List<CMElementDeclaration> generatedElements,
+			Collection<String> existingElementNames) {
+		if (generateOnlyChildren) {
+			Collection<CMElementDeclaration> childLElements = elementDeclaration.getElements();
+			for (CMElementDeclaration child : childLElements) {
+				snippetIndex = generate(child, prefix, true, false, level + 1, snippetIndex, xml,
+						generatedElements, existingElementNames);
+			}
+			return snippetIndex;
+		}
+		if (generatedElements.contains(elementDeclaration)
+				|| existingElementNames.contains(elementDeclaration.getName())) {
 			return snippetIndex;
 		}
 		boolean autoCloseTags = this.autoCloseTags && generateEndTag;
@@ -102,10 +140,11 @@ public class XMLGenerator {
 		Collection<CMElementDeclaration> children = elementDeclaration.getElements();
 		if (children.size() > 0) {
 			xml.closeStartElement();
-			if ((level > maxLevel)) {
+			if ((level < maxLevel)) {
 				level++;
 				for (CMElementDeclaration child : children) {
-					snippetIndex = generate(child, prefix, true, level, snippetIndex, xml, generatedElements);
+					snippetIndex = generate(child, prefix, true, false, level, snippetIndex, xml,
+							generatedElements, existingElementNames);
 				}
 				level--;
 				xml.linefeed();
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/CodeActionRequest.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/CodeActionRequest.java
@@ -17,6 +17,14 @@ import org.eclipse.lemminx.services.extensions.codeaction.ICodeActionRequest;
 import org.eclipse.lemminx.settings.SharedSettings;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.Range;
+import org.eclipse.lemminx.commons.BadLocationException;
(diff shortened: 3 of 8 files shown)


org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
```diff
@@ -1362,6 +1361,9 @@ public void diagnosticsWithWebApp() throws BadLocationException {
 				"</web-app>";
 		Diagnostic diagnostic = d(4, 5, 4, 12, XMLSchemaErrorCode.cvc_complex_type_2_4_b);
 		testDiagnosticsFor(xml, "src/test/resources/catalogs/catalog-web-app.xml", diagnostic);
+		testCodeActionsFor(xml, diagnostic, "src/test/resources/catalogs/catalog-web-app.xml",
```
JessicaJHee, line 1364: I'm not sure why the `testCodeActionsFor` test returns no code action when it works in the demo. (Expected [1] but was [0])

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
```diff
@@ -0,0 +1,117 @@
...
+ * elements inside the parent tag
+ * 
+ */
+public class cvc_complex_type_2_4_bCodeAction implements ICodeActionParticipant {
+
+	private static final Pattern ELEMENT_EXTRACTOR = Pattern
+			.compile("(?<!Child elements are missing from element:\\s)\\s-\\s([A-Za-z-_.]+)");
+
+	@Override
+	public void doCodeAction(Diagnostic diagnostic, Range range, DOMDocument document, List<CodeAction> codeActions,
```
angelozerr, line 64: Please rebase your PR from master and change the method signature with ICodeActionRequest as parameter.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
```diff
@@ -0,0 +1,117 @@
...
+			int startOffset = document.offsetAt(diagnosticRange.getStart()) + 1;
+			DOMNode node = document.findNodeAt(startOffset);
+			if (node == null || !node.isElement()) {
+				return;
+			}
+			DOMElement element = (DOMElement) node;
+			int elementOffset = element.getEndTagOpenOffset();
+			Position childElemenPosition = document.positionAt(elementOffset);
+
+			Collection<String> expectedElements = extractElements(diagnostic.getMessage());
```
angelozerr, line 79: There are several problems with your current implementation:

 * you hard code the line seperator with `\n`, but you must use the line seperator from the XML document. You can do that with `String delimiter = document.getTextDocument().lineDelimiter(0);`
 * you don't take care of required attributes
 * you don't take care of required child element of an expected element.
 * you are using String.concat but it is better to use StringBuilder#append

To fix those all problems, I think we should use CMElementDeclaration with XMLGenerator like we did for completion. I suggest that you try it if it is not easy to use this strategy we will use your strategy with extract of expected elements.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
```diff
@@ -0,0 +1,138 @@
...
+	 */
+	private static Collection<CMElementDeclaration> getChildElements(Collection<CMElementDeclaration> cmElements,
+			XMLGenerator generator)
+			throws Exception {
+		List<CMElementDeclaration> possibleElements = new ArrayList<>();
+
+		for (CMElementDeclaration cmElement : cmElements) {
+			Collection<CMElementDeclaration> childElements = cmElement.getElements();
+			if (childElements.size() != 0) {
+				possibleElements.addAll(childElements);
```
JessicaJHee, line 129: I initially created this function to recursively get the element declaration of all the elements and its children, but at the moment it doesn't work as I wanted and just returns the top level elements. It seems like generate() takes care of the children, so I'm not sure if I need to keep trying to get this function to work recursively.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
```diff
@@ -0,0 +1,138 @@
...
+			ContentModelManager contentModelManager = request.getComponent(ContentModelManager.class);
+
+			for (CMDocument cmDocument : contentModelManager.findCMDocument(element)) {
+				Collection<CMElementDeclaration> cmElements = cmDocument.getElements();
+
+				Collection<CMElementDeclaration> childElements;
+				try {
+					childElements = getChildElements(cmElements, generator);
+					for (CMElementDeclaration expectedElement : childElements) {
+						insertText = insertText.concat(generator.generate(expectedElement, null, 1, true));
```
JessicaJHee, line 95: To get generate() to generate the children in the test case I made, I have arbitrarily set the level to 1 for now. I'm wondering how should I do this in a better way.

JessicaJHee:
issue1218_new
This is how it works for now. I also need to figure out why I am unable to find the CMDocument with the original web-app example (also doesn't work for cvc-elt-1 code action).

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
JessicaJHee: based on our meeting today, I will move this iterator to the XML generator.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
```diff
@@ -0,0 +1,112 @@
...
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ * Red Hat Inc. - initial API and implementation
+ */
+package org.eclipse.lemminx.extensions.contentmodel.participants.codeactions;
+
+import java.text.Collator;
```
angelozerr, line 15: Remove unused java.text.Collator

JessicaJHee:
I still had a few questions about this issue: 

1.
> It is a good question and XMLGenerator must be improved and supports Integer.MAX_VALUE. When XMLGenerator will be initailized with that, it means that any level is allowed.
* I'm not sure what you mean by this - how should we set this value?

2. The tests I have added are failing because it is expecting an ArrayList but is getting a SingletonList, but everything else is the same. I'm not sure how to fix this.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
```diff
@@ -197,6 +197,28 @@ public void cvc_complex_type_3_2_2() throws Exception {
...
 	}
 
+	@Test
+	public void cvc_complex_type_2_4_bCodeAction_basic() throws Exception {
+		String xml = "<root xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"\r\n" + //
+				"    xsi:noNamespaceSchemaLocation=\"src/test/resources/xsd/unique.xsd\">\r\n" + //
+				"</root>";
+		Diagnostic d = d(0, 1, 0, 5, XMLSchemaErrorCode.cvc_complex_type_2_4_b);
+		testDiagnosticsFor(xml, d);
+		testCodeActionsFor(xml, d, ca(d, te(2, 0, 2, 0, "\r\n <authors>\r\n  <author></author>\r\n </authors>")));
```
angelozerr, line 207: The testCodeActionsFor that you are using use tabs instead of spaces https://github.com/eclipse/lemminx/blob/9e879e3de0c363e12651f3aa0cdc82a050d79143/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/XMLAssert.java#L644 So your test must use tabs instead of spaces:

```java
testCodeActionsFor(xml, d, ca(d, te(2, 0, 2, 0, //
				"\r\n" + //
						"\t<authors>\r\n" + //
						"\t\t<author></author>\r\n" + //
						"\t</authors>")));
```

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
```diff
@@ -197,6 +197,28 @@ public void cvc_complex_type_3_2_2() throws Exception {
...
+
+	@Test
+	public void cvc_complex_type_2_4_bCodeAction_child() throws Exception {
+		String xml = "<root xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"\r\n" + //
+				"    xsi:noNamespaceSchemaLocation=\"src/test/resources/xsd/unique.xsd\">\r\n" + //
+				"    <authors></authors>\r\n" + //
+				"</root>";
+		Diagnostic d = d(2, 5, 2, 12, XMLSchemaErrorCode.cvc_complex_type_2_4_b);
+		testDiagnosticsFor(xml, d);
+		testCodeActionsFor(xml, d, ca(d, te(2, 13, 2, 13, "\r\n     [author][/author]")));
```
angelozerr, line 218: Use tabs instead of spaces.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaDiagnosticsTest.java
```diff
@@ -1362,6 +1384,9 @@ public void diagnosticsWithWebApp() throws BadLocationException {
 				"</web-app>";
 		Diagnostic diagnostic = d(4, 5, 4, 12, XMLSchemaErrorCode.cvc_complex_type_2_4_b);
 		testDiagnosticsFor(xml, "src/test/resources/catalogs/catalog-web-app.xml", diagnostic);
+		testCodeActionsFor(xml, diagnostic, "src/test/resources/catalogs/catalog-web-app.xml",
+				ca(diagnostic, te(4, 13, 4, 13,
+						"\r\n     <description></description>\r\n     <display-name></display-name>\r\n     <icon>\r\n      <small-icon></small-icon>\r\n      <large-icon></large-icon>\r\n     </icon>\r\n     <servlet-name></servlet-name>\r\n     <servlet-class></servlet-class>\r\n     <jsp-file></jsp-file>\r\n     <init-param>\r\n      <description></description>\r\n      <param-name></param-name>\r\n      <param-value></param-value>\r\n     </init-param>\r\n     <load-on-startup></load-on-startup>\r\n     <run-as>\r\n      <description></description>\r\n      <role-name></role-name>\r\n     </run-as>\r\n     <security-role-ref>\r\n      <description></description>\r\n      <role-name></role-name>\r\n      <role-link></role-link>\r\n     </security-role-ref>")));
```
angelozerr, line 1389: Use tabs instead of spaces + indent correctly your expected xml.

angelozerr:
> The tests I have added are failing because it is expecting an ArrayList but is getting a SingletonList, but everything else is the same. I'm not sure how to fix this.

Use tabs instead of spaces, see https://github.com/eclipse/lemminx/pull/1226/files#r901367520

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/CodeActionRequest.java
```diff
@@ -47,4 +59,23 @@ public Range getRange() {
 		return range;
 	}
 
+
+	public XMLGenerator getXMLGenerator() throws BadLocationException {
+		if (generator == null) {
+			generator = new XMLGenerator(getSharedSettings(), getSharedSettings().getCompletionSettings().isAutoCloseTags(),
+					getLineIndentInfo().getWhitespacesIndent(), getLineIndentInfo().getLineDelimiter(),
+					false, 0);
```
angelozerr, line 67: Intilaize XMLGenerator with any level by using Integer.MAX_VALUE:

```
generator = new XMLGenerator(getSharedSettings(), getSharedSettings().getCompletionSettings().isAutoCloseTags(),
					getLineIndentInfo().getWhitespacesIndent(), getLineIndentInfo().getLineDelimiter(),
					false, Integer.MAX_VALUE);

```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/XMLGenerator.java
```diff
@@ -76,15 +76,35 @@ public XMLGenerator(SharedSettings sharedSettings, boolean autoCloseTags, String
...
-		generate(elementDeclaration, prefix, generateEndTag, 0, 0, xml, new ArrayList<CMElementDeclaration>());
+		generate(elementDeclaration, prefix, generateEndTag, false, 0, 0, xml, new ArrayList<CMElementDeclaration>());
 		if (canSupportSnippets) {
 			xml.addContent(SnippetsBuilder.tabstops(0)); // "$0"
 		}
 		return xml.toString();
 	}
 
-	private int generate(CMElementDeclaration elementDeclaration, String prefix, boolean generateEndTag, int level,
+	public String generate(CMElementDeclaration elementDeclaration, String prefix, boolean generateEndTag,
```
angelozerr, line 86: Please add Javadoc. I would rename `generateChildren` to `generateOnlyChildren`

angelozerr:
> I'm not sure what you mean by this - how should we set this value?

See my comment at https://github.com/eclipse/lemminx/pull/1226/files#r901386675

JessicaJHee:
24b
At the moment generating a new xml block will always add a new line before it, should that be changed?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
```diff
@@ -0,0 +1,116 @@
...
+			Position childElementPosition = document.positionAt(elementOffset);
+
+			StringBuilder insertText = new StringBuilder();
+
+			XMLGenerator generator = request.getXMLGenerator();
+			ContentModelManager contentModelManager = request.getComponent(ContentModelManager.class);
+
+			// get list of all existing elements
+			Collection<String> existingElements = new TreeSet<String>(Collator.getInstance());
+			for (DOMNode root : document.getRoots()) {
```
angelozerr, line 86: Here you collect existing elements for any level, but you need to collect it for the children of the element which have an error

```java
Set<String> existingElements = new TreeSet<String>();
List<DOMNode> children = element.getChildren();
for (DOMNode child : children) {
	if (child.isElement()) {
		existingElements .add(child.getNodeName());
	}
}
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
```diff
@@ -0,0 +1,116 @@
...
+				Collection<DOMNode> existingChilren = root.getChildren();
+				for (DOMNode existingChild : existingChilren) {
+					existingElements.add(existingChild.getNodeName());
+				}
+			}
+
+			for (CMDocument cmDocument : contentModelManager.findCMDocument(element)) {
+				CMElementDeclaration matchesCMElement = cmDocument.findCMElement(element);
+				if (matchesCMElement != null) {
+					insertText.append(generator.generate(matchesCMElement, null, true, true, existingElements));
```
angelozerr, line 97: I think it is not XML generator which must check that, but you must check that here:

`if (matchesCMElement != null && !existingElements.contains(matchesCMElement .getName())) {`

angelozerr:
> At the moment generating a new xml block will always add a new line before it, should that be changed?

Do you know why? I think it is litlle annoying.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
JessicaJHee: I check in the XML generator because the `matchesCMElement` that gets passed in to `generate()` is the parent ('root'), and we need to check it against the children of root, which is extracted through an iterator inside the generate function. If we want to do the checks outside the generator, I believe we would need to bring the iterator outside and set `generateOnlyChildren` to false (or just remove that option).

JessicaJHee:
> Do you know why? I think it is litlle annoying.

https://github.com/eclipse/lemminx/blob/3ed05cb913f471c582f6de498211a5fc104dbca9/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/XMLGenerator.java#L94

The generate function adds a new line each time if the level is > 0. I will try to see if there is a way to not add a new line under certain conditions without interfering with other functionalities.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
angelozerr: Thanks for your explanation, you are right. The check must be done in XMLGenerator but the existingElements should be filled like I suggested (just for the first level). The check with existingElements must be done only for the first level and nor for children.

JessicaJHee:
I tried a bunch of conditions to try to make the generator not add new lines at the beginning, but it either doesn't work because the generate function is called recursively so it looses information about the previous caller or messes up other functionalities (like completion). Since the unwanted new line is added each time, I just remove it from the string. I believe there are ways to do this but this is the simplest and works. 

```
CodeAction insertExpectedElement = CodeActionFactory.insert(
					"Insert all expected elements", childElementPosition, insertStr.substring(1),
```

Would it be okay?

24b

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
```diff
@@ -0,0 +1,113 @@
...
+			Range diagnosticRange = diagnostic.getRange();
+			int startOffset = document.offsetAt(diagnosticRange.getStart()) + 1;
+			DOMNode node = document.findNodeAt(startOffset);
+
+			if (node == null || !node.isElement()) {
+				return;
+			}
+
+			DOMElement element = (DOMElement) node;
+			int elementOffset = element.getEndTagOpenOffset();
```
angelozerr, line 74: Here you get the end tag open offset:

```xml
<foo>

|</foo>
```

But it is wrong. Imagine you have this XML:


```xml
<foo>

                         |</foo>
```

It will not insert to the correct position.

I think the correct position is 

`int elementOffset = element.getStartTagCloseOffset();`

```xml
<foo>|

</foo>
```

By using this offset, you should generate the `\r\n`  without problem I think (I hope).

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/cvc_complex_type_2_4_bCodeAction.java
```diff
@@ -0,0 +1,117 @@
...
+
+			String insertStr = insertText.toString();
+
+			CodeAction insertExpectedElement = CodeActionFactory.insert(
+					"Insert all expected elements", childElementPosition, insertStr,
+					document.getTextDocument(), diagnostic);
+
+			codeActions.add(insertExpectedElement);
+
+		} catch (
```
angelozerr, line 110: Please remove this empty line.

JessicaJHee:
24b

angelozerr:
@JessicaJHee I played with your PR and it works almost great, but I think it is good to merge it, we could improve it in an another PR.

For instance if you have this xml file:

```xml
<root
	xsi:noNamespaceSchemaLocation="unique.xsd">
	<authors>	</authors>
	<assistant></assistant>
</root>
```
authors have some spaces, after processing your code action, you have:

```xml
<root
	xsi:noNamespaceSchemaLocation="unique.xsd">
	<authors>
		<author></author>
		</authors>
	<assistant></assistant>
</root>
```

The end tag `</authors>` is not properly indented, but it is just a matter of formatting. One idea that I have is to use perhaps the experimental formatter to do that or improving again XMLGenerator?

An another improvememnt that we could have too is to generate only required xs:element and not the whole xs:element.

angelozerr:
Great job @JessicaJHee !

Issue #594, opened 2019-11-15
Improve XML formatter (option to have Tags in the same line)


rwx-777:
Hello, again we spoke earlier already [here](https://github.com/redhat-developer/vscode-xml/issues/200). As requested here we will be listing some examples by Screenshots and the use case of our improvements.

When viewing a larger XML-File with your Tool it always puts some Tags in a New Line.
We have edited this to not be the case anymore.
It would be very appreciated if we could have the option of using this.

Here are some before and after screenshots of an XML-File.

Before our Improvements:

![test-vorher](https://user-images.githubusercontent.com/26744048/68936460-e6fe3980-079a-11ea-82b6-6b5b77be4b93.png)

After our Improvements:

![test-nachher](https://user-images.githubusercontent.com/26744048/68936503-009f8100-079b-11ea-89c7-0a92a5a8adff.png)

Here is the link to the Github Repo where we have stored our changed version:
[https://github.com/rwx-777/lsp4xml-AtSec-Edition](url)

Use Case:
- Tags can appear in the same line

Thank you for your time and we are looking forward to what is to come. 

Best regards,

Joshua

angelozerr:
Before implementing this feature, I would like to clarify the feature about formatting. I have the impression that is a `xml:space` support formatting that it should be done.

Here 3 links why I suppose it's about xml:space:

 * http://www.xmlplease.com/xml/xmlspace/
 * http://tei-l.970651.n3.nabble.com/oXygen-and-the-quot-Format-and-Indent-quot-function-tt2346233.html#a2346236
 * https://www.oxygenxml.com/forum/topic10259.html#p28680

@rwx-777 could you explain us how to you configure this format settings in Oxygen?

enxio:
Hi Angelo, I work with the OP. I am not sure that it's about xml:space what Oxygen did different. Oxygen (at least the version I was using) had three settings:
- A list of elements where space would be preserved
- A list of elements where space was handled "defalt" i.e. normalized.
- A list of elements with mixed content - no linebreaks are inserted between adjacent elements
Oxygen's formatter tried to inline the tags more or less automatically, and almost always produced something that was easy to read. Basically what our formatting hack does.

angelozerr:
> Oxygen's formatter tried to inline the tags more or less automatically, and almost always produced something that was easy to read. Basically what our formatting hack does.

Thanks for your feedback @enxio I would like to understand the Oxygen rule to try to format automatically when it's possible.

[...]

PR #1232, opened 2022-06-21
New formatter

enxio:


enxio:
Sorry about that. Please ignore.

angelozerr:
@enxio I remember that we have some discussion about formatting text at https://github.com/eclipse/lemminx/issues/594

The new experimental formatter should take care of your usecase, but please note that it is not perfect, so don't hesitate to create issues if you find some trouble. Thanks.

enxio:
@angelozerr this is much appreciated! 

BTW: I was able to merge my "old" formatter-hack into the new structure. So the migration to the new formatter should be even less painful.

Thanks and sorry again for this PR, sometimes I just get lost in repos and forks...

enxio:
Yikes, I did not mean to imply that using the new formatter would be painful. In fact I am looking forward to trying it out. I guess it'll be in the next XML extension in a couple of days...

PR #1252, opened 2022-07-14
Support xsiSchemaLocationSplit setting with experimental formatter

### LINKED ISSUE #1246: Support `xml.format.xsiSchemaLocationSplit` setting with experimental formatter
See https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md#xmlformatxsischemalocationsplit

JessicaJHee:
Fixes #1246 

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMDocument.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMDocument.java
@@ -321,7 +321,8 @@ public class DOMDocument extends DOMNode implements Document {
 
 	private static SchemaLocation createSchemaLocation(DOMNode root, String schemaInstancePrefix) {
 		DOMAttr attr = root.getAttributeNode(getPrefixedName(schemaInstancePrefix, "schemaLocation"));
-		if (attr == null) {
+		// Check that the attribute and the attribute value of xsi:schemaLocation is not null
+		if (attr == null || attr.getNodeAttrValue() == null) {
 			return null;
 		}
 		return new SchemaLocation(attr);
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
@@ -19,8 +19,11 @@ import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.extensions.xsi.XSISchemaModel;
 import org.eclipse.lemminx.extensions.xsi.settings.XSISchemaLocationSplit;
 import org.eclipse.lemminx.services.extensions.format.IFormatterParticipant;
+import org.eclipse.lemminx.services.format.XMLFormatterDocumentNew;
 import org.eclipse.lemminx.settings.XMLFormattingOptions;
+import org.eclipse.lemminx.utils.StringUtils;
 import org.eclipse.lemminx.utils.XMLBuilder;
+import org.eclipse.lsp4j.TextEdit;
 
 /**
  * Formatter participant implementation to format xsi:schemaLocation attribute
@@ -155,4 +158,78 @@ public class XSIFormatterParticipant implements IFormatterParticipant {
 		return indent.toString();
 	}
 
+	@Override
+	public boolean formatAttributeValue(DOMAttr attr, XMLFormatterDocumentNew formatterDocument,
+			int indentLevel, XMLFormattingOptions formattingOptions, List<TextEdit> edits) {
+
+		XSISchemaLocationSplit split = XSISchemaLocationSplit.getSplit(formattingOptions);
+
+		if (split == XSISchemaLocationSplit.none || !XSISchemaModel.isXSISchemaLocationAttr(attr.getName(), attr)
+				|| getFirstContentOffset(attr.getOriginalValue()) == -1) {
+			return false;
+		}
+
+		int attrValueStart = attr.getNodeAttrValue().getStart();
+		// Remove extra spaces between start of xsi:schemaLocation attribute value quote
+		// and actual value
+		formatterDocument.removeLeftSpaces(attrValueStart + 1, // <... xsi:schemaLocation="| value"
+				// <... xsi:schemaLocation=" |value"
+				attrValueStart + getFirstContentOffset(attr.getOriginalValue()), edits);
+
+		int tabSize = formattingOptions.getTabSize();
+		int indentSpaceOffset;
+		int startOfLineOffset = formatterDocument.getLineAtOffset(attr.getOwnerElement().getStart());
+
+		if (formattingOptions.isSplitAttributes()) {
+			indentSpaceOffset = (attrValueStart + 1) - attr.getNodeAttrName().getStart()
+					+ formattingOptions.getSplitAttributesIndentSize() * tabSize;
+		} else if (formattingOptions.isPreserveAttributeLineBreaks()) {
+			indentSpaceOffset = attrValueStart
+					- formatterDocument.getOffsetWithPreserveLineBreaks(startOfLineOffset, attrValueStart, tabSize, formattingOptions.isInsertSpaces());
+		} else {
+			indentSpaceOffset = formatterDocument.getNormalizedLength(startOfLineOffset, attrValueStart + 1)
+					- startOfLineOffset;
+		}
+
+		int lineFeed = split == XSISchemaLocationSplit.onElement ? 1 : 2;
+		int locationNum = 1;
+		String attrValue = attr.getOriginalValue();
+
+		for (int i = 0; i < attrValue.length(); i++) {
+			int from = formatterDocument.getLeftWhitespacesOffset(attrValueStart, attrValueStart + i + 1);
+			if (Character.isWhitespace(attrValue.charAt(i)) && !Character.isWhitespace(attrValue.charAt(i + 1))
+					&& !StringUtils.isQuote(attrValue.charAt(from - attrValueStart))) {
+				// Insert newline and indent where required based on setting
+				if (locationNum % lineFeed == 0) {
+					formatterDocument.replaceLeftSpacesWithIndentationWithOffsetSpaces(indentSpaceOffset,
+							attrValueStart + i + 1,
+							true, edits);
+				} else {
+					formatterDocument.replaceLeftSpacesWithOneSpace(indentSpaceOffset, attrValueStart + i + 1, edits);
+				}
+				locationNum++;
+			}
+		}
+		return true;
+	}
+
+	/**
+	 * Returns the offset from opening quote to first non-whitespace character of an
+	 * attribute value
+	 *
+	 * @param originalValue
+	 * @return offset from opening quote to first non-whitespace character of an
+	 *         attribute value
+	 */
+	private static int getFirstContentOffset(String originalValue) {
+		if (originalValue == null) {
+			return -1;
+		}
+		for (int i = 1; i < originalValue.length(); i++) {
+			if (!Character.isWhitespace(originalValue.charAt(i)) && !StringUtils.isQuote(originalValue.charAt(i))) {
+				return i;
+			}
+		}
+		return -1;
+	}
 }
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/format/IFormatterParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/format/IFormatterParticipant.java
@@ -12,12 +12,17 @@
 *******************************************************************************/
 package org.eclipse.lemminx.services.extensions.format;
 
+import java.util.List;
+
 import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.services.format.FormatElementCategory;
+import org.eclipse.lemminx.services.format.XMLFormatterDocumentNew;
 import org.eclipse.lemminx.services.format.XMLFormattingConstraints;
 import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lemminx.settings.XMLFormattingOptions;
 import org.eclipse.lemminx.utils.XMLBuilder;
+import org.eclipse.lsp4j.TextEdit;
 
 /**
  * XML formatter participant.
@@ -45,6 +50,21 @@ public interface IFormatterParticipant {
 	 */
 	default boolean formatAttributeValue(String name, String valueWithoutQuote, Character quote, DOMAttr attr,
 			XMLBuilder xml) {
+		return false;
+	}
+
+	/**
+	 * Format the given attribute value.
+	 * 
+	 * @param formatterDocument the formatter document
+	 * @param indentLevel       the current indent level of the attribute
+	 * @param formattingOptions the formatting options that contain the settings
+	 * @param attr              the attribute node
+	 * @param edits             the text edit list
+	 * @return true if the given attribute can be formatted and false otherwise.
+	 */
+	default boolean formatAttributeValue(DOMAttr attr, XMLFormatterDocumentNew formatterDocument,
+			int indentLevel, XMLFormattingOptions formattingOptions, List<TextEdit> edits) {
 		return false;
 	}
 
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
@@ -17,6 +17,7 @@ import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lemminx.settings.EnforceQuoteStyle;
 import org.eclipse.lemminx.utils.StringUtils;
+import org.eclipse.lemminx.settings.XMLFormattingOptions;
 
 /**
  * DOM attribute formatter.
@@ -78,6 +79,7 @@ public class DOMAttributeFormatter {
 				int attrValueStart = attr.getNodeAttrValue().getStart(); // <foo attr = |""
 				removeLeftSpaces(delimiterOffset, attrValueStart, edits);
 			}
+			formatterDocument.formatAttributeValue(attr, formatterDocument, parentConstraints.getIndentLevel(), getFormattingSettings(), edits);
 		}
 
 		// replace current quote with preferred quote in case of attribute value
@@ -118,6 +120,10 @@ public class DOMAttributeFormatter {
 
 	private int getSplitAttributesIndentSize() {
 		return formatterDocument.getSharedSettings().getFormattingSettings().getSplitAttributesIndentSize();
+	}
+
+	private XMLFormattingOptions getFormattingSettings() {
+		return formatterDocument.getSharedSettings().getFormattingSettings();
 	}
 
 	boolean isPreserveAttributeLineBreaks() {
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/XMLFormatterDocumentNew.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/XMLFormatterDocumentNew.java
@@ -19,6 +19,7 @@ import java.util.logging.Logger;
 
 import org.eclipse.lemminx.commons.BadLocationException;
 import org.eclipse.lemminx.commons.TextDocument;
+import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMDocumentType;
 import org.eclipse.lemminx.dom.DOMElement;
@@ -27,6 +28,8 @@ import org.eclipse.lemminx.dom.DOMProcessingInstruction;
 import org.eclipse.lemminx.dom.DOMText;
 import org.eclipse.lemminx.services.extensions.format.IFormatterParticipant;
 import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lemminx.settings.XMLFormattingOptions;
+import org.eclipse.lemminx.utils.StringUtils;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.TextEdit;
@@ -304,11 +307,29 @@ public class XMLFormatterDocumentNew {
 		}
 	}
 
+	public void formatAttributeValue(DOMAttr attr, XMLFormatterDocumentNew formatterDocument, int indentLevel,
+			XMLFormattingOptions formattingOptions, List<TextEdit> edits) {
+		if (formatterParticipants != null) {
+			for (IFormatterParticipant formatterParticipant : formatterParticipants) {
+				try {
+					if (formatterParticipant.formatAttributeValue(attr, formatterDocument, indentLevel,
+							formattingOptions, edits)) {
+						return;
+					}
+				} catch (Exception e) {
+					LOGGER.log(Level.SEVERE,
+							"Error while processing format attributes for the participant '"
+									+ formatterParticipant.getClass().getName() + "'.", e);
+				}
+			}
+		}
+	}
+
 	void removeLeftSpaces(int to, List<TextEdit> edits) {
 		replaceLeftSpacesWith(to, "", edits);
 	}
 
-	void removeLeftSpaces(int from, int to, List<TextEdit> edits) {
+	public void removeLeftSpaces(int from, int to, List<TextEdit> edits) {
 		replaceLeftSpacesWith(from, to, "", edits);
 	}
 
@@ -316,7 +337,7 @@ public class XMLFormatterDocumentNew {
 		replaceLeftSpacesWith(to, " ", edits);
 	}
 
-	void replaceLeftSpacesWithOneSpace(int from, int to, List<TextEdit> edits) {
+	public void replaceLeftSpacesWithOneSpace(int from, int to, List<TextEdit> edits) {
 		replaceLeftSpacesWith(from, to, " ", edits);
 	}
 
@@ -329,11 +350,11 @@ public class XMLFormatterDocumentNew {
 		createTextEditIfNeeded(from, to, replacement, edits);
 	}
 
-	void replaceQuoteWithPreferred(int from, int to, String replacement, List<TextEdit> edits){
+	void replaceQuoteWithPreferred(int from, int to, String replacement, List<TextEdit> edits) {
 		createTextEditIfNeeded(from, to, replacement, edits);
 	}
 
-	private int getLeftWhitespacesOffset(int leftLimit, int to) {
+	public int getLeftWhitespacesOffset(int leftLimit, int to) {
 		String text = textDocument.getText();
 		int from = leftLimit != -1 ? leftLimit : to - 1;
 		int limit = leftLimit != -1 ? leftLimit : 0;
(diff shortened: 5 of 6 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
```diff
@@ -75,6 +84,10 @@ public void formatAttribute(DOMAttr attr, int prevOffset, boolean singleAttribut
 				// 2.2 Remove extra spaces between delimiter and start of attribute value
 				int attrValueStart = attr.getNodeAttrValue().getStart(); // <foo attr = |""
 				removeLeftSpaces(delimiterOffset, attrValueStart, edits);
+				// 2.3 Remove extra spaces between start of attribute value quote and actual value
+				//removeLeftSpaces(attrValueStart + 1, // <foo attr = "|"
```
JessicaJHee, line 88: This removes the space between the first quote and the start of the actual attribute value. This is commented out for now because if it's left in, it causes the merge sort to mess up and override all the edits (that are not overlapping initially) to the same value (causing overlapping edits) and I'm not sure why.

JessicaJHee:
I believe some of the tests require this setting to get the correct result:
https://github.com/eclipse/lemminx/blob/043e5e1a8ecabe2a66b335e3d3b80489b4d27c24/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/xsi/XSIFormatterExperimentalTest.java#L175

This will prevent the empty content from getting deleted like here:  https://github.com/eclipse/lemminx/blob/043e5e1a8ecabe2a66b335e3d3b80489b4d27c24/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/xsi/XSIFormatterExperimentalTest.java#L165

angelozerr:
@JessicaJHee for this kind of formatting you need to do that with an implementation of IformmatingParticipant. I suggest that you see how old formatter uses this participant. You woll need to add or update the signature of the participant to support edits

angelozerr:
@JessicaJHee if you think it is better to work on another settings before doing this pr dont hesîtate to do that

JessicaJHee:
> @JessicaJHee for this kind of formatting you need to do that with an implementation of IformmatingParticipant. I suggest that you see how old formatter uses this participant. You woll need to add or update the signature of the participant to support edits

I didn't use this because I thought for the new formatter, the idea is to avoid using XMLBuilder by adding/removing spaces instead of rewriting the xml content? But if my understanding is wrong, I will try again and use XSIFormatterParticipant/IFormatterParticipant.

rgrunber (commented):
I noticed that formatting as `onElement` and then back to `onPair` doesn't work.

rgrunber:
Whoops, just saw Angelo's other comments. My guess is you need to go into `DOMAttributeFormatter#formatAttribute(..)` and instead of plugging your own code, set up the `IFormatterParticipant`s being contributed and get the proposed text edits of each one. You would be calling some new (default) method of `IFormatterParticipant` that is returning a text edit (and has no XMLBuilder).

With that in place the task just becomes implementing the new method in `XSIFormatterParticipant` based on the old logic.

@angelozerr feel free to correct me if any of this is wrong.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMDocument.java
```diff
@@ -321,7 +321,7 @@ public boolean isBeforeProlog(int offset) {
 
 	private static SchemaLocation createSchemaLocation(DOMNode root, String schemaInstancePrefix) {
 		DOMAttr attr = root.getAttributeNode(getPrefixedName(schemaInstancePrefix, "schemaLocation"));
-		if (attr == null) {
+		if (attr == null || attr.getNodeAttrValue()==null) {
```
angelozerr, line 324: Could you add a  moment with the two useczses.

Have uou à test with yiur new usecase?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMAttr.java
```diff
@@ -459,4 +463,16 @@ public int getDelimiterOffset() {
 		return delimiter;
 	}
 
+	public int getFirstContentOffset(String originalValue) {
```
angelozerr, line 466: Remove original Value as parameyer since you can zvcess with field

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMAttr.java
```diff
@@ -459,4 +463,16 @@ public int getDelimiterOffset() {
 		return delimiter;
 	}
 
+	public int getFirstContentOffset(String originalValue) {
+		if (originalValue==null){
+			return 0;
+		}
```
angelozerr, line 469: Return -1

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMAttr.java
```diff
@@ -459,4 +463,16 @@ public int getDelimiterOffset() {
...
+	public int getFirstContentOffset(String originalValue) {
+		if (originalValue==null){
+			return 0;
+		}
+		Pattern pattern = Pattern.compile("((\")|(\'))\\s*(.*)");
+		Matcher matcher = pattern.matcher(originalValue);
+		if (matcher.find()) {
+			return matcher.start(4);
+		}
+		return 0;
```
angelozerr, line 475: Return -1

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMAttr.java
angelozerr, line 466: Afd javadoc

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -155,4 +158,58 @@ public String getCurrentLineIndent(XMLBuilder xml, XMLFormattingOptions formatti
 		return indent.toString();
 	}
 
+	@Override
+	public boolean formatAttributeValueNew(XMLFormatterDocumentNew formatterDocument,
+			int indentLevel, XMLFormattingOptions formattingOptions, DOMAttr attr,
+			List<TextEdit> edits) {
+
+		String name = attr.getName();
+		String split = formattingOptions.getXsiSchemaLocationSplit();
```
angelozerr, line 167: Get the Split enum like old formatter do that and uses this enum instead of using the string.

Once yiu have the Split enum xheck if it is none

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -155,4 +158,58 @@ public String getCurrentLineIndent(XMLBuilder xml, XMLFormattingOptions formatti
...
+		int start = attr.getNodeAttrValue().getStart();
+		int tabSize = formattingOptions.getTabSize();
+		String attrValue = attr.getOriginalValue();
+
+		if (split.equals("none") || !XSISchemaModel.isXSISchemaLocationAttr(name, attr)) {
+			return true;
+		}
+
+		// Remove extra spaces between start of attribute value quote and actual value
+		formatterDocument.removeLeftSpaces(start + 1, // <foo attr = "|"
```
angelozerr, line 177: Update your samole comment with xsi:schemaLocation instead of attr

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -155,4 +158,58 @@ public String getCurrentLineIndent(XMLBuilder xml, XMLFormattingOptions formatti
...
+		// <foo attr = " |value"
+		start + attr.getFirstContentOffset(attr.getOriginalValue()), edits);
+
+		for (int i = 0; i < attrValue.length(); i++) {
+			if (Character.isWhitespace(attrValue.charAt(i)) && !Character.isWhitespace(attrValue.charAt(i + 1))) {
+				int from = formatterDocument.getLeftWhitespacesOffset(start, start + i + 1);
+				//Insert newline and indent where required based on setting
+				if ((split.equals("onElement") && !StringUtils.isQuote(attrValue.charAt(from - start)))
```
angelozerr, line 185: Do condition if (!StringUtils.isQuote(attrValue.charAt(from - start))) in a top level if

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -155,4 +158,58 @@ public String getCurrentLineIndent(XMLBuilder xml, XMLFormattingOptions formatti
...
+						|| (split.equals("onPair") && !StringUtils.isQuote(attrValue.charAt(from - start))
```
angelozerr, line 186: Use the Split enum

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -155,4 +158,58 @@ public String getCurrentLineIndent(XMLBuilder xml, XMLFormattingOptions formatti
...
+		String attrValue = attr.getOriginalValue();
+
+		if (split.equals("none") || !XSISchemaModel.isXSISchemaLocationAttr(name, attr)) {
+			return true;
+		}
+
+		// Remove extra spaces between start of attribute value quote and actual value
+		formatterDocument.removeLeftSpaces(start + 1, // <foo attr = "|"
+		// <foo attr = " |value"
+		start + attr.getFirstContentOffset(attr.getOriginalValue()), edits);
```
angelozerr, line 179: Same comment than below

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -155,4 +158,58 @@ public String getCurrentLineIndent(XMLBuilder xml, XMLFormattingOptions formatti
...
+
+		for (int i = 0; i < attrValue.length(); i++) {
+			if (Character.isWhitespace(attrValue.charAt(i)) && !Character.isWhitespace(attrValue.charAt(i + 1))) {
+				int from = formatterDocument.getLeftWhitespacesOffset(start, start + i + 1);
+				//Insert newline and indent where required based on setting
+				if ((split.equals("onElement") && !StringUtils.isQuote(attrValue.charAt(from - start)))
+						|| (split.equals("onPair") && !StringUtils.isQuote(attrValue.charAt(from - start))
+								&& isXSD(attrValue, from - start))) {
```
angelozerr, line 187: I dont understand the isXSD condition. More Im not sure notaire is supported corretcly. I suggest you that you do the same rhing than https://github.com/eclipse/lemminx/blob/91da404d922ac0c68b6265de082529c377e2f38b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java by copying passting the code with xml builder and adapt it with text edits

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMDocument.java
JessicaJHee: It is to check for this case: https://github.com/eclipse/lemminx/blob/8ea20b476980f02094ebecf85abd3c050ed7b5b2/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/xsi/XSIFormatterExperimentalTest.java#L117

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/format/IFormatterParticipant.java
```diff
@@ -48,6 +53,22 @@ default boolean formatAttributeValue(String name, String valueWithoutQuote, Char
...
+	 * Format the given attribute value.
+	 * 
+	 * @param formatterDocument the formatter document
+	 * @param indentLevel       the current indent level of the attribute
+	 * @param formattingOptions the formatting options that contain the settings
+	 * @param attr              the attribute node
+	 * @param edits             the text edit list
+	 * @return true if the given attribute can be formatted and false otherwise.
+	 */
+	default boolean formatAttributeValueNew(XMLFormatterDocumentNew formatterDocument,
```
angelozerr, line 66: I would set attr parameter as the first parameter because it is a formatter for attribute

angelozerr (changes requested):

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -155,4 +158,82 @@ public String getCurrentLineIndent(XMLBuilder xml, XMLFormattingOptions formatti
...
+		int tabSize = formattingOptions.getTabSize();
+		String attrValue = attr.getOriginalValue();
+		int indentLevelOffset = getIndentLevelAtAttrName(tabSize, attr.getNodeAttrName().getStart(),
+				attr.getNodeAttrName().getEnd() + 2, formattingOptions.getSplitAttributesIndentSize());
+
+		int lineFeed = split == XSISchemaLocationSplit.onElement ? 1 : 2;
+		int locationNum = 1;
+		for (int i = 0; i < attrValue.length(); i++) {
+			int from = formatterDocument.getLeftWhitespacesOffset(start, start + i + 1);
+			if (Character.isWhitespace(attrValue.charAt(i)) && !Character.isWhitespace(attrValue.charAt(i + 1))
```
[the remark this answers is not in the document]
JessicaJHee: I have added the check `getFirstContentOffset(attr.getOriginalValue()) == -1` at the start of `formatAttributeValueNew()` and a few more tests for this case.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMDocument.java
```diff
@@ -321,7 +321,8 @@ public boolean isBeforeProlog(int offset) {
 
 	private static SchemaLocation createSchemaLocation(DOMNode root, String schemaInstancePrefix) {
 		DOMAttr attr = root.getAttributeNode(getPrefixedName(schemaInstancePrefix, "schemaLocation"));
-		if (attr == null) {
+		//Check that the attribute and the attribute value of xsi:schemaLocation is not null
+		if (attr == null || attr.getNodeAttrValue()==null) {
```
angelozerr, line 325: Format this line code

angelozerr:
The code looks good. @datho7561 could you test the pr please.

datho7561:
Looks like a good start. From my testing, when you use `onPair` or `onElement`, it doesn't line up the second line with the first one. However, every line after that is lined up correctly:
SecondLineNotLinedUp

datho7561:
I think given that the behaviour of the existing formatter is to line up the URLs even when `splitAttributes` is disabled, it would be a good idea to ensure that this PR is able to do that.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -155,4 +158,99 @@ public String getCurrentLineIndent(XMLBuilder xml, XMLFormattingOptions formatti
 		return indent.toString();
 	}
 
+	@Override
+	public boolean formatAttributeValueNew(DOMAttr attr, XMLFormatterDocumentNew formatterDocument,
+			int indentLevel, XMLFormattingOptions formattingOptions, List<TextEdit> edits) {
```
angelozerr, line 163: Remove New at the end of the method.

datho7561 (approved):
Great work, Jessica!

datho7561:
Wait actually, some of the tests fail on my computer. Can you take a look again and see if anything odd is happening for you?

datho7561:
> Wait actually, some of the tests fail on my computer. Can you take a look again and see if anything odd is happening for you?

Well, it works well for me now. No idea what changed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/XMLFormatterDocumentNew.java
```diff
@@ -368,6 +389,20 @@ boolean hasLineBreak(int startAttr, int start) {
 		return false;
 	}
 
+	public int getNormalizedLengthAtRange(int startline, int initialOffset) {
+		String text = textDocument.getText().substring(startline, initialOffset);
+		int contentOffset = 0;
```
angelozerr, line 394: Please dont use substring which can be bad for performance memory.

I think you czn do the same code without substring.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/xsi/XSIFormatterExperimentalTest.java
```diff
@@ -101,14 +116,55 @@ public void xsiSchemaLocationSplitOnPair() throws BadLocationException {
...
+				+ //
+				"        http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util.xsd\"> </beans>";
+		String expected = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\r\n" + //
+				"<beans xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xsi:schemaLocation=\"http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd\r\n"
 				+ //
-				"\r\n" + //
-				"</beans>";
+				"                                                                                  http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util.xsd\"> </beans>";
 		assertFormat(content, expected, settings);
+		assertFormat(expected, expected, settings);
```
angelozerr, line 140: All tests must define the expected textedit by using te.

angelozerr:
> I think given that the behaviour of the existing formatter is to line up the URLs even when splitAttributes is disabled, it would be a good idea to ensure that this PR is able to do that.

It seems that the PR requires that splitAttributes is set to true https://github.com/eclipse/lemminx/pull/1252/files#diff-c0186ecdc200d591a9caa3dadf037073d77443cee30423e92600647efbf3fddbR337

You should remove this seetings and it should work without splitAttributes. With the current PR if you remove the line code  https://github.com/eclipse/lemminx/pull/1252/files#diff-c0186ecdc200d591a9caa3dadf037073d77443cee30423e92600647efbf3fddbR337 tests fails.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/xsi/XSIFormatterExperimentalTest.java
```diff
@@ -71,13 +69,43 @@ public void xsiSchemaLocationSplitOnElement() throws BadLocationException {
 				"    xsi:schemaLocation=\"http://www.springframework.org/schema/beans\r\n" + //
 				"                        http://www.springframework.org/schema/beans/spring-beans.xsd\r\n" + //
 				"                        http://www.springframework.org/schema/util\r\n" + //
-				"                        http://www.springframework.org/schema/util/spring-util.xsd\">\r\n" + //
-				"\r\n" + //
-				"</beans>";
-		assertFormat(content, expected, settings);
+				"                        http://www.springframework.org/schema/util/spring-util.xsd\"> </beans>";
+		assertFormat(content, expected, settings, te(5, 24, 6, 8, ""),
```
angelozerr, line 73: for each te line, add `//` at the end of the line and format the full Java file.

JessicaJHee:
> It seems that the PR requires that splitAttributes is set to true https://github.com/eclipse/lemminx/pull/1252/files#diff-c0186ecdc200d591a9caa3dadf037073d77443cee30423e92600647efbf3fddbR337

I already have these three tests for the case where split attribute is set to false:
```
xsiSchemaLocationSplitOnElementWithoutSplitAttribute()
xsiSchemaLocationSplitOnPairWithoutSplitAttribute()
xsiSchemaLocationSplitOnElementWithTabsWithoutSplitAttribute()
```

Is the idea to force split attribute to true when schema location split is set to either onPair or onElement?

angelozerr:
> I already have these three tests for the case where split attribute is set to false:

I think you should not have specific test with split attributes to true. The xsi settings (which format the attribute value of xsi:schemaLocation) should be totally independant from splitAttribute (which format attribute name location).

https://github.com/eclipse/lemminx/pull/1252/files#diff-c0186ecdc200d591a9caa3dadf037073d77443cee30423e92600647efbf3fddbR337

> Is the idea to force split attribute to true when schema location split is set to either onPair or onElement?

Even if looks like the same behavior, we should distinguish them. SplitAttribute format the attribute name location, although xsi settings is for attribute value of xsi:schemaLocation.

datho7561:
> > I already have these three tests for the case where split attribute is set to false:
> 
> I think you should not have specific test with split attributes to true.

Why not? If I understand correctly, the behaviour should be different, since you need to indent the schemaLocation entries a different amount. eg.

```xml
<aaa xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://google.ca my-file.xsd
																			   https://stackoverflow.yaml my-other-schema.xsd
																			   https://github.com my-github-schema.xsd">
	<bbb> Hello</bbb>
	<ccc />
</aaa>
```

If you were to indent to the same level as you would do with splitAttributes: "true", then the entries don't line up.

angelozerr:
Good catch @datho7561. Lets keep tests  with Split attributes. But the main tests should be done with no Split attributes.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -155,4 +158,104 @@ public String getCurrentLineIndent(XMLBuilder xml, XMLFormattingOptions formatti
...
+	private static int getIndentLevelAtAttrName(int tabSize, int start, int end, int splitAttributesIndentSize,
+			int startline, int normalizedOffset, boolean isSplitAttributes) {
+		if (isSplitAttributes) {
+			if ((end - start) % tabSize == 0) {
+				return (end - start) / tabSize + splitAttributesIndentSize;
+			} else {
+				return (end - start) / tabSize + splitAttributesIndentSize + 1;
+			}
+		} else {
+			if ((normalizedOffset - startline) % tabSize == 0) {
```
datho7561, line 253: I'm not getting additional spaces to line up the entries exactly anymore, instead it's (neededSpaces/tabWidth) rounded up. Is this intentional? I think it's an okay formatting strategy, but it's different than what was done before, and we will need to adapt the documentation to reflect that.

JessicaJHee: Okay I can change it so that it's the exact number of spaces, I originally thought that we wanted to use tabs.

datho7561: I think the idea is that we use tabs if the formatting settings specify tabs, and spaces if the formatting settings specify spaces. However, we always add a few extra spaces to line it up exactly. At least, this is what the old behaviour was.

angelozerr: Indeed, it was the idea. We should have tests with spaces and with tabs settings.

angelozerr:
After playing with the PR, I noticed that it is important to support too https://github.com/redhat-developer/vscode-xml/blob/main/docs/Formatting.md#xmlformatpreserveattributelinebreaks

If `xml.format.splitAttribute is set to false`, and `xml.format.preserveAttributeLineBreaks is set to true`, it doesn't format correctly the XML.

I think we should have a test with `xsiSchemaLocationSplitOnElementWithTabsWithPreserverAttributeLineBreaks` which is a copy/paste of `xsiSchemaLocationSplitOnElementWithTabsWithSplitAttribute` by relacing https://github.com/eclipse/lemminx/pull/1252/files#diff-c0186ecdc200d591a9caa3dadf037073d77443cee30423e92600647efbf3fddbR298 with 

`settings.getFormattingSettings().setPreserveAttributeLineBreaks(true);`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -155,4 +158,78 @@ public String getCurrentLineIndent(XMLBuilder xml, XMLFormattingOptions formatti
...
+
+		int tabSize = formattingOptions.getTabSize();
+		int indentSpaceOffset;
+		int startOfLineOffset = formatterDocument.getLineAtOffset(attr.getOwnerElement().getStart());
+
+		if (formattingOptions.isSplitAttributes()) {
+			indentSpaceOffset = (attrValueStart + 1) - attr.getNodeAttrName().getStart()
+					+ formattingOptions.getSplitAttributesIndentSize() * tabSize;
+		} else if (formattingOptions.isPreserveAttributeLineBreaks()) {
+			indentSpaceOffset = attrValueStart
```
angelozerr, line 187: It seems it works only with tabs? What about spaces?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/XMLFormatterDocumentNew.java
```diff
@@ -368,6 +400,34 @@ boolean hasLineBreak(int startAttr, int start) {
...
+		}
+		return to;
+	}
+
+	public int getLengthWithTabs(int from, int to, int tabSize) {
+		String text = textDocument.getText();
+		for (int i = to; i > from; i--) {
+			if (text.charAt(i)=='\t'){
+				to -= tabSize;
+			}else if (isLineSeparator(text.charAt(i))){
```
angelozerr, line 422: Format your Java code

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/xsi/XSIFormatterExperimentalTest.java
```diff
@@ -71,16 +102,47 @@ public void xsiSchemaLocationSplitOnElement() throws BadLocationException {
...
+	@Override
+	public String toString() {
+		return "XSIFormatterExperimentalTest []";
 	}
 
-	@Disabled
 	@Test
 	public void xsiSchemaLocationSplitOnPair() throws BadLocationException {
 		SharedSettings settings = createSettings();
+		settings.getFormattingSettings().setSplitAttributes(false);
```
angelozerr, line 122: I think you should force setPreserveAttributeLineBreak to false because it will become to true by default. I think you should force too for other test splitAttributes to false to be sure to be in this context.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/xsi/XSIFormatterExperimentalTest.java
```diff
@@ -101,53 +163,140 @@ public void xsiSchemaLocationSplitOnPair() throws BadLocationException {
...
+				te(3, 24, 4, 8, ""), //
+				te(4, 51, 4, 52, "\r\n\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t "), //
+				te(4, 112, 5, 8, "\r\n\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t "), //
+				te(5, 50, 5, 51, "\r\n\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t "), //
+				te(5, 111, 7, 0, " "));
+		assertFormat(expected, expected, settings);
+	}
+
+	@Test
+	public void xsiSchemaLocationSplitOnElementWithTabsWithSplitAttribute() throws BadLocationException {
```
angelozerr, line 294: It should be good to have a test too with setInsertSpaces to false. You could copy paste this test and set setInsertSpacesto false.

datho7561 (approved):
Works really well. Thanks, Jessica!

PR #1253, opened 2022-07-14
Use proxy config as HTTPS proxy as well

datho7561:


rgrunber (approved):
Makes sense to me. We're just re-using the same proxy address for http & https.

rgrunber (approved):
Makes sense to bump to 0.22.0 as we already have some improvements to the experimental formatter for the upcoming release.

PR #1267, opened 2022-07-29
Set xml.format.xsiSchemaLocationSplit to onPair by default

JessicaJHee:
Fixes https://github.com/redhat-developer/vscode-xml/issues/749

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java
@@ -16,6 +16,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import org.eclipse.lemminx.dom.DOMElement;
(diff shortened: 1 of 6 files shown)


datho7561:
Take a look at `XMLFormattingOptions#setDefaultOptions()`. I think you might need to set the setting there as well.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/settings/XMLFormattingOptions.java
```diff
@@ -41,6 +41,8 @@ public class XMLFormattingOptions extends org.eclipse.lemminx.settings.LSPFormat
 
 	public static final int DEFAULT_SPLIT_ATTRIBUTES_INDENT_SIZE = 2;
 
+	public static final String DEFAULT_XSI_SCHEMA_LOCATION_SPLIT = "onPair";
```
datho7561, line 44: Consider using `XSISchemaLocationSplit.onPair.name()`, so that if we ever decide to rename this setting, this will continue to work.

datho7561 (approved):
Thank you, Jessica!

PR #1271, opened 2022-08-10
Improve readability of the first line in error messages.

### LINKED ISSUE #1145: XML validation errors of "Language Servers" are only partially visible
This is from Eclipse 2021-12 (4.22): this xml snippet causes two validation errors:

```
<application xmlns="https://jakarta.ee/xml/ns/jakartaee"
	xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/application_9.xsd">
	<invalid>
	</invalid>
</application>
```

The "Language Servers" section shows only the headers of the error message, in my sample "Attribute:" and "Invalid element name".
Markers View
The other two errors seem to result from the WTP XML validator.

Copying the details shows the full error:
```
Attribute:
 - version
is required in element:
 - application
```
And:
```
Invalid element name:
 - invalid

One of the following is expected:
 - application-name
 - description
 - display-name
 - icon
 - initialize-in-order
 - module
```

So, the "Markers" views should either remove the newlines or display multiline texts.

Hope this is the right place to report this issue.

JessicaJHee:
image
image

Fixes #1145 

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xerces/LSPMessageFormatter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xerces/LSPMessageFormatter.java
@@ -137,9 +137,6 @@ public class LSPMessageFormatter implements MessageFormatter {
 			return cvc_2_4_b_solution(arguments);
 		case cvc_enumeration_valid:
 			return enumeration_valid_solution(arguments);
-		case cvc_complex_type_4:
-			arguments[1] = "- " + arguments[1];
-			arguments[0] = "- " + arguments[0];
 		default:
 			return arguments;
 		}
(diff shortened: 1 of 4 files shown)


datho7561 (changes requested):
There are a few other error messages that I think could be changed that are in the `XMLSchemaMessagesReformatted.properties` file:
- cvc-complex-type.2.4.a
- cvc-complex-type.4

JessicaJHee:
> There are a few other error messages that I think could be changed that are in the `XMLSchemaMessagesReformatted.properties` file:
> 
> * cvc-complex-type.2.4.a
> * cvc-complex-type.4

I believe these are already updated?
https://github.com/eclipse/lemminx/blob/0d2041b4079c9b80c1a0544b01fa91204c1d5882/org.eclipse.lemminx/src/main/resources/XMLSchemaMessagesReformatted.properties#L28

https://github.com/eclipse/lemminx/blob/0d2041b4079c9b80c1a0544b01fa91204c1d5882/org.eclipse.lemminx/src/main/resources/XMLSchemaMessagesReformatted.properties#L32

datho7561:
> I believe these are already updated?

Yes, they were. My bad.

datho7561 (approved):
Looks good. Thanks Jessica!

Issue #1278, opened 2022-08-17
Code action cvc_complex_type_2_4_b fails when child elements are of complex type


JessicaJHee:
Along with this https://github.com/eclipse/lemminx/pull/1257#discussion_r947977144, there are some issues with this code action in the cases where the child element has no type or complex definition, such as element 'a' in the following example. (ex: no ` type="xs:string" ` or `<xs:complexType> ... </xs:complexType> inside the element definition`). 

With the following definition file:
```
<?xml version="1.0" encoding="UTF-8"?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
	<xs:element name="root">
		<xs:complexType>
			<xs:sequence>
				<xs:element name="a" minOccurs="1" maxOccurs="1" />
				<xs:element name="b" minOccurs="0" maxOccurs="1" type="xs:string" />
				<xs:element name="c" minOccurs="0" maxOccurs="1" type="xs:string" />
				<xs:element name="d" minOccurs="1" maxOccurs="1">
					<xs:complexType>
						<xs:sequence>
							<xs:element name="d1" minOccurs="0" maxOccurs="unbounded" type="xs:string" />
						</xs:sequence>
						<xs:attribute name="status" type="xs:string" />
						<xs:attribute name="age" use="required" type="xs:int" />
					</xs:complexType>
				</xs:element>
				<xs:element name="e" minOccurs="1" maxOccurs="1" type="xs:string" />
			</xs:sequence>
		</xs:complexType>
	</xs:element>
</xs:schema>
```
![Peek 2022-08-17 15-33](https://user-images.githubusercontent.com/73968480/185226982-43f07741-995e-45cd-850e-3748de6db2a4.gif)

The elements inserted are incorrect.

PR #1281, opened 2022-08-18
Check for null elementOptionality entry for code action cvc_complex_type_2_4_b

### LINKED ISSUE #1280: Null pointer exception when processing code action cvc_complex_type_2_4_b
Based on this comment: https://github.com/eclipse/lemminx/pull/1257#discussion_r947977144

When the element does not have an optionality entry in the elementOptionality list, a NPE is triggered.

JessicaJHee:
Fixes #1280 

Signed-off-by: Jessica He [jhe@redhat.com](mailto:jhe@redhat.com)

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/contentmodel/CMXSDElementDeclaration.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/contentmodel/CMXSDElementDeclaration.java
@@ -335,7 +335,8 @@ public class CMXSDElementDeclaration implements CMElementDeclaration {
 					}
 			}
 		}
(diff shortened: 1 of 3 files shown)


angelozerr:
@JessicaJHee do you think we could have a simple test for this PR? If it is too complex, please ignore my comment.

JessicaJHee:
> @JessicaJHee do you think we could have a simple test for this PR? If it is too complex, please ignore my comment.

yes we can, but the result from the code action would be incorrect.

angelozerr:
Yes you are righr but this test will be fixed again with 
https://github.com/eclipse/lemminx/issues/1278

I mean that you will have test that you should adjust.

angelozerr:
Thanks @JessicaJHee !

PR #1282, opened 2022-08-18
Improve download error message with detailed cause

JessicaJHee:
Fixes https://github.com/redhat-developer/vscode-xml/issues/763

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/CacheResourceDownloadedException.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/CacheResourceDownloadedException.java
@@ -26,7 +26,7 @@ public class CacheResourceDownloadedException extends CacheResourceException {
 
 	public enum CacheResourceDownloadedError {
 
-		ERROR_WHILE_DOWNLOADING("Error while downloading ''{0}'' to ''{1}''.");
+		ERROR_WHILE_DOWNLOADING("Error while downloading ''{0}'' to ''{1}'' : ''{2}''.");
 
 		private final String rawMessage;
 
@@ -42,13 +42,13 @@ public class CacheResourceDownloadedException extends CacheResourceException {
 
 	private CacheResourceDownloadedError errorCode;
 
-	public CacheResourceDownloadedException(String resourceURI, Path resourceCachePath, Throwable e) {
-		this(resourceURI, resourceCachePath, CacheResourceDownloadedError.ERROR_WHILE_DOWNLOADING, e);
+	public CacheResourceDownloadedException(String resourceURI, Path resourceCachePath,  String cause, Throwable e) {
+		this(resourceURI, resourceCachePath, cause, CacheResourceDownloadedError.ERROR_WHILE_DOWNLOADING, e);
 	}
 
-	public CacheResourceDownloadedException(String resourceURI, Path resourceCachePath,
+	public CacheResourceDownloadedException(String resourceURI, Path resourceCachePath, String cause, 
 			CacheResourceDownloadedError errorCode, Throwable e) {
-		super(resourceURI, errorCode.getMessage(resourceURI, resourceCachePath), e);
+		super(resourceURI, errorCode.getMessage(resourceURI, resourceCachePath, cause), e);
 		this.errorCode = errorCode;
 	}
 
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/CacheResourcesManager.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/CacheResourcesManager.java
@@ -242,8 +242,12 @@ public class CacheResourcesManager {
 				String error = "[" + rootCause.getClass().getTypeName() + "] " + rootCause.getMessage();
 				LOGGER.log(Level.SEVERE,
 						"Error while downloading " + resourceURI + " to " + resourceCachePath + " : " + error);
+				String httpResponseCode = getHttpResponseCode(conn);
+				if (httpResponseCode != null) {
+					error = error + " with code: " + httpResponseCode;
+				}
 				CacheResourceDownloadedException cacheException = new CacheResourceDownloadedException(resourceURI,
-						resourceCachePath, e);
+						resourceCachePath, error, e);
 				unavailableURICache.put(resourceURI, cacheException);
 				throw cacheException;
 			} finally {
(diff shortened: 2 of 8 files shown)


angelozerr:
Could you add a screendhot with error message please.

It seems CI build fails with some tests

rgrunber:
Odd.. it passes locally for me under Java 1.8 & Java 17.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/CacheResourceDownloadedException.java
```diff
@@ -26,7 +26,7 @@ public class CacheResourceDownloadedException extends CacheResourceException {
 
 	public enum CacheResourceDownloadedError {
 
-		ERROR_WHILE_DOWNLOADING("Error while downloading ''{0}'' to ''{1}''.");
+		ERROR_WHILE_DOWNLOADING("Error while downloading ''{0}'' to ''{1}'' : ''{2}''.");
```
datho7561, line 29: If possible, it would be nice to get the HTTP status code put into this message so that the user has a starting point to start figuring out why the download failed (i.e. 404 could mean they put the schema address in wrong, 407 could mean they configured their proxy incorrectly)

JessicaJHee: I'm not sure if I'm able to get the HTTP status code since we are using URLConnection https://github.com/eclipse/lemminx/blob/0535b585f54e5b22c9af532eb802bd27803c841c/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/CacheResourcesManager.java#L192

rgrunber: We may not be able to get it in every case (ie. if a url connection isn't using http), but it should be possible to check whether the object we're treating as a URLConnection happens to be an HttpURLConnection. If so, the cast should succeed and error code should be there.

https://github.com/eclipse/lemminx/blob/0535b585f54e5b22c9af532eb802bd27803c841c/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/CacheResourcesManager.java#L253

angelozerr: > We may not be able to get it in every case

An another usecase is when connection is refused, we cannot get the HTTP status.

angelozerr:
It seems that the problem of CI build comes from error message which i sdifferent from Java version.

To fix this issue, I suggest that your test works with `[java.net.ConnectException] Connection refused` and you should adjust the actual error message to have every time `[java.net.ConnectException] Connection refused`.  To do that, create in XMLAssert this function (I do that quickly, please clean code and add some comments if required):

```java
// Java exception messages
	
	private static final Map<String, String> exceptionMessagesMapping;

	static {
		exceptionMessagesMapping = new HashMap<String, String>();
		exceptionMessagesMapping.put("[java.net.ConnectException] Connection refused: connect", "[java.net.ConnectException] Connection refused"); // Java 8
		exceptionMessagesMapping.put("[java.net.ConnectException] Connection refused (Connection refused)", "[java.net.ConnectException] Connection refused"); // Java ???
	}

	private static final Function<String, String> fixExceptionMessage = (message) -> {
		for (Map.Entry<String, String> replace : exceptionMessagesMapping.entrySet()) {
			if (message.contains(replace.getKey())) {
				message = message.replace(replace.getKey(), replace.getValue());
			}
		}
		return message;
	};
```

And update existing XMLAssert#assertPublishDiagnostics like this:

```
public static void assertPublishDiagnostics(List<PublishDiagnosticsParams> actual,
		PublishDiagnosticsParams... expected) {
	assertEquals(expected.length, actual.size());
	for (int i = 0; i < expected.length; i++) {
		assertEquals(expected[i].getUri(), actual.get(i).getUri());
		actual.get(i).getDiagnostics().forEach(d -> {
			d.setMessage(fixExceptionMessage.apply(d.getMessage()));
		});
		assertDiagnostics(actual.get(i).getDiagnostics(), expected[i].getDiagnostics(), false);
	}
}
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/uriresolver/CacheResourcesManager.java
```diff
@@ -258,6 +262,25 @@ private CompletableFuture<Path> downloadResource(final String resourceURI, Path
...
+	 * be retrived
+	 *
+	 * @param conn the URL connection
+	 * @return the http response code from a url connection, null if code could not
+	 *         be retrived
+	 */
+	private static String getHttpResponseCode(URLConnection conn) {
+		if (conn != null && conn instanceof HttpURLConnection) {
+			try {
+				return ((HttpURLConnection) conn).getResponseMessage();
```
rgrunber, line 276: I think you want to call `getResponseCode()` here, which would return an integer.

See https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/HttpURLConnection.html#getResponseCode() .

JessicaJHee: I could also do a combination of both where the message would be something like '404 Not Found'? I'm also looking for if there is a good way to test this case.

rgrunber: Yeah, I like the combination of both actually. A little easier for users who may not recall.

JessicaJHee:
In the test I added: schemaWithUrlWithCacheHttpCode(), I expected to have DownloadProblem as I do when I test the extension, but I get schema_reference.4 instead. 

image

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/contentmodel/XMLSchemaPublishDiagnosticsTest.java
```diff
@@ -153,18 +155,53 @@ public void schemaWithUrlWithCache() throws Exception {
...
+				new Diagnostic(r(2, 31, 2, 67),
+						"The resource 'https://www.w3.org/2001XMLSchema.xsd' is downloading in the cache path '"
+								+ xsdCachePath + "'.",
+						DiagnosticSeverity.Information, "xml", ExternalResourceErrorCode.DownloadingResource.getCode()),
+				new Diagnostic(r(1, 1, 1, 8), "cvc-elt.1.a: Cannot find the declaration of element 'invoice'.",
+						DiagnosticSeverity.Error, "xml", XMLSchemaErrorCode.cvc_elt_1_a.getCode())));
+
+		TimeUnit.SECONDS.sleep(5); // HACK: to make the timing work on slow machines
+
+		XMLAssert.testDiagnosticsFor(xml,
```
angelozerr, line 196: Replace with:

```java
XMLAssert.testPublishDiagnosticsFor(xml, fileURI, ls,
				pd(fileURI,
						new Diagnostic(r(2, 31, 2, 67),
								"Error while downloading 'https://www.w3.org/2001XMLSchema.xsd' to '" + xsdCachePath
										+ "' : java.io.FileNotFoundException",
								DiagnosticSeverity.Error, "xml", ExternalResourceErrorCode.DownloadProblem.getCode()),
						new Diagnostic(r(1, 1, 1, 8), "cvc-elt.1.a: Cannot find the declaration of element 'invoice'.",
								DiagnosticSeverity.Error, "xml", XMLSchemaErrorCode.cvc_elt_1_a.getCode())));
```

Test should work.

angelozerr:
It works like a charm:

image

Great job @JessicaJHee !

PR #1289, opened 2022-09-02
maven client is called mvn on Unix™ platforms

mcepl:
Just fixing a typo.


fbricon:
Not a typo. mvnw is a mvn wrapper and is locally present in the repo. Will download mvn is it's missing from the user's machine.

mcepl:
I see. Right, I prefer mvn from my system vendor, but I understand.

PR #1290, opened 2022-09-02
Suppress validation kinds based on file pattern

### LINKED ISSUE #1275: Suppress validation kinds based on file pattern
Is there a way to opt of validation for specific know content types? 

The question arises relates to this issue raised for PDE:

https://github.com/eclipse-pde/eclipse.pde/issues/294

PDE has abused the XML Schema namespace for its own purpose leading to all kinds of validation errors that are not really errors for PDE's actual use of *.exsd resources.

Similarly a PDE *.target file doesn't specify a schema, but the editor knows the correct structure so a warning about the schema being missing is just noise to the user. 

The Lemminx functionality is generally very cool and very useful, but it seems the only way to disable this for PDE's resources is to disable it in general, which definitely not desired either...

angelozerr:
Suppress validation kinds based on file pattern

Fixes #1275

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationFilter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationFilter.java
@@ -1,0 +1,52 @@
+/**
+ *  Copyright (c) 2022 Red Hat Inc. and others.
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v2.0
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *  Red Hat Inc. - initial API and implementation
+ */
+package org.eclipse.lemminx.extensions.contentmodel.settings;
+
+import org.eclipse.lemminx.settings.PathPatternMatcher;
+
+/**
+ * XML validation settings used for a given pattern file.
+ * 
+ * <code>
+ * {
+      "pattern": "**{.project,.classpath,plugin.xml,feature.xml,category.xml,.target,.product}",
+      "noGrammar": "ignore"
+   }
+ * </code>
+ * 
+ * @author Angelo ZERR
+ *
+ */
+public class XMLValidationFilter extends XMLValidationSettings {
+
+	private String pattern;
+
+	private transient PathPatternMatcher matcher;
+
+	public boolean matches(String uri) {
+		if (matcher == null) {
+			matcher = new PathPatternMatcher();
+			matcher.setPattern(pattern);
+		}
+		return matcher.matches(uri);
+	}
+
+	public String getPattern() {
+		return pattern;
+	}
+
+	public void setPattern(String pattern) {
+		this.pattern = pattern;
+		this.matcher = null;
+	}
+}
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationRootSettings.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationRootSettings.java
@@ -1,0 +1,133 @@
+/**
+ *  Copyright (c) 2022 Red Hat Inc. and others.
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v2.0
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *  Red Hat Inc. - initial API and implementation
+ */
+package org.eclipse.lemminx.extensions.contentmodel.settings;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * XML validation settings root which defines validation rules for all files.
+ * 
+ * The {@link XMLValidationRootSettings}{@link #filters} can be used to define
+ * custom validation rule for a given file pattern:
+ * 
+ * <code>
+[
+  {
+    "pattern": "**.exsd",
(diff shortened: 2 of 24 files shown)


angelozerr:
It starts working.  If you add in your `settings.json`

```json
"xml.validation.filters": [
      {
         "pattern": "**/*.exsd",
         "enabled": false
      },
      {
         "pattern": "plugin.xml",
         "noGrammar": "ignore"
      },
      {
         "pattern": ".project",
         "noGrammar": "ignore"
      },
      {
         "pattern": ".classpath",
         "noGrammar": "ignore"
      }
   ]
```

It should disable validation for exsd files, and disable the noGrammar hint for plugin.xml, .classpath, .project files.

I have not managed severity for error code because we don't need it for the moment.

The main idea that I implemented is that you can define all validation settings (xml.validation.enabled, xml.validation.noGrammar, etc) that we have for a given pattern.

For instance by default we have:

`xml.validation.noGrammar : "hint"` which display an information on root tag element which says that there are no grammar. It is a global validation settings.

For plugin.xml we don't want this "hint", so you can write a filter like this:

```json
{
    "pattern": "plugin.xml",
    "noGrammar": "ignore"
 },
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationRootSettings.java
```diff
@@ -0,0 +1,100 @@
...
+		if (!super.equals(obj))
+			return false;
+		if (getClass() != obj.getClass())
+			return false;
+		XMLValidationRootSettings other = (XMLValidationRootSettings) obj;
+		return Arrays.equals(filters, other.filters);
+	}
+
+	private static XMLValidationFilter[] createDefaultFilters() {
+		List<XMLValidationFilter> filters = new ArrayList<>();
```
angelozerr, line 85: @merks here the default validation filter that LemMinX will use by default. Do you see another filter which should be done for other files?

merks:
Thanks for asking.  With PDE there are also *.product files (root element <product>) for defining products, *.target files (root element <target>) for defining target platforms, and category.xml (root element <site>) for defining update sites.

angelozerr:
> Thanks for asking. With PDE there are also *.product files (root element ) for defining products, *.target files (root element ) for defining target platforms, and category.xml (root element ) for defining update sites.

Oh yes you are right, thanks for your feedback. I add those 3 file patterns.

angelozerr:
@datho7561 please see too the vscode-xml PR https://github.com/redhat-developer/vscode-xml/pull/774 which provides this new validation filter setting.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationSettings.java
```diff
@@ -16,7 +16,8 @@
 import org.eclipse.lsp4j.PublishDiagnosticsCapabilities;
 
 /**
- * XMLValidationSettings
+ * XML validation settings for a given file.
```
datho7561, line 19: Is this class ever instantiated directly? If not, I think it might be worthwhile to rename it so that we don't confuse it with `XMLValidationRootSettings`. Maybe `XMLValidationFileSettings`?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationFilter.java
```diff
@@ -0,0 +1,51 @@
...
+		}
+		return matcher.matches(uri);
+	}
+
+	public String getPattern() {
+		return pattern;
+	}
+
+	public void setPattern(String pattern) {
+		this.pattern = pattern;
```
datho7561, line 49: I think the `matcher` should be set to `null` in this method, so that the old matcher is not reused if the pattern is replaced with a new pattern.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationRootSettings.java
```diff
@@ -0,0 +1,126 @@
...
+	static {
+		DEFAULT_FILTERS = createDefaultFilters();
+	}
+
+	public XMLValidationRootSettings() {
+		super();
+		setFilters(DEFAULT_FILTERS);
+	}
+
+	public XMLValidationFilter[] getFilters() {
```
datho7561, line 54: I think a javadoc here explaining what the filters are for would be helpful.

angelozerr:
@merks I have tested quickly with WWD and my PR and I can confirm that it works: You will benefit with No grammar constraints for other XML files and not for plugin.xml for instance:

image

merks:
@angelozerr Thanks!  And also thanks from the uncounted users!!

datho7561 (changes requested):
Works well. I also tried out the binary using neovim (so the lemminx settings use the defaults), and it works properly there as well. There we a few things I noticed in the code, and once they are addressed I think it should be good to merge.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationSettings.java
angelozerr: I would like to avoid doing that because it will break some APIs.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationRootSettings.java
angelozerr: Fixed.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/settings/XMLValidationSettings.java
datho7561: Okay. I think we will get confused by the name in the future at some point, but if we really can't change it, then that's okay.

datho7561 (approved):
Thank you, Angelo!

PR #1293, opened 2022-09-14
Do not complete paths in attr unless beginning of value looks like a path

datho7561:
Fixes redhat-developer/vscode-xml#668

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java
@@ -77,20 +77,20 @@ public class FilePathCompletionParticipant extends CompletionParticipantAdapter 
 	public void onAttributeValue(String value, ICompletionRequest request, ICompletionResponse response,
 			CancelChecker cancelChecker) throws Exception {
 		// File path completion on attribute value
-		addCompletionItems(value, request, response);
+		addCompletionItems(value, request, response, false);
 	}
 
 	@Override
 	public void onDTDSystemId(String value, ICompletionRequest request, ICompletionResponse response,
 			CancelChecker cancelChecker) throws Exception {
 		// File path completion on DTD DOCTYPE SYSTEM
-		addCompletionItems(value, request, response);
+		addCompletionItems(value, request, response, true);
 	}
 
-	private static void addCompletionItems(String value, ICompletionRequest request, ICompletionResponse response)
+	private static void addCompletionItems(String value, ICompletionRequest request, ICompletionResponse response, boolean isInDoctype)
 			throws Exception {
 		String fullValue = value;
-		if (isEmpty(fullValue)) {
+		if (isEmpty(fullValue) && !isInDoctype) {
 			return;
 		}
 
@@ -101,7 +101,7 @@ public class FilePathCompletionParticipant extends CompletionParticipantAdapter 
 		// ex value="file:///C:/fold|er"
 		int valuePathStartOffset = xmlDocument.offsetAt(request.getReplaceRange().getStart());
 		int endPathOffset = request.getOffset(); // offset after the typed character
-		int startPathOffset = StringUtils.getOffsetAfterWhitespace(fullValue, endPathOffset - valuePathStartOffset)
+		int startPathOffset = (fullValue.length() == 0 ? 0 : StringUtils.getOffsetAfterWhitespace(fullValue, endPathOffset - valuePathStartOffset))
 				+ valuePathStartOffset; // first character of URI
 		Range filePathRange = XMLPositionUtility.createRange(startPathOffset, endPathOffset, xmlDocument);
 		String originalValuePath = text.substring(startPathOffset, endPathOffset);
@@ -132,6 +132,9 @@ public class FilePathCompletionParticipant extends CompletionParticipantAdapter 
 					return;
 				}
 			}
+		} else if (!hasPathBeginning(valuePath) && !isInDoctype) {
+			// the user probably didn't intend to complete a path
+			return;
 		}
 		// On Linux, Mac OS replace '\\' with '/'
 		if (!isWindows) {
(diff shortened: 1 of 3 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -259,4 +262,13 @@ private static void createFilePathCompletionItem(File f, Range replaceRange, ICo
 		response.addCompletionItem(item);
 	}
 
+	private static boolean hasPathBeginning(String currentText) {
+		return currentText.startsWith("/")
+		|| currentText.matches("^[A-Za-z]:\\\\.*")
```
angelozerr, line 267: Is it possible to do that without RegeExp?

datho7561: At the very least I change this to a `Pattern`. I think that a RegExp is the cleanest solution, but I could also try to do something with looping over the characters.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -132,6 +132,9 @@ private static void addCompletionItems(String value, ICompletionRequest request,
 					return;
 				}
 			}
+		} else if (!hasPathBeginning(valuePath)) {
```
angelozerr, line 135: This logic is very annoying with the case of attribute that we are sure that it is a path like
`<!DOCTYPE foo SYSTEM "|">` must open the file path completion.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java
angelozerr: Indeed the first improvement is to cache the regexp in Pattern. My fear is this code will be called a lot when user will type some content in attribute and we shoudl have good performance and not some memory problem.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java
datho7561: I looked into the code, and this case would already not return completion, since it's an empty string. I've changed that, so that in doctypes, when you're within `""`, you always get filepath completion. My implementation is pretty naive, but luckily, paths are not suggested when you attempt completion in the subset, eg.

```xml
<!DOCTYPE aaa [
    <!ELEMENT aaa (#PCDATA)>
    <!ENTITY entity-name "|">
]>
<aaa Name=""></aaa>
```
There is no file completion returned at `|`.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java
datho7561: I switched to using a function instead of a regex

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -89,17 +89,19 @@ public void onDTDSystemId(String value, ICompletionRequest request, ICompletionR
...
 
 		DOMDocument xmlDocument = request.getXMLDocument();
 		String text = xmlDocument.getText();
-
 		// Get value and range for file path declared inside the attribute value
 		// ex value="file:///C:/fold|er"
 		int valuePathStartOffset = xmlDocument.offsetAt(request.getReplaceRange().getStart());
+
+		String fullValue = value;
+		boolean inDoctype = isInDoctype(valuePathStartOffset, xmlDocument);
```
angelozerr, line 100: You don't need to have  `isInDoctype` method because `addCompletionItems` by `onDTDSystemId`, so you can add a boolean parameter `isInDoctype` in  `addCompletionItems` which is filled correctly.

datho7561: Good catch! I've fixed this

datho7561:
@angelozerr if you have time, could you please take a look at this?

angelozerr:
I promise you to review all your PR as soon as I will finish my contribution with generic support for qute.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java
```diff
@@ -259,4 +262,33 @@ private static void createFilePathCompletionItem(File f, Range replaceRange, ICo
...
+	}
+
+	private static boolean isAbsoluteWindowsPath(String currentText) {
+		if (currentText.length() < 3) {
+			return false;
+		}
+		char driveLetter = currentText.charAt(0);
+		if (driveLetter < 'A' || (driveLetter > 'Z' && driveLetter < 'a') || (driveLetter) > 'z') {
+			return false;
+		}
```
rgrunber, line 283: Seems like `Character.isLetter(..)` should work here. In fact if you're going to try matching a letter, followed by ':' & '\\\\' , might it be easier to just use a regular expression, or just String#matches(..) ?

rgrunber (approved):
Change looks pretty good to me so I think we can probably merge. Just a minor adjustment.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/general/completion/FilePathCompletionParticipant.java
datho7561: Okay. I'm avoiding regexp according to Angelo's feedback. I'll switch to `Character.isLetter()`, that's way more legible that what I'm doing now

rgrunber: Fair enough.

datho7561:
I think it'd be good to check that this works properly on Windows before merging. I haven't been able to test if it works there yet.

angelozerr:
@datho7561 could you move the file path completion participant and the test in `extensions/filepath` instead of `extensions/general`.

angelozerr:
I'm sorry @datho7561 several tests on Windows OS fail.

 * testFilePathCompletionForEmptyDoctype
 * testFilePathCompletionFolderBAbsolutePathBackSlash
 * testFilePathCompletionFolderBAbsolutePath

It found 0 items for those tests.

datho7561:
I can investigate it on my Windows machine. I'll try to get these fixed today

datho7561:
Let me know if this works for you, Angelo. I think something is wrong with CodeQL, and Jenkins didn't trigger for some reason, so I'll check the changes under linux

datho7561:
@angelozerr could you check now please?

datho7561:
@angelozerr is it okay to merge this PR?

angelozerr:
Tests pass now in Windows OS. Thanks @datho7561 !

PR #1299, opened 2022-09-22
Fix formatting with multiple line content

JessicaJHee:
Fixes https://github.com/redhat-developer/vscode-xml/issues/600


datho7561:
It looks like the test fails. I think it would be a good idea to have versions of this test with `joinContentLines` enabled and disabled now that it works.

datho7561 (approved):
Looks good to me. Thanks, Jessica!

PR #1305, opened 2022-09-28
Ensure xml.format.emptyElements respects grammar constraints

### LINKED ISSUE #492: Auto Tag Close formatting
XML formatting should give a preferences like `xml.format.autoTagClose` to auto close the element which is formatted.

In other words, formatting :

```xml
<element name="myName"></element>
```

should format like this:

```xml
<element name="myName" />
```

JessicaJHee:
Fixes #492 

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
@@ -34,7 +34,7 @@ import org.eclipse.lemminx.settings.SharedSettings;
  * </p>
  * 
  * @author Angelo ZERR
- *
+ * 
  */
 public class ContentModelFormatterParticipant implements IFormatterParticipant {
 
@@ -67,4 +67,29 @@ public class ContentModelFormatterParticipant implements IFormatterParticipant {
 		return null;
 	}
 
+	@Override
+	public boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings sharedSettings) {
+		boolean enabled = sharedSettings.getFormattingSettings().isGrammarAwareFormatting();
+		if (!enabled) {
+			return true;
+		}
+		if (!("true".equals(element.getAttribute("xsi:nil")))) {
+			// Only check the schema for value of nillable when xsi:nil="true" in element
+			return true;
+		}
+		Collection<CMDocument> cmDocuments = contentModelManager.findCMDocument(element);
+		if (cmDocuments.isEmpty()) {
+			// The DOM document is not linked to a grammar, the collapse can be done.
+			return true;
+		}
+		for (CMDocument cmDocument : cmDocuments) {
+			CMElementDeclaration cmElement = cmDocument.findCMElement(element);
+			if (cmElement != null && !cmElement.isNillable()) {
+				// Collapse is not allowed in the case where nillable="false" in xsd and
+				// xsi:nil="true" for the element in xml
+				return false;
+			}
+		}
+		return true;
+	}
 }
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/format/IFormatterParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/format/IFormatterParticipant.java
@@ -83,4 +83,18 @@ public interface IFormatterParticipant {
 			XMLFormattingConstraints parentConstraints, SharedSettings sharedSettings) {
 		return null;
 	}
+
+	/**
+	 * Returns true if the given element can be collapsed according to grammar
+	 * constraints.
+	 * 
+	 * @param element        the DOM element.
+	 * @param sharedSettings the shared settings.
+	 * 
+	 * @return true if the given element can be collapsed according to grammar
+	 *         constraints.
+	 */
+	default boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings sharedSettings) {
+		return true;
+	}
 }
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMElementFormatter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMElementFormatter.java
@@ -16,6 +16,7 @@ import java.util.List;
 import org.eclipse.lemminx.commons.BadLocationException;
 import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMElement;
+import org.eclipse.lemminx.settings.SharedSettings;
 import org.eclipse.lemminx.settings.XMLFormattingOptions.EmptyElements;
 import org.eclipse.lemminx.utils.StringUtils;
 import org.eclipse.lsp4j.TextEdit;
@@ -85,7 +86,7 @@ public class DOMElementFormatter {
 		case MixedContent:
 			// Remove spaces and indent if the content between start tag and parent start
 			// tag is some white spaces
-			// before formatting: <a> <b> </b> example text </a>
+			// before formatting: <a> [space][space] <b> </b> example text </a>
 			// after formatting: <a>\n <b> </b> example text </a>
 			int parentStartCloseOffset = element.getParentElement().getStartTagCloseOffset() + 1;
 			if (parentStartCloseOffset != startTagOffset
@@ -153,8 +154,10 @@ public class DOMElementFormatter {
 			}
 			case collapse: {
 				// collapse empty element: <example></example> -> <example />
-				if (!element.isSelfClosed() && (end == -1 || element.getEndTagOpenOffset() + 1 < end)) {
-					// Do not collapse if range is does not cover the element
+				if (!element.isSelfClosed() && (end == -1 || element.getEndTagOpenOffset() + 1 < end)
+						&& (shouldCollapseEmptyElement(element, formatterDocument.getSharedSettings()))) {
+					// Do not collapse if range is does not cover the element or is prohibited by
+					// grammar constraint
 					StringBuilder tag = new StringBuilder();
 					if (isSpaceBeforeEmptyCloseTag()) {
 						tag.append(" ");
@@ -217,11 +220,11 @@ public class DOMElementFormatter {
 	/**
 	 * Formats the start tag's closing bracket (>) according to
 	 * {@code XMLFormattingOptions#isPreserveAttrLineBreaks()}
-	 *
+	 * 
 	 * {@code XMLFormattingOptions#isPreserveAttrLineBreaks()}: If true, must add a
 	 * newline + indent before the closing bracket if the last attribute of the
 	 * element and the closing bracket are in different lines.
-	 *
+	 * 
 	 * @param element
 	 * @throws BadLocationException
 	 */
@@ -285,9 +288,8 @@ public class DOMElementFormatter {
 			// Preserve existing spaces
 			break;
 		case MixedContent:
-			// Remove spaces and indent if the last child is an element or comment block,
-			// not text
-			// before formatting: <a> example text <b> </b> [space][space] </a>
+			// Remove spaces and indent if the last child is an element, not text
+			// before formatting: <a> example text <b> </b> [space][space]</a>
 			// after formatting: <a> example text <b> </b>\n</a>
 			if ((element.getLastChild().isElement() || element.getLastChild().isComment())
 					&& Character.isWhitespace(formatterDocument.getText().charAt(endTagOffset - 1))
@@ -326,7 +328,7 @@ public class DOMElementFormatter {
 
 	/**
 	 * Return the option to use to generate empty elements.
-	 *
+	 * 
 	 * @param element the DOM element
 	 * @return the option to use to generate empty elements.
 	 */
@@ -361,7 +363,7 @@ public class DOMElementFormatter {
 	 * 
 	 * 1. splitAttribute must be set to true 2. there must be at least 2 attributes
 	 * in the element
-	 *
+	 * 
 	 * @param element the DOM element
 	 * @return true if should format according to closingBracketNewLine setting.
 	 */
@@ -460,6 +462,10 @@ public class DOMElementFormatter {
 	private FormatElementCategory getFormatElementCategory(DOMElement element,
 			XMLFormattingConstraints parentConstraints) {
 		return formatterDocument.getFormatElementCategory(element, parentConstraints);
+	}
+
+	private boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings settings) {
+		return formatterDocument.shouldCollapseEmptyElement(element, settings);
 	}
 
 	private int getMaxLineWidth() {
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/XMLFormatterDocumentNew.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/XMLFormatterDocumentNew.java
@@ -44,7 +44,7 @@ import org.w3c.dom.Text;
  * update spaces / indent.
  * 
  * @author Angelo ZERR
- *
+ * 
  */
 public class XMLFormatterDocumentNew {
 
@@ -511,10 +511,10 @@ public class XMLFormatterDocumentNew {
 
 	/**
 	 * Returns the format element category of the given DOM element.
-	 * 
+	 *
 	 * @param element           the DOM element.
 	 * @param parentConstraints the parent constraints.
-	 * 
+	 *
 	 * @return the format element category of the given DOM element.
 	 */
 	public FormatElementCategory getFormatElementCategory(DOMElement element,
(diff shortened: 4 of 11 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/XMLFormatterDocumentNew.java
```diff
@@ -592,6 +592,13 @@ void createTextEditIfNeeded(int from, int to, String expectedContent, List<TextE
 		}
 	}
 
+	public boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings sharedSettings) {
+		for (IFormatterParticipant participant : formatterParticipants) {
+				return participant.shouldCollapseEmptyElement(element, sharedSettings);
```
angelozerr, line 597: fix indentation format of code.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/format/experimental/XMLFormatterEmptyElementsTest.java
```diff
@@ -221,9 +224,60 @@ public void collapseEmptyElementsInRange() throws BadLocationException {
...
+				"  xsi:noNamespaceSchemaLocation=\"src/test/resources/xsd/isEmpty.xsd\">\r\n" + //
+				"  <empty></empty>\r\n" + //
+				"</root>";
+		String expected = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\r\n" + //
+				"<root xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"\r\n" + //
+				"  xsi:noNamespaceSchemaLocation=\"src/test/resources/xsd/isEmpty.xsd\">\r\n" + //
+				"  <empty />\r\n" + //
+				"</root>";
+		assertFormat(content, expected, settings, //
+		te(3, 8, 3, 17, " />"));
```
angelozerr, line 243: format code

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
```diff
@@ -67,4 +67,23 @@ public FormatElementCategory getFormatElementCategory(DOMElement element,
...
+		boolean enabled = sharedSettings.getFormattingSettings().isGrammarAwareFormatting();
+		if (!enabled) {
+			return false;
+		}
+
+		Collection<CMDocument> cmDocuments = contentModelManager.findCMDocument(element);
+		for (CMDocument cmDocument : cmDocuments) {
+			CMElementDeclaration cmElement = cmDocument.findCMElement(element);
+			if (cmElement != null && (cmElement.isEmpty()
+					|| (cmElement.isNillable() && "true".equals(element.getAttribute("xsi:nil"))))) {
```
datho7561, line 81: I think this or here should be an and, right? We only want to collapse when the element doesn't have any content and the schema says that it's okay for it to be self-closing.

JessicaJHee: I used or here because isEmpty() here checks for a specific case: https://github.com/eclipse/lemminx/blob/068e862dde1bd55539ecb8be29e1af9f854e27f5/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsd/contentmodel/CMXSDElementDeclaration.java#L461

Perhaps it's a good idea to add the nillable conditions to this method?

datho7561: Okay. The `||` here is correct. I was confused about how `nillable` worked. I think what I ran into while testing is that I expected an element with a simple type that could be empty, such as `xs:string`, to be collapsed when formatting if it was empty.

eg. I expected `childA` to be collapsed in the following case:

```xml
<root>
  <childA></childA>
</root>
```

given the associated schema:

```xml
  <xs:element name="root">
    <xs:complexType>
      <xs:sequence>
        <xs:element ref="childA" minOccurs="0" maxOccurs="1" />
      </xs:sequence>
    </xs:complexType>
  </xs:element>
  <xs:element name="childA">
    <xs:simpleType>
      <xs:restriction base="xs:string" />
    </xs:simpleType>
  </xs:element>
</xs:schema>
```

The issue is that `isEmpty` checks if the element MUST be empty, not if it's allowed to be empty. Unfortunately, it looks like a lot of work would be required to figure out if an elements content is allowed to be empty. So I think the implementation you have currently is good.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
```diff
@@ -67,4 +67,23 @@ public FormatElementCategory getFormatElementCategory(DOMElement element,
 		return null;
 	}
 
+	@Override
+	public boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings sharedSettings) {
+		boolean enabled = sharedSettings.getFormattingSettings().isGrammarAwareFormatting();
+		if (!enabled) {
+			return false;
```
datho7561, line 74: I think it would be a good idea to collapse empty elements if the grammar aware formatting is off and emptyElements=collapse. My reasoning is that by default we have emptyElements=ignore and grammar aware formatting on, so if a user sets emptyElements=collapse and disables the grammar aware formatting, it's because they want to collapse empty elements even if it doesn't make sense with the grammar.

datho7561 (changes requested):

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
datho7561: What do you think about this Jessica, @angelozerr ?

JessicaJHee: Right I agree, that makes sense

angelozerr:
@JessicaJHee please fix conflicts.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
angelozerr: I agree with you @datho7561.

angelozerr, line 74: `return true;`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
```diff
@@ -67,4 +67,23 @@ public FormatElementCategory getFormatElementCategory(DOMElement element,
...
 
+	@Override
+	public boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings sharedSettings) {
+		boolean enabled = sharedSettings.getFormattingSettings().isGrammarAwareFormatting();
+		if (!enabled) {
+			return false;
+		}
+
+		Collection<CMDocument> cmDocuments = contentModelManager.findCMDocument(element);
+		for (CMDocument cmDocument : cmDocuments) {
```
angelozerr, line 78:
```java
if (cmDocuments.isEmpty()) {
  // The DOM document is not linked to a grammar, the collapse can be done.
  return true;
}
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/format/IFormatterParticipant.java
```diff
@@ -83,4 +83,18 @@ default FormatElementCategory getFormatElementCategory(DOMElement element,
...
+	 * constraints.
+	 * 
+	 * @param element        the DOM element.
+	 * @param sharedSettings the shared settings.
+	 * 
+	 * @return true if the given element can be collapsed according to grammar
+	 *         constraints.
+	 */
+	default boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings sharedSettings) {
+		return false;
```
angelozerr, line 98: I will return true by default.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMElementFormatter.java
```diff
@@ -142,8 +143,11 @@ private int formatStartTagElement(DOMElement element, XMLFormattingConstraints p
 			}
 			case collapse: {
 				// collapse empty element: <example></example> -> <example />
-				if (!element.isSelfClosed() && (end == -1 || element.getEndTagOpenOffset() + 1 < end)) {
-					// Do not collapse if range is does not cover the element
+				if (!element.isSelfClosed() && (end == -1 || element.getEndTagOpenOffset() + 1 < end)
+						&& (!isGrammarAwareFormatting()
```
angelozerr, line 147: Here the element formatter should not use isGrammarAwareFormatting() (it is the participant), I will write just:

`shouldCollapseEmptyElement(element, formatterDocument.getSharedSettings())`

instead of 

`(!isGrammarAwareFormatting() || shouldCollapseEmptyElement(element, formatterDocument.getSharedSettings()))`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMElementFormatter.java
```diff
@@ -479,6 +483,10 @@ private boolean isPreserveEmptyContent() {
 		return formatterDocument.getSharedSettings().getFormattingSettings().isPreserveEmptyContent();
 	}
 
+	private boolean isGrammarAwareFormatting() {
```
angelozerr, line 486: Remove this method.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/XMLFormatterDocumentNew.java
```diff
@@ -580,7 +580,14 @@ void createTextEditIfNeeded(int from, int to, String expectedContent, List<TextE
 		}
 	}
 
-	public String getIndentSpaces(int level, boolean addLineSeparator) {
+	public boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings sharedSettings) {
+		for (IFormatterParticipant participant : formatterParticipants) {
+			return participant.shouldCollapseEmptyElement(element, sharedSettings);
+		}
+		return false;
```
angelozerr, line 587: This method loop for the first participant. I think the method should look like this:

```java
public boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings sharedSettings) {
  for (IFormatterParticipant participant : formatterParticipants) {
    if (!participant.shouldCollapseEmptyElement(element, sharedSettings)) {
      return false;
    }
  }
  return true;
}
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
angelozerr: I think the collapse should be enabled for the most usecases. I have the impression is that the only usecase is when XML defines xsi:nil="true" and the XSD defines nillable="false", collapse should be disabled. I see just this usecase.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
```diff
@@ -67,4 +67,30 @@ public FormatElementCategory getFormatElementCategory(DOMElement element,
 		return null;
 	}
 
+	@Override
+	public boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings sharedSettings) {
+		boolean enabled = sharedSettings.getFormattingSettings().isGrammarAwareFormatting();
+		if (!enabled) {
+			return true;
+		}
+		if (element.getAttribute("xsi:nil") != null) {
```
angelozerr, line 76: I think the only usecase where schema must be search is when you have `xsi:nil=true"`

In other words, I will write that:

```java
if (!("true".equals(element.getAttribute("xsi:nil"))) {
    return true;
}
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
```diff
@@ -67,4 +67,30 @@ public FormatElementCategory getFormatElementCategory(DOMElement element,
...
+		}
+		if (element.getAttribute("xsi:nil") != null) {
+			Collection<CMDocument> cmDocuments = contentModelManager.findCMDocument(element);
+			if (cmDocuments.isEmpty()) {
+				// The DOM document is not linked to a grammar, the collapse can be done.
+				return true;
+			}
+			for (CMDocument cmDocument : cmDocuments) {
+				CMElementDeclaration cmElement = cmDocument.findCMElement(element);
+				if (cmElement != null && cmElement.isEmpty()) {
```
angelozerr, line 84: Why are you doing this test with isEmpty? Have you a test for that? Please adda comment which explains this condition.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
```diff
@@ -67,4 +67,30 @@ public FormatElementCategory getFormatElementCategory(DOMElement element,
...
+					return true;
+				} else if (cmElement != null
+						&& (!cmElement.isNillable() && "true".equals(element.getAttribute("xsi:nil")))) {
```
angelozerr, line 87: As condition "true".equals(element.getAttribute("xsi:nil"))) is done below, you don't need to test it. 

cmElement.isNillable() returns false if XSD doesn't defines nillable attribute, is it correct?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
JessicaJHee: It was suggested in the comments here (https://github.com/eclipse/lemminx/issues/492#issuecomment-1253308411), I will add a test case for this

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
angelozerr, line 76: I will write:

```java
if (!("true".equals(element.getAttribute("xsi:nil")) {
  return true;
}
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
JessicaJHee: This test case I had for this case still passes: `collapseEmptyElementsWithGrammarCanBeEmpty()`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
JessicaJHee: Yes it will return false because nillable is set to false by default.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
angelozerr: Ok I will write:

```java
if (cmElement != null) {
  if (cmElement.isEmpty()) {
    // TODO : add a sample
     return true;
  }
  if (!cmElement.isNillable()) {
    // TODO : add a sample
    return false;
  }
}
```

JessicaJHee: I have removed this isEmpty() since it works well without it with the current logic

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
```diff
@@ -67,4 +67,29 @@ public FormatElementCategory getFormatElementCategory(DOMElement element,
...
+			return true;
+		}
+		Collection<CMDocument> cmDocuments = contentModelManager.findCMDocument(element);
+		if (cmDocuments.isEmpty()) {
+			// The DOM document is not linked to a grammar, the collapse can be done.
+			return true;
+		}
+		for (CMDocument cmDocument : cmDocuments) {
+			CMElementDeclaration cmElement = cmDocument.findCMElement(element);
+			if (cmElement != null && (!cmElement.isNillable())) {
```
angelozerr, line 87: `if (cmElement != null && !cmElement.isNillable()) {`

angelozerr:
It works perfectly now, great hob @JessicaJHee !

PR #1308, opened 2022-10-06
Report only XML syntax error for *.exsd files

angelozerr:
Report only XML syntax error for *.exsd files

According this comment https://github.com/eclipse-pde/eclipse.pde/issues/175#issuecomment-1268147558, *.exd files must report XML syntaxs error without XSD errors.


datho7561:
If I add a `<?xml-model...?>` processing instruction, schema-related errors are still reported in `.exsd` files

angelozerr:
> If I add a <?xml-model...?> processing instruction, schema-related errors are still reported in .exsd files

Could we see that in an another PR please. The main goal today is to have a support of *.exsd files in Eclipse IDE, and xml-model is never used in *.exsd files.

datho7561 (approved):
Works well for me.

Issue #1291, opened 2022-09-07
Replace remark-java


datho7561:
The fork of remark-java that we use has been archived by the maintainer. Look into alternative libraries to accomplish the same thing.

See:
* https://github.com/eclipse/lemminx/pull/1288 (we need to update a library to address a security vulnerability, but updating the library breaks remark-java)
* https://github.com/kotcrab/remark-java
* https://github.com/eclipse/eclipse.jdt.ls/issues/2214

rgrunber:
As you mentioned, we can't use flexmark-java because it requires a Java 11 runtime. Is there any chance we can upstream a fix to revert it to Java 8 ? I got the impression they only started using Java 11 everywhere for fixing some test issues : https://github.com/vsch/flexmark-java/issues/494#issue-1127640636

datho7561:
It seems there isn't much activity on vsch/flexmark-java these days, and Fred ran into a blocker issue on it that hasn't been addressed. So, maybe it's not an option.

PR #1309, opened 2022-10-07
Enable dependabot

akurtakov:


rgrunber:
Looks the same as https://github.com/eclipse/eclipse.jdt.ls/blob/master/.github/dependabot.yml . @datho7561 , do you see any problems with this ?

datho7561:
I know that at one point a version of dependabot was running on this repository. You can see PRs that it has opened: #1250, #1288. We might need to reach out to Fred to disable the other version of dependabot once this PR has been merged. Did you deal with something similar on eclipse.jdt.ls, @rgrunber ?

Other than that, this PR looks good to me.

datho7561:
Oh okay, so this will make PRs for version updates, even if there aren't security updates?

datho7561 (approved):
This seems very helpful. Thanks @akurtakov !

akurtakov:
You're welcome. Btw, is there a reason why JSoup and Jetty PRs dependabot opened are pending and not merged?

datho7561:
The JSoup update has breaking API changes, and a dependency we are using (which is deprecated) depends on the old version. I'll take a look into Jetty.

rgrunber:
> You're welcome. Btw, is there a reason why JSoup and Jetty PRs dependabot opened are pending and not merged?

I think updating jsoup would break remark-java (https://github.com/eclipse/lemminx/issues/1291). Not sure about jetty.

akurtakov:
> It looks like the jetty version which fixes the vulnerability requires Java 11

Is requiring Java 11 problem for Lemminx?

rgrunber:
I don't think there's any technical issues. It's mainly that clients would need to support users with only Java 1.8 on their system. Ther's a good amount of people probably still on Java 1.8. @datho7561 , If we had to switch, I think the client-side could detect when a JVM doesn't meet the minimum required version (eg. 11) and activate the binary version of LemMinX in those cases. Right ? As an example for vscode-xml, it would somewhere at https://github.com/redhat-developer/vscode-xml/blob/fcab66d6c35e497fbfb534c4f0ea6f4fa7d4f822/src/server/serverStarter.ts#L28 .

Moving to 11 would allow use to use flexmark-java though.

datho7561:
> If we had to switch, I think the client-side could detect when a JVM doesn't meet the minimum required version (eg. 11) and activate the binary version of LemMinX in those cases. Right?

Yep, looking through the code that seems to be what vscode-xml will do.

PR #1336, opened 2022-10-17
Improve text formatting for experimental formatter

### LINKED ISSUE #1331: Have xml.format.joinContentLines to true as default behavior in experimental formatter
Currently the experimental formatter will normalize the space before and after the entire text content by default. Here is the current behavior: 
Before formatting: 
```
<a>
  Content
  <b>
    Content2
    Content3
  </b>
</a>
```
After formatting: 
```
<a> Content <b> Content2
    Content3
  </b>
</a>
```

But I believe it would make more sense to have the behavior when `xml.format.joinContentLines` (https://github.com/redhat-developer/vscode-xml/blob/main/docs/Formatting.md#xmlformatjoincontentlines) is set to true by default as it is the case in oxygen.
After formatting: 
```
<a> Content <b> Content2 Content3 </b>
</a>
```
If this behavior makes sense to have as the default, we can then remove the joinContentLines setting and just have the `xml.format.preserveEmptyContent` setting available if an user wishes to keep the text unformatted/preserve any whitespaces. (https://github.com/redhat-developer/vscode-xml/blob/main/docs/Formatting.md#xmlformatpreserveemptycontent)

JessicaJHee:
Fixes #1331 

### DIFF
--- a/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/format/experimental/XMLFormatterExperimentalTest.java
+++ b/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/format/experimental/XMLFormatterExperimentalTest.java
@@ -89,6 +89,7 @@ public class XMLFormatterExperimentalTest extends AbstractCacheBasedTest {
 
 	@Test
 	public void invalidEndTag() throws BadLocationException {
+
 		String content = "</";
 		String expected = content;
 		assertFormat(content, expected);
@@ -700,10 +701,28 @@ public class XMLFormatterExperimentalTest extends AbstractCacheBasedTest {
 				"  Content <!-- comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment -->"
 				+ //
 				"</a>";
+		String expected = "<a>\n" + //
+				"  Content <!-- comment comment comment comment comment comment comment comment\n" + //
+				"  comment comment comment comment comment comment comment comment comment\n" + //
+				"  comment comment comment comment comment comment comment comment --></a>";
+		SharedSettings settings = new SharedSettings();
+		assertFormat(content, expected, settings, //
+				te(1, 78, 1, 79, "\n  "), //
+				te(1, 150, 1, 151, "\n  "));
+		assertFormat(expected, expected, settings);
+	}
+
+	@Test
+	public void testCommentLongWrapJoinContentLines() throws BadLocationException {
+		String content = "<a>\n" + //
+				"  Content <!-- comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment -->"
+				+ //
+				"</a>";
 		String expected = "<a> Content <!-- comment comment comment comment comment comment comment comment\n" + //
 				"  comment comment comment comment comment comment comment comment comment\n" + //
 				"  comment comment comment comment comment comment comment comment --></a>";
 		SharedSettings settings = new SharedSettings();
+		settings.getFormattingSettings().setJoinContentLines(true);
 		assertFormat(content, expected, settings, //
 				te(0, 3, 1, 2, " "), //
 				te(1, 78, 1, 79, "\n  "), //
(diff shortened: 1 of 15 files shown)


org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/format/experimental/XMLFormatterExperimentalTest.java
```diff
@@ -678,37 +679,55 @@ public void testCommentMultiLineContent() throws BadLocationException {
 
 	@Test
 	public void testCommentLongWrap() throws BadLocationException {
-		String content = "<a>\n" + //
-				"  Content <!-- comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment -->"
+		String content = "<a> Content <!-- comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment comment -->"
```
datho7561, line 682: Instead of adjusting `content`, consider maybe set joinContent to true? (or making a duplicate of this test with joinContent=true)

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/format/experimental/XMLFormatterJoinCommentLinesTest.java
```diff
@@ -90,20 +90,19 @@ public void testJoinCommentLinesNested() throws BadLocationException {
 
 	@Test
 	public void testCommentFormatSameLine() throws BadLocationException {
-		String content = "<a>" + lineSeparator() + //
-				" Content" + lineSeparator() + //
+		String content = "<a>\n" + //
```
datho7561, line 93: Make a duplicate of this test with joinContentLines = true (if there isn't one already)

datho7561 (changes requested):

datho7561 (changes requested):

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/format/experimental/XMLFormatterExperimentalTest.java
```diff
@@ -89,16 +89,17 @@ public void singleEndTag() throws BadLocationException {
...
-		String expected = content;
-		assertFormat(content, expected);
-
-		content = "</a";
-		expected = content;
-		assertFormat(content, expected);
-
-		content = "<a></";
-		expected = "<a>" + //
+		/*
```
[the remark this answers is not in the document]
JessicaJHee: Yesn thank you I missed this, it does pass, just forgot to uncomment it

datho7561 (approved):
I am good with this change. Thanks Jessica! I think we discussed waiting until after the release to make this change though.

datho7561:
@angelozerr what do you think about this change? Jessica wanted to wait until after the release before merging it, but I think it should be good to go now.

angelozerr:
I wonder if this PR should cover the usecase:

```xml
<foo>
	<bar>
					</bar>
</foo>
```

to indent correctly bar

```xml
<foo>
	<bar>
	</bar>
</foo>
```

but perhaps teh Oxyfen behavior doesn't allow that?

JessicaJHee:
> When I see tests, it seems we are different from Oxygen formatter, right? It doesn't follow the behavior described in https://github.com/redhat-developer/vscode-xml/blob/main/docs/Formatting.md#xmlformatexperimental ?

Yes it is a bit different when it comes to text formatting since oxygen will always join text content/ normalize the space between text content by default. It is more similar to IntelliJ. I waned to see how you feel about the changes since it is quite different from before. There are a lot of cases to consider but I think with this PR it would be easier to discuss how we want the behavior for text formatting to be like!

> I wonder if this PR should cover the usecase:
> 
> ```
> <foo>
> 	<bar>
> 					</bar>
> </foo>
> ```
> 
> to indent correctly bar
> 
> ```
> <foo>
> 	<bar>
> 	</bar>
> </foo>
> ```
> 
> but perhaps teh Oxygen behavior doesn't allow that?

I wasn't sure because in this case we normalize the space if the content is empty (contains only white space and new lines). Unfortunately I don't have access to try oxygen anymore.
https://github.com/eclipse/lemminx/blob/037589a61986a348f3acd5849f900406e28211cb/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/format/experimental/XMLFormatterPreserveEmptyContentTest.java#L45

angelozerr:
> Yes it is a bit different when it comes to text formatting since oxygen will always join text content/ normalize the space between text content by default. It is more similar to IntelliJ. I waned to see how you feel about the changes since it is quite different from before. There are a lot of cases to consider but I think with this PR it would be easier to discuss how we want the behavior for text formatting to be like!

Indeed we need to specify more teh behavior of the formatting. My wish is to have the IJ and Oxygen behavior both (with a new settings).

I think it is important to preserve Oxygen behavior, because it fixed https://github.com/eclipse/lemminx/issues/594 (I suggest that you read all comments). If we could provide a new settings which format ala IJ, it should be very nice. 

> I wasn't sure because in this case we normalize the space if the content is empty (contains only white space and new lines).

IMHO I think we should take care of this usecase (setting a la IJ).

datho7561:
> `joinContentLines = true` by default so we have the same behavior as oxygen

I think that this is a good idea. The Oxygen formatter handles mixed content very well, so using this behaviour by default is a good idea.

> IMHO I think we should take care of this usecase (setting a la IJ). (angelo)
> For the case empty tags, perhaps we could add a setting (something like 'normalizeSpaceInEmptyElement') if necessary? (jessica)

What should the default behaviour be for empty tags? I think it would be nice to completely remove the content (eg. `<aaa></aaa>`) unless `preserveEmptyContent` is enabled. I don't know what oxygen's default behaviour is, but this is what Intellij does.

angelozerr:
Impressive @JessicaJHee ! Thanks so much. Now we should update vscode-xml doc.

PR #1343, opened 2022-10-24
Verify pull requests on windows with github actions

### LINKED ISSUE #1311: Consider adding a Windows PR verification job
Currently, https://github.com/eclipse/lemminx/blob/main/Jenkinsfile is used for PR verification, and snapshot builds at https://download.eclipse.org/lemminx/snapshots/ which runs on Linux agents only. It would be really nice to also run the PR verification on Windows. This isn't easy to do on Eclipse infrastructure, so we may have to consider doing this verification on GH actions.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

So the options here are :

1.Keep https://github.com/eclipse/lemminx/blob/main/Jenkinsfile as-is and just add a PR verification job on GH actions for Windows
2. Disable only the PR-verification aspect of https://github.com/eclipse/lemminx/blob/main/Jenkinsfile (https://ci.eclipse.org/lemminx/job/lemminx/) and do it all on GH actions (including Windows)
3. Migrate all of https://github.com/eclipse/lemminx/blob/main/Jenkinsfile to GH actions, which would mean deciding what to do with https://download.eclipse.org/lemminx/snapshots/

JessicaJHee:
Fixes #1311 

### DIFF
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -1,0 +1,28 @@
+# This workflow will build a Java project with Maven, and cache/restore any dependencies to improve the workflow execution time
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: pr-verify
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    steps:
+    - name: Check out code 
+      uses: actions/checkout@v3
+    - name: Set up JDK 8
+      uses: actions/setup-java@v3
+      with:
+        java-version: '8'
+        distribution: 'adopt'
+        cache: 'maven'
+    - name: Run Tests on Windows
+      if: runner.os == 'Windows'
+      run: ./mvnw.cmd -B '-Dlemminx.cacheInRepoDir' clean verify
+    - name: Run Tests on Linux
+      if: runner.os == 'Linux'
+      run: ./mvnw -B clean verify
--- a/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/AbstractCacheBasedTest.java
+++ b/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/AbstractCacheBasedTest.java
@@ -14,9 +14,11 @@ package org.eclipse.lemminx;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.UUID;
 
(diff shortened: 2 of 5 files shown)


org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/AbstractCacheBasedTest.java
```diff
@@ -37,7 +39,13 @@ public abstract class AbstractCacheBasedTest {
 
 	static {
 		try {
-			parentDir = Files.createTempDirectory("lemminx");
+			String cacheInRepoDir = System.getProperty("cacheInRepoDir");
```
rgrunber, line 42: Slight preference for something like `lemminx.cacheInRepoDir`, just so that it's properly namespaced and unlikely to clash with anything (though it would be very rare).

rgrunber: Also, just realized that

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/AbstractCacheBasedTest.java
```diff
@@ -37,7 +39,13 @@ public abstract class AbstractCacheBasedTest {
...
+			if (StringUtils.isEmpty(cacheInRepoDir)) {
```
rgrunber, line 43: With the example in the build file one would think `-DcacheInRepoDir=false` should disable it, but it doesn't. Maybe it's a silly thing to expect, but I would make the check simpler. Just test if the property is not null to activate it, and otherwise use the regular behaviour. This way the build file can be activated with just `-DcacheInRepoDir` (defining it makes it non-null).

.github/workflows/pr-verify.yml
```diff
@@ -0,0 +1,21 @@
+# This workflow will build a Java project with Maven, and cache/restore any dependencies to improve the workflow execution time
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: pr-verify
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: windows-latest
```
rgrunber, line 10: Can we enable on linux also and then we can simplify the Jenkins file and have it do just the release ?

datho7561:
Once you address Roland's feedback, I think it's good to merge.

rgrunber (approved):
Feel free to merge after the tests pass. I think this should address things. I would just verify afterwards that the Jenkins build either stops showing the "PR" tab or that it no longer triggers on pull requests.

datho7561 (approved):

PR #1359, opened 2022-11-02
Extend maxLineWidth setting to support attributes and tag elements

JessicaJHee:

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
@@ -20,6 +20,7 @@ import org.eclipse.lemminx.extensions.xsi.XSISchemaModel;
 import org.eclipse.lemminx.extensions.xsi.settings.XSISchemaLocationSplit;
 import org.eclipse.lemminx.services.extensions.format.IFormatterParticipant;
 import org.eclipse.lemminx.services.format.XMLFormatterDocumentNew;
+import org.eclipse.lemminx.services.format.XMLFormattingConstraints;
 import org.eclipse.lemminx.settings.XMLFormattingOptions;
 import org.eclipse.lemminx.utils.StringUtils;
 import org.eclipse.lemminx.utils.XMLBuilder;
@@ -159,12 +160,16 @@ public class XSIFormatterParticipant implements IFormatterParticipant {
 	}
 
 	@Override
-	public boolean formatAttributeValue(DOMAttr attr, XMLFormatterDocumentNew formatterDocument, int indentLevel,
-			XMLFormattingOptions formattingOptions, List<TextEdit> edits) {
+	public boolean formatAttributeValue(DOMAttr attr, XMLFormatterDocumentNew formatterDocument,
+			XMLFormattingConstraints parentConstraints, XMLFormattingOptions formattingOptions, List<TextEdit> edits) {
 
 		XSISchemaLocationSplit split = XSISchemaLocationSplit.getSplit(formattingOptions);
 
 		if (split == XSISchemaLocationSplit.none || !XSISchemaModel.isXSISchemaLocationAttr(attr.getName(), attr)) {
+			if (formatterDocument.isMaxLineWidthSupported()) {
+				parentConstraints
+						.setAvailableLineWidth(parentConstraints.getAvailableLineWidth() - attr.getValue().length());
+			}
 			return false;
 		}
 
@@ -197,11 +202,20 @@ public class XSIFormatterParticipant implements IFormatterParticipant {
 		int lineFeed = split == XSISchemaLocationSplit.onElement ? 1 : 2;
 		int locationNum = 1;
 		String attrValue = attr.getOriginalValue();
+		int lastAttrValueTermIndex = 0;
+		int availableLineWidth = parentConstraints.getAvailableLineWidth();
 
 		for (int i = firstContentOffset; i < attrValue.length(); i++) {
 			int from = formatterDocument.adjustOffsetWithLeftWhitespaces(attrValueStart, attrValueStart + i + 1);
 			if (Character.isWhitespace(attrValue.charAt(i)) && !Character.isWhitespace(attrValue.charAt(i + 1))
 					&& !StringUtils.isQuote(attrValue.charAt(from - attrValueStart))) {
+				availableLineWidth -= i - lastAttrValueTermIndex;
+				lastAttrValueTermIndex = i;
+				if (availableLineWidth < 0 && formatterDocument.isMaxLineWidthSupported()
+						&& !formattingOptions.isSplitAttributes()) {
+					indentSpaceOffset = (attrValueStart + 1) - attr.getNodeAttrName().getStart()
+							+ (parentConstraints.getIndentLevel() + 1) * tabSize;
+				}
 				// Insert newline and indent where required based on setting
 				if (locationNum % lineFeed == 0) {
 					formatterDocument.replaceLeftSpacesWithIndentationWithOffsetSpaces(indentSpaceOffset,
@@ -211,6 +225,11 @@ public class XSIFormatterParticipant implements IFormatterParticipant {
 				}
 				locationNum++;
 			}
+		}
+		if (formatterDocument.isMaxLineWidthSupported()) {
+			parentConstraints
+					.setAvailableLineWidth(formatterDocument.getMaxLineWidth() - (attrValueStart - indentSpaceOffset)
+							- attr.getValue().length());
 		}
 		return true;
 	}
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
@@ -34,38 +34,27 @@ public class DOMAttributeFormatter {
 
 	public void formatAttribute(DOMAttr attr, int prevOffset, boolean singleAttribute, boolean useSettings,
 			XMLFormattingConstraints parentConstraints, List<TextEdit> edits) {
+		int indentLevel = parentConstraints.getIndentLevel();
 		// 1) format before attribute name : indent left of the attribute name
 		// ex : <foo[space][space]attr=""
 		// --> <foo[space]attr=""
 		boolean alreadyIndented = false;
 		if (useSettings) {
-			int indentLevel = parentConstraints.getIndentLevel();
+			// move the attribute to a new line and indent it
 			if (isPreserveAttributeLineBreaks() && hasLineBreak(prevOffset, attr.getStart())) {
 				replaceLeftSpacesWithIndentation(indentLevel + 1, prevOffset, attr.getStart(), true, edits);
 				alreadyIndented = true;
 			} else if (isSplitAttributes() && !singleAttribute) {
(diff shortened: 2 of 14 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
```diff
@@ -143,4 +164,12 @@ private char getQuotationAsChar() {
 	private EnforceQuoteStyle getEnforceQuoteStyle() {
 		return formatterDocument.getSharedSettings().getFormattingSettings().getEnforceQuoteStyle();
 	}
+
+	private int getTabSize() {
+		return formatterDocument.getSharedSettings().getFormattingSettings().getTabSize();
+	}
+
+	private int getMaxLineWidth() {
+		return formatterDocument.getMaxLineWidth();
```
angelozerr, line 173: @JessicaJHee I think we should give the capability to igonre the maxLineWidth settings (to format without max line width). To do that there are 2 options : 

 * keep getMaxLineWidth as int and check if getMaxLineWidth is set a given value like -1 or 0 which means that getMaxLineWidth format setting should be ignored.
 * replace getMaxLineWidth with Integer to have the capability to set a value to null

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMElementFormatter.java
```diff
@@ -67,8 +68,8 @@ public void formatElement(DOMElement element, XMLFormattingConstraints parentCon
 			// Format end tag element with proper indentation
 			if (element.hasEndTag()) {
 				nb = formatEndTagElement(element, parentConstraints, constraints, edits);
+				parentConstraints.setAvailableLineWidth(constraints.getAvailableLineWidth() - nb);
```
angelozerr, line 71: replace `nb` variable name with an understand name.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/format/experimental/XMLFormatterMaxLineWithTest.java
```diff
@@ -69,6 +69,129 @@ public void noSplit() throws BadLocationException {
 		assertFormat(content, expected, 20);
 	}
 
+	@Test
+	public void splitWithAttribute() throws BadLocationException {
+		SharedSettings settings = new SharedSettings();
+		settings.getFormattingSettings().setTabSize(4);
```
angelozerr, line 75: My first impression when I see this code, is why this test doesn't configure maxLineWidth in the test although test tests this max line width. The answer is that it uses default value, but it is hard to understand that when you read the test. More if you change the default value for max line widg (ex : null), this test will not work again.

In otherwords, please define the max line width for each tests which tests max line width.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
```diff
@@ -78,7 +73,33 @@ public void formatAttribute(DOMAttr attr, int prevOffset, boolean singleAttribut
...
+			parentConstraints.setAvailableLineWidth(availableLineWidth - attributeNamelength);
+			formatAttributeValue(attr, parentConstraints, edits);
+		}
+
+		if (!alreadyIndented) {
+			int from = prevOffset;
+			int to = attr.getStart();
+			availableLineWidth = parentConstraints.getAvailableLineWidth();
+			if (availableLineWidth < 0 && !isSplitAttributes()
+					&& !XSISchemaModel.isXSISchemaLocationAttr(attr.getNodeName(), attr)) {
```
angelozerr, line 88: XSISchemaModel is an extension, so you should not use XSISchemaModel here. 

I wonder why you need to do a test with XSISchemaModel?

If you need really do this test, do that with IFormatterParticipant by adding it a new method.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/format/experimental/XMLFormatterMaxLineWithTest.java
JessicaJHee: The assertFormat method in this test takes in the maxLineWidth as an argument like this: 

`	private static void assertFormat(String unformatted, String expected, int maxLineWidth, TextEdit... expectedEdits)
`

but I can see how it is unclear, I can change it so that it is more explicitly defined if that makes more sense?

angelozerr: Oh I see. I think to be more consistent with other test. You should update the max line with in shared settings like tab size

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
JessicaJHee: I have made the changes so that if the maxLineWidth is set to 0, it is ignored.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -197,11 +200,20 @@ public boolean formatAttributeValue(DOMAttr attr, XMLFormatterDocumentNew format
...
+		int lastAttrValueTermIndex = 0;
+		int availableLineWidth = parentConstraints.getAvailableLineWidth();
 
 		for (int i = firstContentOffset; i < attrValue.length(); i++) {
 			int from = formatterDocument.adjustOffsetWithLeftWhitespaces(attrValueStart, attrValueStart + i + 1);
 			if (Character.isWhitespace(attrValue.charAt(i)) && !Character.isWhitespace(attrValue.charAt(i + 1))
 					&& !StringUtils.isQuote(attrValue.charAt(from - attrValueStart))) {
+				availableLineWidth -= i - lastAttrValueTermIndex;
+				lastAttrValueTermIndex = i;
+				if (availableLineWidth < 0 && formatterDocument.getMaxLineWidth() != 0
```
angelozerr, line 212: Instead of writing formatterDocument.getMaxLineWidth() != 0 I will introduce a new method  like

formatterDocument.isMaxLineWidthSupported()

It will give the capability to differentiate the use of value of maxLineWith and the check if maxLineWith is enabled.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
```diff
@@ -43,29 +44,23 @@ public void formatAttribute(DOMAttr attr, int prevOffset, boolean singleAttribut
 			if (isPreserveAttributeLineBreaks() && hasLineBreak(prevOffset, attr.getStart())) {
 				replaceLeftSpacesWithIndentation(indentLevel + 1, prevOffset, attr.getStart(), true, edits);
 				alreadyIndented = true;
+				availableLineWidth = getMaxLineWidth() - getTabSize() * (indentLevel + 1);
 			} else if (isSplitAttributes() && !singleAttribute) {
 				// move the attribute to a new line and indent it.
 				replaceLeftSpacesWithIndentation(indentLevel + getSplitAttributesIndentSize(), prevOffset,
 						attr.getStart(), true, edits);
 				alreadyIndented = true;
+				availableLineWidth = getMaxLineWidth() - getTabSize() * (indentLevel + getSplitAttributesIndentSize());
```
angelozerr, line 53: As we will have formatterDocument.isMaxLineWidthSupported()

I wonder if we should use it to avoid computing availableLineWidth

If yiu do this check it will be easy to retrieve in the code where available line width is computed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
angelozerr: Thanks!

But please wrap this check !=0 in a New method like
formatterDocument.isMaxLineWidthSupported()

angelozerr:
@JessicaJHee please fix the conflict

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
JessicaJHee: In my new changes I have tried to group the availableLineWidth computation together as much as possible but there are still a few that are missing that relies on more logic. I'm not sure if it is productive to do the `isMaxLineWidthSupported()` check for each time I modify availableLineWidth.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
```diff
@@ -78,7 +67,47 @@ public void formatAttribute(DOMAttr attr, int prevOffset, boolean singleAttribut
...
+
+		if (!alreadyIndented) {
+			int from = prevOffset;
+			int to = attr.getStart();
+			if (isMaxLineWidthSupported() && parentConstraints.getAvailableLineWidth() < 0
+					&& !isSplitAttributes()) {
+				replaceLeftSpacesWithIndentation(indentLevel + 1, from, to, true, edits);
+				parentConstraints.setAvailableLineWidth(
+						getMaxLineWidth() - getTabSize() * (indentLevel + 1) - attributeNamelength
+								- attr.getValue().length());
```
angelozerr, line 100: I wonder if you take care of this usecase with attribute:

 * attr=
 * attr="
 * attr         =
 * attr         =        "
 *  attr         =        "    "

angelozerr: I suggest that you write test with invalid attribute like I described below to check that we have not some NPE (I fear that attr.getValue() can return null in some invalid usecase attribute) and I have the impression that you don't take care of spaces beteween delimiter (=), but perhaps I'm wrong.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/xsi/participants/XSIFormatterParticipant.java
```diff
@@ -212,6 +224,9 @@ public boolean formatAttributeValue(DOMAttr attr, XMLFormatterDocumentNew format
 				locationNum++;
 			}
 		}
+		parentConstraints
+				.setAvailableLineWidth(formatterDocument.getMaxLineWidth() - (attrValueStart - indentSpaceOffset)
+						- attr.getValue().length());
 		return true;
 	}
```
angelozerr, line 232: if formatterDocument.isMaxLineWidthSupported()

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMAttributeFormatter.java
angelozerr: I like your change, but to be consistent I think you should do that. It will help to retrieve where max lien width is computed by finding references from formatterDocument.isMaxLineWidthSupported() with an IDE

angelozerr:
Ok let's merge the PR and fix corner usecases like miss of attribute value in a separate PR. Thanks @JessicaJHee !

PR #1369, opened 2022-11-14
Improve full support for max line width

### LINKED ISSUE #1248: Full support for max line width with experimental formatter
The max line width from the experimental formatter https://github.com/redhat-developer/vscode-xml/blob/master/docs/Formatting.md#xmlformatmaxlinewidth doesnt take care of attributes and tag elements. 

Today it works only with text node but we should take care of other nodes like attributes or tag elements.

JessicaJHee:
Fixes #1248 


datho7561 (changes requested):
I found two things to nitpick about, but otherwise looks good and works well!

PR #1371, opened 2022-11-16
Referenced RelaxNG schema not found via XML catalog

angelozerr:
Referenced RelaxNG schema not found via XML catalog

Fixes https://github.com/redhat-developer/vscode-xml/issues/804


datho7561 (approved):
Looks good and works for me!

PR #1372, opened 2022-11-17
Improve performance with grammar aware formatting

### LINKED ISSUE #1368: Experimental formatter is slower than current formatter with large XML file
There are interesting large TEI XML files at https://www.wwp.northeastern.edu/outreach/seminars/_current/handouts/tei_samples/index.html

If you download 
 * the XML file https://www.wwp.northeastern.edu/outreach/seminars/_current/handouts/tei_samples/proust/recherche-v1-p1.xml
 * the DTD file https://www.wwp.northeastern.edu/outreach/seminars/_current/handouts/tei_samples/proust/tei_tite.dtd

And you try to format the XML with experimental formatter, it takes some times although with the current formatter it is very q

We need to investigate the performance issue with this sample.

 * with experimental formatter, it takes around 1323ms.
 * with the current formatter, it takes around 58ms.

When grammar aware formatting is disabled, it takes around 148ms.

JessicaJHee:
Fixes #1368


JessicaJHee:
On my machine without this PR, it takes: 
* 97ms with the old formatter
* 320ms with the experimental formatter
* 126ms with the experimental formatter with grammar aware formatting set to false

Currently with this PR, it takes: 
* 165ms with the experimental formatter
* 135ms with the experimental formatter with grammar aware formatting set to false

angelozerr:
> 165ms with the experimental formatter

Nice! The performance is divided by 2.

> 135ms with the experimental formatter with grammar aware formatting set to false

It should be interesting to see (in an another PR) how to improve that again.

> 97ms with the old formatter

It should be interesting to see which process can take more time than old formatter. I wonder if we merge all TextEdit in one if it will improve performance. HTML formatter does that when there are too many TextEdit which are generated.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
```diff
@@ -40,8 +43,11 @@ public class ContentModelFormatterParticipant implements IFormatterParticipant {
 
 	private final ContentModelManager contentModelManager;
 
+	private final Map<String, Collection<CMDocument>> cmDocumentsCache;
+
 	public ContentModelFormatterParticipant(ContentModelManager contentModelManager) {
 		this.contentModelManager = contentModelManager;
+		cmDocumentsCache = Collections.synchronizedMap(new HashMap<>());
```
angelozerr, line 50: The cache should be not hosted in ContentModelFormatterParticpant but it should be hosted in a Map initialized by the formatter. Because when you implement a cache, there is the problem of the cache evict.


I will change the signature of getFormatElementCategory like this:

```java
public FormatElementCategory getFormatElementCategory(DOMElement element,
			XMLFormattingConstraints parentConstraints, Map<String, Object> formattingContext, SharedSettings sharedSettings) {
```

The formattingContext is a kind of cache that any formatter participant could use. After that I will store your map which stores mapping between namespaceURI and CMDocument collections not directly in formattingContext, The formattingContext should store with a given key (ex : CM_DOCUMENTS_MAPPINGS_KEY ) your cmDocumentsCache.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/dtd/contentmodel/CMDTDDocument.java
```diff
@@ -200,6 +200,9 @@ private static int getEntityNameStartColumnNumber(String entityName, ScannedEnti
 			char[] ch = scannedEntity.ch;
 			int wordIndex = entityName.length(); //
 			int startEntityNameIndex = -1;
+			if (endEntityIndex > ch.length) {
```
angelozerr, line 203: Why you did that?

Perhaps you try to fix an another bug. If it that please rénové this code and create an another PR

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
```diff
@@ -92,4 +104,16 @@ public boolean shouldCollapseEmptyElement(DOMElement element, SharedSettings sha
 		}
 		return true;
 	}
+
+	private Collection<CMDocument> getCMDocumentFromCache(String key,
+			Map<String, Collection<CMDocument>> formattingContext) {
+		Collection<CMDocument> cmDocuments = null;
+		cmDocuments = formattingContext.get(key);
```
angelozerr, line 111: return formattingContext.get(key);

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/dtd/contentmodel/CMDTDDocument.java
JessicaJHee: This was the fix for the exception that Fred brought up here: https://github.com/eclipse/lemminx/issues/1368#issuecomment-1313924572

Should I do that in another PR?

angelozerr: Ok let's keep like this.

datho7561 (changes requested):

datho7561:
I tried formatting on `content.xml` with this PR (a large XML file which doesn't have a grammar) and I got the following numbers:
- experimental with grammar aware: 1338ms
- experimental without grammar aware: 1257ms
- old formatter: 1710ms

Although I only did one through of each so maybe these times have high varience.

PR #1374, opened 2022-11-21
XML Completion element apply generate invalid XML content

### LINKED ISSUE #1373: XML Completion element apply generate invalid XML content
The XML completion apply for XML element should be really improved in the case when element already exists.

Take an XML file which is bound with a grammar which defines foo as element

`<|` will generate `<foo></foo>`, it is good.

But now if you open completion 

`<fo|o></foo>`

and you apply the completion, it will generate

`<foo></foo></foo>` The completion apply should update the start and end tag element only. To do that we need to use `additionalTextEdits`. We should generate one TextEditor which update the start tag element and one in additionalTextEdits which updat ethe end tag element with foo.

We should consider too when element have attributes

`<foo attr="" ></foo>`

If you open  completion :

`<fo|o attr="" ></foo>`

and you apply completion, it generates 

`<foo></foo> attr="" ></foo>`

angelozerr:
XML Completion element apply generate invalid XML content

Fixes #1373

This PR requires https://github.com/redhat-developer/vscode-xml/pull/815


angelozerr:
This PR requires https://github.com/redhat-developer/vscode-xml/pull/815 to know if linked editing is activated or not.

Here a demo with XML apply completion which update existing end tag instead of creating a new end tag 

It is managed with additiontextEdits when linked editing is not activated and with nothing when linked editing is enabled

ApplyCompletionImprove

This PR resolve : 

 * the documentation of element coming from a grammar
 * the addition text edits which update the end tag if need

Test has been adjusted but I need towrite more tests

datho7561:
One thing I ran into was with the `substitution.xsd`. If I have:

```xml
<?xml-model href="substitutionGroupSample.xsd"?>
<BookStore>

  <Na|me>HJKL</Name>
  <Book>
    <Title>hjkl</Title>
    <Author>me</Author>
    <Date>2019</Date>
    <ISBN>123456789</ISBN>
    <Publisher>me</Publisher>
  </Book>

</BookStore>
```

If I open completion and apply the `StoreName` element completion, then the closing tag becomes `</StoreNameStoreName>`. However, doing the `Name` completion works as expected.

angelozerr:
I think you are in linked editing mode. Do you use https://github.com/redhat-developer/vscode-xml/pull/815 ?

angelozerr:
There is one limitation that I don't know how to fix it. If you activate linked editing and you go to a  start tag element, and you open completion quickly, vscode have no time to activate the linked editing range in the editor, and you will have the bug that you encounter.

Ideally we should know when linked editing is activated really in the editor (not only activate linked editing enable by settings, this information that I'm using) and completion should use information about activation.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/completion/AttributeValueCompletionResolver.java
```diff
@@ -69,4 +71,11 @@ private static MarkupContent getDocumentationForAttributeValue(CMElementDeclarat
 		return null;
 	}
 
+	@Override
+	protected void resolveCompletionItem(ICompletionItemResolverRequest request, CompletionItem toResolve,
+			DOMText text) {
+		// TODO Auto-generated method stub
```
datho7561, line 77: Add an implementation or change the comment to explain why an implementation isn't needed.

datho7561:
Maybe I'm missing something, but completion documentation doesn't appear to be working.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/data/DataEntryField.java
```diff
@@ -150,4 +154,31 @@ public static Integer getPropertyAsInt(Object data, String fieldName) {
...
+	/**
+	 * Returns the value of the field as an boolean or null if the field is empty.
+	 *
+	 * @param data      the object to read the field of
+	 * @param fieldName the field to read
+	 * @throws ClassCastException if the field is not an boolean
+	 * @return the value of the field as an boolean or null if the field is empty
+	 */
+	public static Boolean getPropertyAsBoolean(Object data, String fieldName) {
+		// FIXME: catch exception return null
```
datho7561, line 166: Address the `FIXME:` comment

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/completion/AbstractCompletionResolver.java
```diff
@@ -9,58 +9,61 @@
...
 	 * @param toResolve     the unresolved completion item to resolve
 	 * @param parentElement the parent element to where completion was opened on
 	 * @param attr          the attribute that completion was opened on, may be null
 	 */
-	protected abstract void addDocumentationToCompletion(ICompletionItemResolverRequest request,
+	protected abstract void resolveCompletionItem(ICompletionItemResolverRequest request,
 			CompletionItem toResolve,
 			DOMElement parentElement, DOMAttr attr);
 
+	protected abstract void resolveCompletionItem(ICompletionItemResolverRequest request, CompletionItem toResolve,
```
datho7561, line 67: Add a javadoc for this method

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/completion/ICompletionItemResolverRequest.java
```diff
@@ -57,18 +57,50 @@ public interface ICompletionItemResolverRequest extends ISharedSettingsRequest,
...
+	public boolean isResolveAdditionalTextEditsSupported();
+
 	/**
 	 * Returns the shared settings.
 	 *
 	 * @return the shared settings.
 	 */
 	SharedSettings getSharedSettings();
 
+	Integer getCompletionOffset();
```
datho7561, line 104: Add a Javadoc

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/completion/ICompletionRequest.java
```diff
@@ -58,10 +58,23 @@ public interface ICompletionRequest extends IPositionRequest, ISharedSettingsReq
...
+	public boolean isResolveAdditionalTextEditsSupported();
+
 	/**
 	 * Returns the proper insert text format according the support of snippet.
 	 *
 	 * @return the proper insert text format according the support of snippet.
 	 */
 	InsertTextFormat getInsertTextFormat();
+
+	Range getReplaceRangeForTagName();
```
datho7561, line 79: Add a Javadoc

datho7561 (changes requested):
Okay, I found a few things in the code, but overall it seems to be working great!

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/data/DataEntryField.java
angelozerr: There is this FIXME because I copied / pasted from getPropertyAsInt, but I don't know the problem. Could we let like this like existing getPropertyAsInt?

datho7561: Okay. Thinking back to it, this shouldn't be a huge problem, since this is data that we generate, so it should always have the right type.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/extensions/completion/ICompletionItemResolverRequest.java
angelozerr: fixed

datho7561 (approved):
Looks good. Thanks Angelo!

PR #1376, opened 2022-11-22
Remove preserveEmptyContent setting from experimental formatter

### LINKED ISSUE #1346: Consider removing xml.format.preserveEmptyContent for experimental formatter
The experimental formatter introduces a new element category called `preserve space`, which preserves all white spaces within the specified element tag. Currently, when `preserveEmptyContent = true`, it has the same behavior as `preserve space`, making this setting redundant. 

### Note: consider its behavior when used alongside with other settings: 

https://github.com/redhat-developer/vscode-xml/blob/main/docs/Formatting.md#xmlformatpreservednewlines
* **preservedNewLines**: this setting is already prioritized over `preserveEmptyContent`, it's behavior would be unaffected with the removal of `preserveEmptyContent`. It formats the xml nicely while allowing the user to set a specified number of new lines they would like to keep. 

_Example:_ 
Before with `preservedNewLines =  2`: 
```
<foo>

test1           test2


    test3
</foo>
```

After: 
```
<foo>

    test1 test2


    test3
</foo>
```

https://github.com/redhat-developer/vscode-xml/blob/main/docs/Formatting.md#xmlformatjoincontentlines
* **joinContentLines** If an user wants to normalize the white spaces with in an empty tag, they can use `xml.format.joinContentLines` instead, which currently has lower priority compared to `preserveEmptyContent` and the `preserve space` category.

_Example:_ 
Before with `joinContentLines =  true` and `preserveEmptyContent = false`: 
```
<a>

</a>
```

After: 
```
<a> </a>
```

JessicaJHee:
Fixes #1346 


datho7561 (approved):
Looks good. Thank you, Jessica!

PR #1378, opened 2022-11-24
Autoclose tag generates an unexpected >

### LINKED ISSUE #1221: Autoclose tag generates an unexpected >
Given this XML file:

`<foo</foo>`

when we close the foo start tag:

`<foo>`

the auto class tag (I think) generates a `>`

`<foo</foo>>`

angelozerr:
Autoclose tag generates an unexpected >

Fixes #1221

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLLinkedEditing.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLLinkedEditing.java
@@ -33,16 +33,6 @@ import org.eclipse.lsp4j.jsonrpc.CancelChecker;
 class XMLLinkedEditing {
 
 	private static Logger LOGGER = Logger.getLogger(XMLLinkedEditing.class.getName());
-
-	// Full XML Element name pattern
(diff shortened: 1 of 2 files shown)


angelozerr:
@vrubezhny @mickaelistria this PR removes the wordPattern which fixes https://github.com/eclipse/lemminx/issues/1221 issue.

I have tried quickly with WWD and without this PR, there is the same problem than vscode.

It seems that you improve LSP4e linked editing, so I would like to apply this PR:

 * to be consistent with HTML LS (linked editing doesn't return wordPattern)
 * to have a JSON response which is smaller
 * to avoid the issue https://github.com/eclipse/lemminx/issues/1221

datho7561 (approved):
This works for me. Thank you, Angelo!

angelozerr:
Thanks @datho7561 for your review.

@vrubezhny @mickaelistria can we merge the PR?

angelozerr:
@mickaelistria @vrubezhny I suggest we merge the PR since HTML LS works without Word pattern. 

If there are some problem in eclipse ide we will use again Word pattern but by taking care of the fixed issue.

PR #1381, opened 2022-11-28
Set experimental formatter to default

JessicaJHee:
* removes xml.format.experimental setting
* introduces new xml.format.legacy to revert to old formatter
* renames and moves tests

Fixes #1377

Signed-off-by: Jessica He <jhe@redhat.com>

PR #1385, opened 2022-11-29
Use experimental formatter for grammar generation tests

### LINKED ISSUE #1382: Generate grammar fails on first attempt with experimental formatter
1. Clear the lemminx cache
2. Use the experimental formatter
3. Make an XML file with some content that is not associated with any schema
4. Use the "generate schema and associate" quick fix (you should be able to use any of the generate quick fixes)
5. Nothing will happen
In the logs, you should see a `org.eclipse.lemminx.uriresolver.CacheResourceDownloadingException`.
6. Try again, this time it should succeed.

datho7561:
Also fixes some bugs with grammar aware formatting and the grammar generator.

Closes #1382

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
@@ -13,6 +13,9 @@ package org.eclipse.lemminx.extensions.contentmodel.participants;
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.extensions.contentmodel.model.CMDocument;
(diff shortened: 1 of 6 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelFormatterParticipant.java
```diff
@@ -61,7 +62,16 @@ public FormatElementCategory getFormatElementCategory(DOMElement element,
...
 		if (cmDocuments == null) {
-			cmDocuments = contentModelManager.findCMDocument(element);
+			try {
+				cmDocuments = contentModelManager.findCMDocument(element);
+			} catch (CacheResourceDownloadingException e) {
+				// block; we need the content model to format correctly.
+				// if we don't block and instead return null here,
+				// then some parts of the document will be
+				// grammar-aware formatted and others won't be.
+				e.getFuture().join();
```
angelozerr, line 72: Generaly we should avoid using join which can block the thread pooling.

I think the proper solution is to use CompletableFuture (get cmDocuments) and thenApply (format), but I fear it is a big refactoring.

Perhaps it should be safer to use a timeout (e.getFuture().get(2000) instead of join?

JessicaJHee (approved):
I was able to reproduce the issue and it works well now with this fix! Just need to update this test: `schemaWithExperimentalFormatter` to test with the old formatter instead when we switch to the experimental formatter in this PR: https://github.com/eclipse/lemminx/pull/1381

datho7561:
Ok. Do you think I should do that in this PR or do you want to handle it in #1381 ?

JessicaJHee:
> Ok. Do you think I should do that in this PR or do you want to handle it in #1381 ?

Will do in #1381

PR #1389, opened 2022-12-05
Wrap element

angelozerr:
Wrap element

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMParser.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMParser.java
@@ -98,7 +98,30 @@ public class DOMParser {
 			}
 			if (previousTokenWasEndTagOpen) {
 				previousTokenWasEndTagOpen = false;
-				if (token != TokenType.EndTag) {
+				boolean linkToEmptyStartTag = false;
+				if (token == TokenType.EndTagClose) {
+					// ex : </>
+					String closeTag = null;
+
+					/**
+					 * Try to find a start element with no tag : <>
+					 */
+					while (!(curr.isElement() && ((DOMElement) curr).isSameTag(closeTag)) && curr.parent != null) {
+						curr.end = endTagOpenOffset;
+						curr = curr.parent;
+					}
+					if (curr != xmlDocument) {
+						linkToEmptyStartTag = true;
+						curr.closed = true;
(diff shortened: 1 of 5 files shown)


angelozerr:
Tests exists now, @datho7561 please review it again.

datho7561:
If I used linked editing, I can add spaces and newlines into the element name

```xml
<elt|></elt>
```

If I type in `<space>hello<enter>`, then I get:

```xml
<elt hello
></elt hello
  >
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMParser.java
```diff
@@ -98,559 +98,583 @@ public DOMDocument parse(TextDocument document, URIResolverExtensionManager reso
 			}
 			if (previousTokenWasEndTagOpen) {
 				previousTokenWasEndTagOpen = false;
-				if (token != TokenType.EndTag) {
+				boolean linkToEmptyStartTag = false;
+				if (token == TokenType.EndTagClose) {
+					// ex : <\>
```
datho7561, line 103: `</>`

datho7561 (changes requested):
If you are okay with the behaviour of the linked editing as a result of the parser changes, then I think the PR is good. However, it seems a bit buggy to me. I noticed a small typo in some documentation

angelozerr:
> If I used linked editing, I can add spaces and newlines into the element name

I cannot reproduce, could you do a demo please.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/dom/DOMParser.java
angelozerr: fixed

datho7561:
surround-with-weird-linked-editing

angelozerr:
Oh I see what you mean.  In this case you are using snippet and not linked editing range. I'm not  sure that we can control that.

If you wish to test with linked editing range support, do like your demo to generate the start / end tag element without tag and press on Escape to quit the snippet mode. Type some content in the tag and linked editing range will be activated in this case and you should not have the issue of your demo.

datho7561 (approved):
> In this case you are using snippet and not linked editing range. 

Okay. Sorry it looks like I just misunderstood what was happening. This looks good to me! Thanks, Angelo!

PR #1390, opened 2022-12-06
Add codelens and command to register/unregister catalog

JessicaJHee:
Peek 2022-12-06 12-41

Fixes https://github.com/redhat-developer/vscode-xml/issues/572

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java
@@ -132,9 +132,20 @@ public class XMLTextDocumentService implements TextDocumentService {
 	/**
 	 * Save context.
 	 */
-	class SaveContext extends AbstractSaveContext {
+	public class SaveContext extends AbstractSaveContext {
 
 		private final Collection<ModelTextDocument<DOMDocument>> documentsToValidate;
+
+		private boolean isRefreshCodeLenses;
+
+		public boolean isRefreshCodeLenses() {
+			return isRefreshCodeLenses;
+		}
+
+		public void setRefreshCodeLenses(boolean isRefreshCodeLenses) {
+			this.isRefreshCodeLenses = isRefreshCodeLenses;
+			return;
+		}
 
 		public SaveContext(Object settings) {
 			super(settings);
@@ -588,6 +599,9 @@ public class XMLTextDocumentService implements TextDocumentService {
 	void doSave(SaveContext context) {
 		getXMLLanguageService().doSave(context);
 		context.triggerValidationIfNeeded();
+		if (context.isRefreshCodeLenses()) {
+			xmlLanguageServer.getLanguageClient().refreshCodeLenses();
+		}
 	}
 
 	private void triggerValidationFor(Collection<ModelTextDocument<DOMDocument>> documents) {
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/ClientCommands.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/client/ClientCommands.java
@@ -43,4 +43,8 @@ public class ClientCommands {
 	 */
 	public static final String OPEN_URI = "xml.open.uri";
 
+	/**
+	 * Update configuration
+	 */
+	public static final String UPDATE_CONFIGURATION = "xml.update.configuration";
 }
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/config/ConfigurationItemEdit.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/config/ConfigurationItemEdit.java
@@ -1,0 +1,77 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.commons.config;
+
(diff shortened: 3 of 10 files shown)


angelozerr:
It looks promising @JessicaJHee !

I noticed that you created an custom command on client side for only XML catalog.

My idea was to do like we did for vscode-microprofile, to provide a generic command which updates a given configuration on client side.

 * in vscode-microprofile side, we have https://github.com/redhat-developer/vscode-microprofile/blob/7d6cac22917cb47af8c04151269432b9999dd2be/src/lsp-commands.ts#L24
 * on LS side, we update settings with code action https://github.com/eclipse/lsp4mp/blob/fd508c6b2efc5c9b75e3355a6c88ef427dcff4b1/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/commons/MicroProfileCodeActionFactory.java#L91

The main idea is to provide a generic command which update settings on client side. If you think it is doable, it should be nice otherwise let's give up my idea, but I think if we have more and more codeaction, codelens which need to update settings we should change that.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/ConfigurationItemEdit.java
```diff
@@ -0,0 +1,76 @@
...
+	private Object value;
+	private ConfigurationItemEditType editType;
+	private ConfigurationItemValueKind valueKind;
+
+	/**
+	 *
+	 * @param section   config section to change
+	 * @param operation type of change
+	 * @param value     the value for the change
+	 */
```
datho7561, line 31: add an entry for `valueKind`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/ConfigurationItemEditType.java
```diff
@@ -0,0 +1,21 @@
...
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.extensions.contentmodel.utils;
+
+/**
+ * Configuration item edit type. Represents a configuration edit operation
+ */
+public enum ConfigurationItemEditType {
+	add, delete
```
datho7561, line 20: The names of enums in the language server protocol generally have the first letter of their name capitalized, and the Java convention is usually to use all caps. I think it might be a good idea to use one of those conventions in order to be consistent.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/ConfigurationItemValueKind.java
```diff
@@ -0,0 +1,21 @@
...
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.extensions.contentmodel.utils;
+
+/**
+ * Configuration item value kind. Represents the type of data that is being sent.
+ */
+public enum ConfigurationItemValueKind {
+	file
```
datho7561, line 20: Same here

datho7561 (changes requested):

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/ConfigurationItemEditType.java
```diff
@@ -0,0 +1,21 @@
...
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.extensions.contentmodel.utils;
```
angelozerr, line 14: Please move to `org.eclipse.lemminx.commons.config`

angelozerr:
@JessicaJHee please fix conflicts

angelozerr:
It should be really nice to refresh coelens when `xml.catalogs` changed (on server side).

To do that I suggets to add this following code after https://github.com/eclipse/lemminx/blob/de9c702be26f5380be84b3a38d6d16f6913011df/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java#L590

```java
if (context.isRefreshCodeLenses()) {
	xmlLanguageServer.getLanguageClient().refreshCodeLenses();
}
```

`isRefreshCodeLenses() `is a new method which returns a boolean to know if codelens must be refreshed or not when a settings a changed.

After https://github.com/eclipse/lemminx/blob/74689af7f882732c9e59f27ce23ea0652fe2dbc5/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/ContentModelPlugin.java#L139

you update the refresh code lens flag 

`context.setRefreshCodeLenses(true)`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/ConfigurationItemEdit.java
```diff
@@ -0,0 +1,76 @@
...
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.extensions.contentmodel.utils;
```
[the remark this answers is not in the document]
angelozerr: It seems you don't move this class to `org.eclipse.lemminx.commons.config` package.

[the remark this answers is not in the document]
JessicaJHee: sorry I missed that, it is moved now :+1:

angelozerr:
It works like a charm. Thanks @JessicaJHee !

PR #1392, opened 2022-12-07
Adjust selection range for wrap element when empty selection is in start or end tag.

angelozerr:
Adjust selection range for wrap element when empty selection is in start or end tag.


angelozerr:
Here a demo:

SurroundWithTagsAdjust

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/commands/SurroundWithCommand.java
```diff
@@ -138,9 +141,6 @@ protected Object executeCommand(DOMDocument document, ExecuteCommandParams param
 					startText.append(tag);
 				}
 				startText.append(">");
-				if (emptySelection && snippetsSupported) {
```
datho7561, line 141: Do you think this feature isn't helpful? If so, why?

datho7561 (changes requested):

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/commands/SurroundWithCommand.java
angelozerr: Ok I support again this pattern

datho7561 (approved):
Looks good! Thanks Angelo!

PR #1400, opened 2022-12-11
Codelens, completion, definition, documentLink, highlighting, references, rename support in RNG file

angelozerr:
Codelens, completion, definition, documentLink, highlighting, references, rename support in RNG file

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/grammar/rng/RNGRenameParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/grammar/rng/RNGRenameParticipant.java
@@ -1,0 +1,129 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.extensions.relaxng.grammar.rng;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.eclipse.lemminx.dom.DOMAttr;
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.dom.DOMElement;
+import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.dom.DOMRange;
+import org.eclipse.lemminx.extensions.relaxng.utils.RelaxNGUtils;
+import org.eclipse.lemminx.services.extensions.IRenameParticipant;
+import org.eclipse.lemminx.services.extensions.IRenameRequest;
+import org.eclipse.lemminx.utils.DOMUtils;
+import org.eclipse.lemminx.utils.XMLPositionUtility;
+import org.eclipse.lsp4j.Location;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.TextEdit;
+
+/**
+ * RNG rename
+ * 
+ */
+public class RNGRenameParticipant implements IRenameParticipant {
+
+	@Override
+	public void doRename(IRenameRequest request, List<TextEdit> locations) {
+		DOMDocument xmlDocument = request.getXMLDocument();
+
+		if (!DOMUtils.isRelaxNGXMLSyntax(xmlDocument)) {
+			return;
+		}
+		locations.addAll(getRenameTextEdits(request));
+
+	}
+
+	private List<TextEdit> getRenameTextEdits(IRenameRequest request) {
+		DOMDocument document = request.getXMLDocument();
+		DOMNode node = request.getNode();
+
+		if (!node.isAttribute()) {
+			return Collections.emptyList();
+		}
+
+		DOMAttr attr = (DOMAttr) node;
+		DOMElement ownerElement = attr.getOwnerElement();
+
+		if (ownerElement == null) {
+			return Collections.emptyList();
+		}
+
+		String newText = request.getNewText();
+
+		if (RelaxNGUtils.isDefine(ownerElement)) {
+			// <define
+			if (attr.getName().equals("name")) {
+				// <define name=""
+				List<Location> locations = getReferenceLocations(ownerElement);
+				return renameAttributeValueTextEdits(document, attr, newText, locations);
+			}
+		}
+
+		return Collections.emptyList();
+	}
+
+	private List<Location> getReferenceLocations(DOMNode node) {
+
+		List<Location> locations = new ArrayList<>();
+
+		RelaxNGUtils.searchRNGOriginAttributes(node,
+				(origin, target) -> locations.add(XMLPositionUtility.createLocation(origin.getNodeAttrValue())),
+				null);
+
+		return locations;
+	}
+
+	private List<TextEdit> renameAttributeValueTextEdits(DOMDocument document, DOMAttr attribute, String newText,
+			List<Location> locations) {
+		DOMRange attrValue = attribute.getNodeAttrValue();
+		List<TextEdit> textEdits = new ArrayList<>();
+
+		int valueStart = attrValue.getStart();
+		int valueEnd = attrValue.getEnd();
+		Range range = XMLPositionUtility.createRange(valueStart, valueEnd, document);
+
+		// make range not cover " on both ends
+		reduceRangeFromBothEnds(range, 1);
+
+		textEdits.add(new TextEdit(range, newText));
+
+		for (Location location : locations) {
+			Range textEditRange = location.getRange();
+			reduceRangeFromBothEnds(textEditRange, 1);
+
+			TextEdit textEdit = new TextEdit(textEditRange, newText);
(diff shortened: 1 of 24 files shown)


angelozerr:
I need to write tests, but this PR should work and provide the same feature than XSD : 

 * codelens : it should display the number of references of `define` referenced by `ref` element. A click on this codelens, opens the references
 * completion: completion on ref/@name to show all define/@name
 * documentLink for include/@ref
 * definition : jump to ref/@name to the proper define/@name
 * higlighting :
   *  if define/@name has the cursor it should highlight all referenced ref/@name.
   *  if ref/@name has the cursor it should highlight the referenced ref/@name
 * rename : rename define/@name should rename all referenced ref/@name
 * references: find references from define/@name find all references ref/@name

You can play with the tei_all.rng which is used for our test.

angelozerr:
Here a screenshot with codelens in tei_all.rnc:

image

click on codelens open the references dialog to show all ref/@name which reference the define:

image

completion of ref/@name show all define/@name:

image

Here a demo with definition:

RNGDefinition

angelozerr:
@adunning @otichy I think this feature could interest you.

angelozerr:
Tests are now written, the PR can be reviewed.

angelozerr:
Please review too the PR from vscode-xml https://github.com/redhat-developer/vscode-xml/pull/834 which add doc about those features

datho7561:
If I try to rename `asdf` from the `|`:

```xml
  <start>
    <ref name="root-element"/>
  </start>
  <define name="root-element">
      <element name="root-element">
        <ref name="as|df"></ref>
      </element>
  </define>
  <div>
    <div>
      <define name="asdf">
        <attribute name="child"></attribute>
      </define>
    </div>
  </div>
</grammar>
```

Nothing happens. However, it does work if I rename from line 13 (where it's defined).

datho7561:
Similarily, find all references from `<ref name="as|df"></ref>` doesn't work.

angelozerr:
The reference is for ref/@name -> define/@name, so rename is working for define/@name and not ref/@name, it the same thing for find references for ref/@name which is not implemented because a ref cannot be referenced, it is a define which can be referenced.

I suggest that you see my demo https://github.com/redhat-developer/vscode-xml/pull/834/files#diff-7b6c8333ff3310f051cc3ff2c08dfaa2d6ac2b2cf71149ef68b89cdf270f371c

angelozerr:
> Okay. I understand why "find all references" doesn't show the definition, but I think that renaming the reference should rename the definition.

I agree, but I wanted to be consistent with XSD which does that. Could we do that in an another PR to fix XSD too?

datho7561:
> I agree, but I wanted to be consistent with XSD which does that. Could we do that in an another PR to fix XSD too?

Okay.

datho7561:
I also noticed that the"# references" CodeLens doesn't work if the `define` element is nested under a `div` element like in the schema I sent.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/grammar/rng/RNGRenameParticipant.java
```diff
@@ -0,0 +1,149 @@
...
+			reduceRangeFromBothEnds(textEditRange, 1);
+
+			String oldAttrValue;
+			try {
+				oldAttrValue = getAttrTextValueFromPosition(document, location.getRange().getStart());
+			} catch (BadLocationException e1) {
+				return Collections.emptyList();
+			}
+
+			int colonIndex = oldAttrValue.indexOf(":");
```
datho7561, line 116: According to the diagnostics, you can't have colons in the `name` attribute. There is a separate attribute, `ns`, to specify the namespace.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/utils/RelaxNGUtils.java
```diff
@@ -38,4 +68,307 @@ private static String getRelaxNGURI(RelaxNGGrammar grammar) {
 		return grammar.getGrammarDescription().getExpandedSystemId();
 	}
 
+	/**
+	 * Binding type of rng grammar
```
datho7561, line 72: Is this enum needed? It seems like DEFINE is used to signal that an element is `ref` with a `name` attribute. Are there other binding types that you think you may add? Otherwise, maybe it's better to rename `getBindingType` to `isRefBoundToName` and make it boolean.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/utils/RelaxNGUtils.java
```diff
@@ -38,4 +68,307 @@ private static String getRelaxNGURI(RelaxNGGrammar grammar) {
...
+				if (documentElement == null) {
+					break;
+				}
+				Node parent = documentElement;
+				addTargetAttrs(parent, referencedNodes);
+		}
+		return referencedNodes;
+	}
+
+	public static void addTargetAttrs(Node parent, List<DOMAttr> referencedNodes) {
```
datho7561, line 279: Add a Javadoc

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/utils/RelaxNGUtils.java
```diff
@@ -38,4 +68,307 @@ private static String getRelaxNGURI(RelaxNGGrammar grammar) {
...
 
+	/**
+	 * Binding type of rng grammar
+	 *
+	 */
+	public enum BindingType {
+		DEFINE, NONE;
+	}
+
+	public static BindingType getBindingType(DOMAttr originAttr) {
```
datho7561, line 79: Add a Javadoc

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/DOMUtils.java
```diff
@@ -89,21 +91,56 @@ public static boolean isXSD(String uri) {
...
+
+	/**
+	 * Returns true if the given URI is a RelaxNG grammar XML syntax and false
+	 * otherwise.
+	 * 
+	 * @param uri the URI to check
+	 * @return true if the given URI is a RelaxNG grammar XML syntax and false
+	 *         otherwise.
+	 */
+	public static boolean isRelaxNGXMLSyntax(String uri) {
```
datho7561, line 132: I think it would be a good idea to rename this function to incorporate "uri" in the name, so it's obvious that this doesn't try to check the namespace like the other `isRelaxNGXMLSyntax does

datho7561 (changes requested):
I found a few things, but appears to be working well.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/utils/RelaxNGUtils.java
angelozerr: I would to keep this enum to be consistent with XSD and perhaps there is another binding type.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/grammar/rng/RNGRenameParticipant.java
angelozerr: Good catch! I remove this code (which was a copy/paste from XSD rename)

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/DOMUtils.java
angelozerr: What about `isRelaxNGUriXMLSyntax` ?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/utils/RelaxNGUtils.java
angelozerr: fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/utils/RelaxNGUtils.java
angelozerr: fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/DOMUtils.java
angelozerr: fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/relaxng/utils/RelaxNGUtils.java
datho7561: Okay. Let's keep the enum then. I think you should improve the javadoc to better describe what the enum represents.

angelozerr: I add a comment.

datho7561 (approved):
Looks good, and it's very helpful for editing .rng grammars. Thanks, Angelo!

PR #1408, opened 2022-12-13
Bind to a schema from an empty document

datho7561:
Allows you to use the code action to bind to a schema when a document doesn't have a root element. Also, the bind schema CodeLens is shown when the document doesn't have a root element.

It generates an element `placeholder-element-name` when the binding
strategy requires a root element to function (eg. `schemaLocation`, `noNamespaceSchemaLocation`). It also uses `placeholder-element-name` when binding to a `.dtd` using a DOCTYPE declaration.

Closes redhat-developer/vscode-xml#819

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/nogrammarconstraints/NoGrammarConstraintsCodeAction.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/nogrammarconstraints/NoGrammarConstraintsCodeAction.java
@@ -46,7 +46,6 @@ import org.eclipse.lsp4j.CodeActionKind;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.TextDocumentEdit;
-import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lsp4j.jsonrpc.CancelChecker;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 
@@ -58,6 +57,8 @@ import com.google.gson.JsonObject;
 public class NoGrammarConstraintsCodeAction implements ICodeActionParticipant {
 
 	private static final Logger LOGGER = Logger.getLogger(NoGrammarConstraintsCodeAction.class.getName());
+	// FIXME: the element name should be derived from the content model if possible
+	private static final String PLACEHOLDER_ELEMENT_NAME = "root-element";
 	private final Map<String, ICodeActionResolvesParticipant> resolveCodeActionParticipants;
 
 	public NoGrammarConstraintsCodeAction() {
@@ -176,7 +177,7 @@ public class NoGrammarConstraintsCodeAction implements ICodeActionParticipant {
 					GenerateXSINoNamespaceSchemaCodeActionResolver.PARTICIPANT_ID);
 		} else {
 			TextDocumentEdit noNamespaceSchemaLocationEdit = createXSINoNamespaceSchemaLocationEdit(schemaFileName,
-					document);
+					document, request.getSharedSettings());
 			return createGrammarFileAndBindIt(title, schemaURI, schemaTemplate, noNamespaceSchemaLocationEdit,
 					diagnostic);
 		}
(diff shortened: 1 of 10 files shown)


datho7561:
I need to write tests, but the code should be good.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/nogrammarconstraints/NoGrammarConstraintsCodeAction.java
```diff
@@ -246,18 +249,32 @@ public static TextDocumentEdit createXSINoNamespaceSchemaLocationEdit(String sch
...
 		String delimiter = document.getTextDocument().lineDelimiter(0);
 		DOMElement documentElement = document.getDocumentElement();
-		int beforeTagOffset = documentElement.getStartTagOpenOffset();
-		int afterTagOffset = beforeTagOffset + 1 + documentElement.getTagName().length();
+		int beforeTagOffset = documentElement != null ? documentElement.getStartTagOpenOffset() : document.getLastChild() != null ? document.getLastChild().getEnd() : 0;
+		int afterTagOffset = documentElement != null ? beforeTagOffset + 1 + documentElement.getTagName().length() : beforeTagOffset;
 
 		StringBuilder insertText = new StringBuilder();
 
+		if (documentElement == null) {
```
angelozerr, line 257: Move this code in a private static method like `generateStartTag(DOMElement documentElement) `to reuse it with other create*Edit

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/nogrammarconstraints/NoGrammarConstraintsCodeAction.java
```diff
@@ -246,18 +249,32 @@ public static TextDocumentEdit createXSINoNamespaceSchemaLocationEdit(String sch
...
+		}
+
 		insertText.append(" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"");
 		insertText.append(delimiter);
 
 		insertText.append(" xsi:noNamespaceSchemaLocation=\"");
 		insertText.append(schemaFileName);
 		insertText.append("\"");
 
+		if (documentElement == null) {
```
angelozerr, line 272: Move this code in a private static method like `generateEndTag(DOMElement documentElement) `to reuse it with other create*Edit

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/nogrammarconstraints/NoGrammarConstraintsCodeAction.java
```diff
@@ -301,45 +333,51 @@ public static TextDocumentEdit createXmlModelEdit(String schemaFileName, String
...
+			// insert namespace in root element
+			StringBuilder xmlNamespaceInsertText = new StringBuilder();
+			xmlNamespaceInsertText.append(" xmlns=\"");
+			xmlNamespaceInsertText.append(targetNamespace);
+			xmlNamespaceInsertText.append("\" ");
+
+			int afterTagOffset = beforeTagOffset + 1 + documentElement.getTagName().length();
+			Position xmlNamespacePosition = document.positionAt(afterTagOffset);
+			edits.add(CodeActionFactory.insertEdit(xmlNamespaceInsertText.toString(), xmlNamespacePosition));
+		}
```
angelozerr, line 356: is there any reason why you don't generate an element for xml_model? IMHO to be consistent, you should generate an element like other  grammar.

datho7561: I think this is a good idea since then we can declare the proper namespace (if we have the schema locally). Should I also do it for `.dtd`?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/nogrammarconstraints/NoGrammarConstraintsCodeAction.java
angelozerr: I think you should use XMLBuilder every time instead of StringBuilder, You will have the capability to have one method `generateEndTag(DOMElement documentElement, XMLBuilder builder)`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/nogrammarconstraints/NoGrammarConstraintsCodeAction.java
angelozerr: I think we should generate in any case the tag element. In the future we will load the grammar to retrieve the proper tag name.

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/services/extensions/ErrorParticipantLanguageServiceTest.java
```diff
@@ -114,34 +103,18 @@ static class ErrorParticipantLanguageService extends XMLLanguageService {
...
-			this.registerCodeActionParticipant(new ICodeActionParticipant() {
-				@Override
-				public void doCodeAction(ICodeActionRequest request, List<CodeAction> codeActions,
-						CancelChecker cancelChecker) {
-					Diagnostic diagnostic = request.getDiagnostic();
-					codeActions.add(ca(diagnostic, te(0, 0, 0, 0, "a")));
-				}
-			});
+			this.registerCodeActionParticipant((request, codeActions, cancelChecker) -> {
+throw new RuntimeException("This participant is broken");
```
datho7561, line 107: I had the `lambdaExpression` clean up enabled. Looks like it sometimes struggles with getting indentation correct. cc @rgrunber

datho7561: (I fixed it, but essentially sometimes the second 3rd, 4th... lines of a generated lambda aren't indented correctly.

rgrunber: I've noticed this as well. Seems to place regular spaces even when the file is using tabs, in my case.

datho7561: This is upstream, I'll file it.

angelozerr:
@datho7561 could you fix the conflict please.

angelozerr:
It works like a charm, thanks @datho7561 !

PR #1410, opened 2022-12-15
Fix "Surround with ..." when using the binary server

datho7561:


datho7561:
Before anyone merges, I'm going to update the CHANGELOG in this PR as well

PR #1426, opened 2022-12-19
Add code action to generate RelaxNG grammar

### LINKED ISSUE #1405: CodeAction to generate RelaxNG RNG file
When XML si not associate to a grammar we have those code actions to generate XSD or DTD

image

It should be nice to have a CodeAction action which generates an rng file (RelaxNG XML syntax)

JessicaJHee:
Peek 2022-12-19 16-07

Fixes #1405 

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/generators/xml2relaxng/XML2RelaxNGGenerator.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/generators/xml2relaxng/XML2RelaxNGGenerator.java
@@ -1,0 +1,289 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.extensions.generators.xml2relaxng;
+
+import java.util.Collection;
+import java.util.Set;
+
+import org.eclipse.lemminx.dom.DOMAttr;
+import org.eclipse.lemminx.extensions.generators.AbstractXML2GrammarGenerator;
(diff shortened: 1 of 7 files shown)


JessicaJHee:
Currently I have the start of the generated file as: 

```
<element name="rootElementName" xmlns="http://relaxng.org/ns/structure/1.0">
...
```

I see in other examples they have something like :
```
         xmlns:sch="http://purl.oclc.org/dsdl/schematron"
    <start>
        <element name="rootElementName">
        ...
```

I wonder if the second case is better/more accepted?

angelozerr:
It looks promising !

I think you should generate grammar element because

* the most rng file that I have seen uses grammar
* we have a snippet in  rng file which générales grammar. To be consistent with this snippet I think it is better to generate grammar element

In your demo you generate an rng prefix. I think yiu should generate without prefix. Please see the rng snippet

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/generators/xml2relaxng/XML2RelaxNGGenerator.java
```diff
@@ -0,0 +1,269 @@
...
+
+	@Override
+	protected void generate(Grammar grammar, RelaxNGGeneratorSettings settings, XMLBuilder schema,
+			CancelChecker cancelChecker) {
+		String prefix = "rng";
+		// <?xml version="1.0" encoding="UTF-8"?>
+		schema.startPrologOrPI("xml");
+		schema.addContent(" version=\"1.0\" encoding=\"UTF-8\"");
+		schema.endPrologOrPI();
+		// <element name="rootElementName" xmlns:rng="http://relaxng.org/ns/structure/1.0">
```
angelozerr, line 76: Please generate the same content than snippet

* without rng prefix
* with grammar element
* with define pattern

angelozerr:
@JessicaJHee could you write the same test than https://github.com/eclipse/lemminx/tree/main/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/generators/xml2dtd, to be consistent with XSD and DTD tests generator

angelozerr:
The PR looks very good. If CI build works I m OK to merge the PR.

We will improve it and fix bugs on another PR

Issue #1256, opened 2022-07-20
CodeAction for cvc-complex-type.2.4.b: inserts always at start, breaking sequences


GitMensch:
Given a sequence of elements and missing elements at the end the code action inserts those directly at the beginning, leading to "invalid element name" errors.

If the complete complex element is empty then adding the entries via code action inserts those in the correct order, so it is only a thing with "partial" missing elements.

Is it possible to insert the missing elements at the correct place?

_Originally posted by @GitMensch in https://github.com/eclipse/lemminx/issues/1218#issuecomment-1190078320_

angelozerr:
Could you give us a very simple  ml and xsd for this issue. We could use them for tests.

Thanks!

GitMensch:
Not complete but possibly enough:

```xsd
	<xs:element name="cmplx">
		<xs:complexType>
			<xs:sequence>
				<xs:element name="a" minOccurs="1" maxOccurs="1"/>
				<xs:element name="b" minOccurs="0" maxOccurs="1"/>
				<xs:element name="c" minOccurs="0" maxOccurs="1"/>
				<xs:element name="d" minOccurs="1" maxOccurs="1"/>
				<xs:element name="e" minOccurs="1" maxOccurs="1"/>
			</xs:sequence>
		</xs:complexType>
	</xs:element>
```

and

```xsd
	<cmplx>
	  <a></a>
	  <c></c>
	  <d></d>
	</cmplx>
```

-> inserts "e" before "a" (and as long as #1255 is not solved also "b" before that); leading to a message afterwards that "e" / "b" is not expected but "a".

angelozerr:
Thanks @GitMensch for your sample.

PR #1429, opened 2022-12-22
Add code action to add missing required elements for RelaxNG

### LINKED ISSUE #1418: CodeAction to add missing required elements with RelaxNG
Given this RelaxNG schema: https://github.com/eclipse/lemminx/blob/main/org.eclipse.lemminx/src/test/resources/relaxng/tei_all.rng,
and this XML: 

```xml
<?xml-model href="tei_all.rng"?>
<TEI xmlns="http://www.tei-c.org/ns/1.0">
</TEI>
```
We get the following error: 

Screenshot from 2022-12-16 12-30-00

It would be nice to provide a code action that inserts the required elements according to the schema.

JessicaJHee:
Fixes #1418 

### DIFF
--- a/org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGElementDeclaration.java
+++ b/org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGElementDeclaration.java
@@ -14,6 +14,7 @@ package com.thaiopensource.relaxng.pattern;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
@@ -54,7 +55,7 @@ public class CMRelaxNGElementDeclaration implements CMElementDeclaration {
 
 	private Collection<CMAttributeDeclaration> attributes;
 
-	private Set<Name> requiredElementNames;
+	private Set<String> requiredElementNames;
 
 	CMRelaxNGElementDeclaration(CMRelaxNGDocument document, ElementPattern pattern) {
 		this.cmDocument = document;
@@ -232,16 +233,7 @@ public class CMRelaxNGElementDeclaration implements CMElementDeclaration {
 
 	@Override
 	public boolean isOptional(String childElementName) {
-		if (requiredElementNames == null) {
-			RequiredElementsFunction elementsFunction = new RequiredElementsFunction();
-			requiredElementNames = pattern.apply(elementsFunction);
-			for (Name name : requiredElementNames) {
-				if (childElementName.equals(name.getLocalName())) {
-					return false;
-				}
-			}
-		}
-		return true;
+		return !getRequiredElementNames().contains(childElementName);
 	}
 
 	@Override
@@ -272,5 +264,23 @@ public class CMRelaxNGElementDeclaration implements CMElementDeclaration {
 		result.append("\"");
 		result.append(" />");
 		return result.toString();
+	}
+
+	@Override
+	public Set<CMElementDeclaration> getRequiredElements() {
+		Set<CMElementDeclaration> requiredElements = new LinkedHashSet<>();
+		for (String elementName : getRequiredElementNames()) {
+			requiredElements.add(findCMElement(elementName, null));
+		}
+		return requiredElements;
+	}
+
+	private Set<String> getRequiredElementNames() {
+		if (requiredElementNames == null) {
+			requiredElementNames = new LinkedHashSet<>();
+			MyRequiredElementsFunction elementsFunction = new MyRequiredElementsFunction(requiredElementNames);
+			getPattern().getContent().apply(elementsFunction);
+		}
+		return requiredElementNames;
 	}
 }
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/required_element_missingCodeAction.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/required_element_missingCodeAction.java
@@ -32,19 +32,17 @@ import org.eclipse.lsp4j.jsonrpc.CancelChecker;
  * CodeAction to insert expected child elements within an element
  * 
  * Given this XML where the expected child elements are not present as defined
- * in the XSD:
+ * in the relaxNG schema:
  * 
  * <servlet></servlet> Error: Child elements are missing from element: - servlet
- *
- * The following elements are expected: - description - display-name - icon -
- * servlet-name
+ * 
  * 
  * To fix the error, the code action will suggest inserting the expected
  * elements inside the parent tag
  * 
  */
 
-public class cvc_complex_type_2_4_bCodeAction implements ICodeActionParticipant {
+public class required_element_missingCodeAction implements ICodeActionParticipant {
 
 	@Override
 	public void doCodeAction(ICodeActionRequest request, List<CodeAction> codeActions, CancelChecker cancelChecker) {
@@ -66,20 +64,22 @@ public class cvc_complex_type_2_4_bCodeAction implements ICodeActionParticipant 
 			Range targetRange = new Range(childElementPositionStartTag, childElementPositionEndTag);
 			XMLGenerator generator = request.getXMLGenerator();
 
-			String insertStrAll = generator.generateMissingElements(request, element, false);
 			String insertStrRequired = generator.generateMissingElements(request, element, true);
-
-			CodeAction insertAllExpectedElement = CodeActionFactory.replace("Insert all expected elements",
-					targetRange, insertStrAll, document.getTextDocument(), diagnostic);
-
-			codeActions.add(insertAllExpectedElement);
+			String insertStrAll = generator.generateMissingElements(request, element, false);
 
 			CodeAction insertRequriedExpectedElement = CodeActionFactory.replace("Insert only required elements",
 					targetRange, insertStrRequired, document.getTextDocument(), diagnostic);
 
 			codeActions.add(insertRequriedExpectedElement);
 
-		} catch (BadLocationException e) {
+			CodeAction insertAllExpectedElement = CodeActionFactory.replace("Insert all expected elements",
+					targetRange, insertStrAll, document.getTextDocument(), diagnostic);
+
+			codeActions.add(insertAllExpectedElement);
+
+		} catch (
+
+		BadLocationException e) {
 			// do nothing
 		}
 	}
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/XMLGenerator.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/XMLGenerator.java
@@ -17,14 +17,19 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
 
 import org.eclipse.lemminx.commons.SnippetsBuilder;
 import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.dom.DOMNode;
 import org.eclipse.lemminx.extensions.contentmodel.model.CMAttributeDeclaration;
+import org.eclipse.lemminx.extensions.contentmodel.model.CMDocument;
 import org.eclipse.lemminx.extensions.contentmodel.model.CMElementDeclaration;
+import org.eclipse.lemminx.extensions.contentmodel.model.ContentModelManager;
 import org.eclipse.lemminx.services.extensions.ISharedSettingsRequest;
+import org.eclipse.lemminx.services.extensions.codeaction.ICodeActionRequest;
 import org.eclipse.lemminx.settings.SharedSettings;
 import org.eclipse.lemminx.utils.MarkupContentFactory;
 import org.eclipse.lemminx.utils.StringUtils;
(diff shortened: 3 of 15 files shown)


org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/relaxng/xml/codeaction/missingChildElementCodeActionTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.eclipse.lemminx.AbstractCacheBasedTest;
+import org.eclipse.lemminx.extensions.relaxng.xml.validator.RelaxNGErrorCode;
+import org.eclipse.lsp4j.Diagnostic;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test for generating missing child elements for RELAX NG
+ *
+ */
+public class missingChildElementCodeActionTest extends AbstractCacheBasedTest {
```
angelozerr, line 29: Rename class with **M**issingChildElementCodeActionTest

org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/extensions/relaxng/xml/codeaction/missingChildElementCodeActionTest.java
```diff
@@ -0,0 +1,48 @@
...
+		testDiagnosticsFor(xml, d);
+		testCodeActionsFor(xml, d,
+				ca(d, te(2, 8, 2, 8, //
+						"\t\t<name></name>\r\n" + //
+								"\t\t<email></email>")),
+				ca(d, te(2, 8, 2, 8, //
+						"\t\t<name></name>\r\n" + //
+								"\t\t<email></email>")));
+	}
+}
```
angelozerr, line 48: Test should looks like this:
 
```java
@Test
public void incomplete_element_required_element_missing_simple() throws Exception {
	String xml = "<?xml-model href=\"src/test/resources/relaxng/addressBook_v1.rng\" ?>\r\n" + //
			"<addressBook>\r\n" + //
			"	<card>\r\n" + //
			"	</card>\r\n" + //
			"</addressBook>";
	Diagnostic d = d(2, 2, 2, 6, RelaxNGErrorCode.incomplete_element_required_element_missing);
	testDiagnosticsFor(xml, d);
	testCodeActionsFor(xml, d,
			ca(d, te(2, 7, 2, 7, //
					"\r\n" + //
					"\t\t<name></name>\r\n" + //
							"\t\t<email></email>")),
			ca(d, te(2, 7, 2, 7, //
					"\r\n" + //
					"\t\t<name></name>\r\n" + //
							"\t\t<email></email>")));
}
```

In otherwise:

 * you should use tabulation to indent XML (you are using spaces)
 * it misses the `\r\n` before generation of XML element

The test fails because there is a bug with isOptional in CMRelaxNGElementDeclaration.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/XMLGenerator.java
```diff
@@ -149,6 +197,8 @@ private int generate(CMElementDeclaration elementDeclaration, String prefix, boo
 		if (generatedElements.contains(elementDeclaration)
 				|| existingElementNames.contains(elementDeclaration.getLocalName())) {
 			return snippetIndex;
+		}else if (level > 5){
+			return snippetIndex;
```
angelozerr, line 201: Could you give me the full sample which causes problem please to check level?

JessicaJHee: If I use https://github.com/eclipse/lemminx/blob/main/org.eclipse.lemminx/src/test/resources/relaxng/tei_all.rng

angelozerr: Could you please write a test with this usecase

JessicaJHee:
This PR should also fix https://github.com/eclipse/lemminx/issues/1256 since it replaces the entire content of the parent element.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/XMLGenerator.java
JessicaJHee: I have added a test for this case and it seems to be working, but returns a lot to content. I wonder if it's a good idea to restrict the level of children we generate?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/required_element_missingCodeAction.java
```diff
@@ -0,0 +1,86 @@
...
+			Range targetRange = new Range(childElementPositionStartTag, childElementPositionEndTag);
+			XMLGenerator generator = request.getXMLGenerator();
+
+			String insertStrAll = generator.generateMissingElements(request, element, false);
+			String insertStrRequired = generator.generateMissingElements(request, element, true);
+
+			CodeAction insertAllExpectedElement = CodeActionFactory.replace("Insert all expected elements",
+			targetRange, insertStrAll, document.getTextDocument(), diagnostic);
+
+			codeActions.add(insertAllExpectedElement);
```
angelozerr, line 73: Insert `insertRequriedExpectedElement` code action before `insertAllExpectedElement` because the  `insertRequriedExpectedElement`  will be more used than the `insertAllExpectedElement`

org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGElementDeclaration.java
```diff
@@ -233,15 +234,19 @@ public boolean isMixedContent() {
 	@Override
 	public boolean isOptional(String childElementName) {
 		if (requiredElementNames == null) {
-			RequiredElementsFunction elementsFunction = new RequiredElementsFunction();
-			requiredElementNames = pattern.apply(elementsFunction);
-			for (Name name : requiredElementNames) {
-				if (childElementName.equals(name.getLocalName())) {
-					return false;
+			requiredElementNames = new HashSet<>();
```
angelozerr, line 237: After some investigation, we cannot use the standard RequiredElementsFunction  because it returns only the first required element. I suggest that you create thefollowing class:

```java
package com.thaiopensource.relaxng.pattern;

import java.util.Set;

public class MyRequiredElementsFunction extends AbstractPatternFunction<Void> {

	private final Set<String> requiredElementNames;

	public MyRequiredElementsFunction(Set<String> requiredElementNames) {
		this.requiredElementNames = requiredElementNames;
	}

	@Override
	public Void caseOther(Pattern p) {
		return null;
	}

	@Override
	public Void caseElement(ElementPattern p) {
		return caseNamed(p.getNameClass());
	}

	@Override
	public Void caseGroup(GroupPattern p) {
		return union(p);
	}

	@Override
	public Void caseChoice(ChoicePattern p) {
		// For choice, get the first pattern which is not nullable
		Pattern p1 = p.getOperand1();
		if (!p1.isNullable()) {
			p1.apply(this);
		} else {
			Pattern p2 = p.getOperand2();
			if (!p2.isNullable()) {
				p2.apply(this);
			}
		}
		return null;
	}

	@Override
	public Void caseInterleave(InterleavePattern p) {
		return union(p);
	}

	@Override
	public Void caseAfter(AfterPattern p) {
		return p.getOperand1().apply(this);
	}

	@Override
	public Void caseOneOrMore(OneOrMorePattern p) {
		return p.getOperand().apply(this);
	}

	private Void caseNamed(NameClass nc) {
		if (!(nc instanceof SimpleNameClass))
			return null;
		requiredElementNames.add(((SimpleNameClass) nc).getName().getLocalName());
		return null;
	}
	
	private Void union(BinaryPattern p) {
		Pattern p1 = p.getOperand1();
		if (!p1.isNullable()) {
			p1.apply(this);
		}
		Pattern p2 = p.getOperand2();
		if (!p2.isNullable()) {
			p2.apply(this);
		}
		return null;
	}
}

```

and replace isOptionalmethod like this:

```java
@Override
public boolean isOptional(String childElementName) {
	if (requiredElementNames == null) {
		requiredElementNames = new HashSet<>();
		MyRequiredElementsFunction elementsFunction = new MyRequiredElementsFunction(requiredElementNames);
			getPattern().getContent().apply(elementsFunction);
	}
	return !requiredElementNames.contains(childElementName);
}
```

For the usecase of TEI, it should generate only:

```xml
<teiHeader>
	<fileDesc>
		<titleStmt>
			<title>
			</title>
		</titleStmt>
		<publicationStmt>
			<publisher>
			</publisher>
		</publicationStmt>
		<sourceDesc>
			<p>
			</p>
		</sourceDesc>
	</fileDesc>
</teiHeader>
<text>
	<body>
		<div>
		</div>
	</body>
</text>
```

angelozerr:
@JessicaJHee I suggested you a fix for isOptional. I noticed that we generate this XML for TEI:

```xml
<teiHeader>
	<fileDesc>
		<titleStmt>
			<title>
			</title>
		</titleStmt>
		<publicationStmt>
			<publisher>
			</publisher>
		</publicationStmt>
		<sourceDesc>
			<p>
			</p>
		</sourceDesc>
	</fileDesc>
</teiHeader>
<text>
	<body>
		<div>
		</div>
	</body>
</text>
```

I wonder if it should be better to avoid generating line feed when element is empty (ex : for title). Instead of gerenating:

```xml
<titleStmt>
	<title>
	</title>
</titleStmt>
```

I think we should generate:

```xml
<titleStmt>
	<title></title>
</titleStmt>
```

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/XMLGenerator.java
```diff
@@ -162,19 +196,23 @@ private int generate(CMElementDeclaration elementDeclaration, String prefix, boo
 		snippetIndex = generate(attributes, level, snippetIndex, xml, elementDeclaration.getLocalName());
 		// Elements children
 		Collection<CMElementDeclaration> children = elementDeclaration.getElements();
+		boolean childGenerated = false;
```
angelozerr, line 199: I suggest that you provide a new method `getRequiredElements()` to write:

```java
Collection<CMElementDeclaration> children = generateOnlyRequired  ? elementDeclaration.getRequiredElements(); : elementDeclaration.getElements();
```

For XSD and DTD this method could be implemented by looping getElements()  and check if isOptional return false (like today).

For RelaxNG, you compute the requiredElementsName (like it is done today) and you loop for each name and build the list with CMRelaxNGElementDeclaration.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/utils/XMLGenerator.java
```diff
@@ -472,8 +510,15 @@ public static MarkupContent createMarkupContent(CMElementDeclaration cmElement,
 	 * @return true if the element child element is to be generated.
 	 */
 	private boolean isGenerateChild(CMElementDeclaration elementDeclaration, boolean generateOnlyRequired,
-			String childName) {
-		if (!generateOnlyRequired || (!elementDeclaration.isOptional(childName) && generateOnlyRequired)) {
+			CMElementDeclaration child) {
+		// With RelaxNG, no need to generate if the child element is the second element
+		// in <choice>
+		if (elementDeclaration instanceof CMRelaxNGElementDeclaration
```
angelozerr, line 516: Please don't write specific usecase in a generic component. With the idea of getRequiredElements, you should not need this isGenerateChild method.

I think your code covers the bug with choice when all elements are generated. It is because getElements() need to be improved, so let's keep this bug.

datho7561 (approved):

datho7561:
oops I approved the wrong PR

org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGElementDeclaration.java
```diff
@@ -273,4 +271,49 @@ public String toString() {
...
+				return true;
+			}
+		}
+		return false;
+	}
+
+	@Override
+	public Set<CMElementDeclaration> getRequiredElements() {
+		Set<CMElementDeclaration> requiredElements = new LinkedHashSet<>();
+		for (CMElementDeclaration element : getElements()) {
```
angelozerr, line 301: To build this list  I will add à private merhod getRequiredElementNames which will return the computedr requiredElementNames  and I will retrieve the element déclaration by name.

I would like to avoid having the isChoice method.

org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGElementDeclaration.java
```diff
@@ -273,4 +271,49 @@ public String toString() {
...
+			if (!isOptional(element.getLocalName()) && !isSecondChoice((CMRelaxNGElementDeclaration) element)) {
+				requiredElements.add(element);
+			}
+		}
+		return requiredElements;
+	}
+
+	@Override
+	public Collection<CMElementDeclaration> getElementsToGenerate() {
```
angelozerr, line 310: Could you please remove this method. We need to invesitgate more how to manage this usecase with XSD, DTD, and RNG.

The code action which will insert all elements will have some bugs but the most important code action is required elements.

angelozerr:
I test it with TEI and Pretext document and it works perfectly with "Insert only required element". We ned to improve the code action "Insert all elements" to take care of choice (in RelaxNG, but in XSD and DTD too).

Let's see that in an another PR.

Thanks so much @JessicaJHee for your great work!

angelozerr:
@oscarlevin I think you will be interested with this code action, see teh following demo:

InsertRequiredElementPretext

angelozerr:
> Yes, this is fantastic!

Glad you like this feature :)

Please note that it works only if the element can host one element. If there are a choice of element (like pretext) it will not work. See https://github.com/eclipse/lemminx/issues/1448

PR #1431, opened 2023-01-02
Suppor for textDocument/documentColor

### LINKED ISSUE #639: Suppor for `textDocument/documentColor`
Some XML files declare colors like
https://developer.android.com/guide/topics/resources/color-list-resource

It should be nice if we could display color for this kind of XML files with `textDocument/documentColor`

The question is how to say that `android:color` describes a color with a generic mean.

angelozerr:
Suppor for textDocument/documentColor

Fixes #639

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/colors/XMLColorsPlugin.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/colors/XMLColorsPlugin.java
@@ -1,0 +1,98 @@
+/*******************************************************************************
+* Copyright (c) 2023 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.extensions.colors;
+
+import org.eclipse.lemminx.extensions.colors.participants.XMLDocumentColorParticipant;
+import org.eclipse.lemminx.extensions.colors.settings.XMLColorsSettings;
+import org.eclipse.lemminx.services.extensions.IDocumentColorParticipant;
+import org.eclipse.lemminx.services.extensions.IXMLExtension;
+import org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry;
+import org.eclipse.lemminx.services.extensions.save.ISaveContext;
+import org.eclipse.lsp4j.InitializeParams;
+
+/**
+ * XML colors plugin.
+ * 
+ * This plugin provides the capability to support color with 'xml.colors'
+ * settings.
+ * 
+ * <code>
+ "xml.colors": [
+   // XML colors applied for text node for android colors.xml files
+   {
+      "pattern": "/res/values/colors.xml",
+      "expressions": [
+         {
+            "path": "resources/color/text()"
+         }
+      ]
+   },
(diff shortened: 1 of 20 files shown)


angelozerr:
To play with this PR with vscode-xml, please use the PR https://github.com/redhat-developer/vscode-xml/pull/849

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/colors/XMLColorsPlugin.java
```diff
@@ -0,0 +1,103 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
```
fbricon, line 2: 2023 all the way now

angelozerr:
To test this PR, write this XML color file (android)

```xml
<resources>
	<color name="opaque_red">#f00</color>
	<color name="translucent_red">#80ff0000</color>
</resources>
```
Write this settings:

```json
"xml.colors": [
      {
         "pattern": "**/res/values/colors.xml",
         "expressions": [
            {
               "path": "resources/color/text()"
            }
         ]
      }
   ]
```
You should see color:

image

angelozerr:
See https://github.com/redhat-developer/vscode-xml/pull/853 for declaring `xml.colors` setting sin vscode-xml side.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/colors/XMLColorsPlugin.java
angelozerr: Good catch, fixed

fbricon:
please fix the commit message typo

angelozerr:
> please fix the commit message typo

fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/colors/utils/ColorUtils.java
```diff
@@ -0,0 +1,374 @@
...
+		label.append(toTwoDigitHex(green256));
+		label.append(toTwoDigitHex(blue256));
+		if (alpha != null) {
+			label.append(toTwoDigitHex(Math.round(alpha * 255)));
+		}
+		return label.toString();
+	}
+
+	private static String toTwoDigitHex(int n) {
+		String r = String.valueOf(n);
```
JessicaJHee, line 318: I think the issue with converting between rgb and here comes from here because `valueOf()`  will convert the hex to integer (ex: `0xff` will become `255`, and in the next like will become `0255`), which doesn't work.

If you change this line to `String r = Integer.toHexString(n);` (`toHexString()` will convert `0xff` to `ff`), it should work like this: 

Peek 2023-01-05 11-11

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLDocumentColor.java
```diff
@@ -0,0 +1,84 @@
+/**
+ *  Copyright (c) 2018 Angelo ZERR
```
JessicaJHee, line 2: Please update header

JessicaJHee (changes requested):
Other than the comment I left which hopefully will improve conversion between hex/hexa and rgb/rgba color, it works well and the code looks good!

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLDocumentColor.java
angelozerr: fixed.

JessicaJHee (approved):

PR #1433, opened 2023-01-03
Fix bug in comment formatting with xml.format.maxLineWidth

JessicaJHee:
Fixes https://github.com/redhat-developer/vscode-xml/issues/851

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMCommentFormatter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMCommentFormatter.java
@@ -54,12 +54,18 @@ public class DOMCommentFormatter {
 		}
 		int spaceStart = -1;
 		int spaceEnd = -1;
-		availableLineWidth -= 4; // count for '<-!--'
+		availableLineWidth -= 4; // count for '<!--'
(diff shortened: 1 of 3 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMCommentFormatter.java
```diff
@@ -54,12 +54,22 @@ public void formatComment(DOMComment commentNode, XMLFormattingConstraints paren
...
+		availableLineWidth -= 4; // count for '<!--'
+		int lineSeparatorOffset = -1;
 
 		for (int i = commentNode.getStartContent(); i < commentNode.getEndContent(); i++) {
 			char c = text.charAt(i);
 			if (Character.isWhitespace(c)) {
-				// Whitespaces
+				if (isLineSeparator(c)) {
+					// spaceStart offset not including line separator characters
+					lineSeparatorOffset = i;
```
datho7561, line 65: you should be able to set this variable outside of the if/else statement, then simplify the if/else statement

datho7561 (changes requested):
Works pretty well, I just found somewhere that the code could be simplified

datho7561 (approved):
Great work! Thanks, Jessica!

PR #1438, opened 2023-01-05
Fix `textDocument/colorPresentation` in binary

### LINKED ISSUE #1436: Exception while using `textDocument/documentColor` with binary
When I try to use the new colors feature with the binary version of lemminx in VS Code, I get:

`Message: Unable to create instance of class org.eclipse.lsp4j.ColorPresentationParams. Registering an InstanceCreator or a TypeAdapter for this type, or adding a no-args constructor may fix this problem.`

I think that all that's needed to fix it is to update `reflect-config.json` in order to register the no-args constructor for reflection.

datho7561:
Fixes #1436


angelozerr:
Thanks so much @datho7561 !

PR #1442, opened 2023-01-09
Fix max line width wrapping with attributes

### LINKED ISSUE #1439: Wrap long lines
The behavior described [here](https://bugs.eclipse.org/bugs/show_bug.cgi?id=501252) seems to happen to POM formatting, namely, the first line is not wrapped. Instead, it displays an overly long line, and even merges the lines if they are manually wrapped.

I believe it should wrap the line at some reasonable length, by default (and I suppose this could be configurable with some line length parameter).

JessicaJHee:
Fixes #1439 


datho7561 (approved):
Looks good. Thanks Jessica!

PR #1449, opened 2023-01-13
SAXParseException in the language server trace when editing an .rng file

### LINKED ISSUE #1441: `SAXParseException` in the language server trace when editing an `.rng` file
If you edit a RelaxNG file that uses the XML syntax, and the document becomes invalid, eg:

```xml
  <start>
    <ref name="root-element"/>
  </start>
  <define name="root-element">
      <element name="root-element">
        <ref name="asdf"></ref>
      </element>
  </define>
  <div>
    <div>
      <define name="asdf">
        <attribute name="child"></attribute>
        <element>
      </define>
    </div>
  </div>
</grammar>
```

(note the unclosed `<element>`)

Then a `SAXParseException` is output in the language server trace. It would be nice if this Exception didn't appear, since having an invalid XML file is expected when you are editing it.

angelozerr:
SAXParseException in the language server trace when editing an .rng file

Fixes #1441


datho7561 (approved):
Looks good. Thanks, Angelo!

PR #1451, opened 2023-01-13
XML references linked editing range should not work with DOM document

angelozerr:
XML references linked editing range should not work with DOM document


angelozerr:
Create an XML file like:

```xml
<pretext>
	<article>
	</article>
</pretext>
```

which is linked to some references.

If you set the cursor at the begin file, linked editing range is performed:

image

This PR fix this issue.

datho7561 (approved):
Looks good! No idea how I didn't catch this in my review, my bad.

PR #1452, opened 2023-01-13
Support references in included files for codelens + diagnostics.

angelozerr:
Search references inincluded files for codelens + diagnostics.

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/references/participants/XMLReferencesCodeLensParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/references/participants/XMLReferencesCodeLensParticipant.java
@@ -48,7 +48,8 @@ public class XMLReferencesCodeLensParticipant implements ICodeLensParticipant {
 	@Override
 	public void doCodeLens(ICodeLensRequest request, List<CodeLens> lenses, CancelChecker cancelChecker) {
 		DOMDocument document = request.getDocument();
-		Collection<ReferenceLink> links = SearchEngine.getInstance().searchLinks(document, plugin.getReferencesSettings(),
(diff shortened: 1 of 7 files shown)


angelozerr:
Given this XML file:

```xml
<docbook>
	<xref linkend="s1" />
	<section id="s1" />
	<xref linkend="ch1" />
	<xi:include href="sub-book.xml" />
</docbook>
```

with `sub-book.xml` file:

```xml
<book>
	<chapter id="ch1" />
	<chapter id="ch2"/>
	<xref linkend="s1" />
	<xref linkend="s1" />
	<xref linkend="ch1" />
	<xref linkend="ch1" />
</book>
```

and an expression like this:

```json
{
  "pattern": "**/*.xml",
  "expressions": [
    {
      "from": "xref/@linkend",
      "to": "@id"
    }
  ]
}
```

 * the included xref  in sub-book.xml is not used to count codelens: `<section id="s1" />` should show [3 references]
 * the diagnostic report 	`ch1` as an error although it is defined in the sub-book.xml

angelozerr:
> I think that it would be nice to disable this when "xml.validation.xInclude.enabled": false,

I would like to avoid doing that, because my idea is to remove all XSD, RNG support and delegate references with XML references. For XSD, xsd:include is not linked to xiinclude.

For the moment the include is hard coded, but my idea is to give the capability to customize it when you write an xm reference expression like:

{
  "pattern": "**/*.xml",
  "include": [
      "include/@href",
      "import/@href"
   ],
  "expressions": [
    {
      "from": "xref/@linkend",
      "to": "@id"
    }
  ]
}

angelozerr:
> I get a FileNotFoundException when the referenced xi:include/@href doesn't exist

I removed the log

datho7561 (approved):
Looks good. Thanks, Angelo!

PR #1453, opened 2023-01-13
Ignore linked editing range when there are no referenced node to update

angelozerr:
Ignore linked editing range when there are no referenced node to update


angelozerr:
Given this XML file:

```xml
<docbook>
	<xref linkend="s2" />
	<section id="s1" />
</docbook>
```
with an expression like this:

```json
{
  "pattern": "**/*.xml",
  "expressions": [
    {
      "from": "xref/@linkend",
      "to": "@id"
    }
  ]
}
```

If `s1` is selected, linked editing range not to be activated.

This PR fixes this issue.

datho7561 (approved):
This fixes the issue you mentioned. Thanks, Angelo!

PR #1462, opened 2023-01-24
Generate end tag with xml.completion.autoCloseTags set to false

### LINKED ISSUE #1458: "Insert element" CodeActions don't add close tags when auto close tag is disabled
If you set `"xml.completion.autoCloseTags": false`, then any code action that generates missing elements doesn't generate the end tag.

This happens for XML files linked to XSD and RNG. (files linked to DTD don't have this type of code action as of the time of writing).

JessicaJHee:
Fixes #1458 


datho7561 (approved):
Looks good and works well! Thanks, Jessica!

PR #1482, opened 2023-02-16
Log file handler at level from "log.level"

scottkurz:
Since all we can do with java.util.logging is to use this particular FileHandler, it doesn't do any good to set the `log.level` property to a specific value but then to hard-code this FileHandler to `setLevel(Level.INFO)`.   We might as well set the level set via the `log.level` sysprop.

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/logs/LogHelper.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/logs/LogHelper.java
@@ -59,7 +59,7 @@ public class LogHelper {
 			if (!StringUtils.isBlank(path)) {
 				createDirectoryPath(path);
 				try {
-					FileHandler fh = LogHelper.getFileHandler(path);
+					FileHandler fh = getFileHandler(path);
 					logger.addHandler(fh);
 				} catch (SecurityException | IOException e) {
 					logger.warning("Error at creation of FileHandler for logging");
@@ -123,7 +123,7 @@ public class LogHelper {
 			FileHandler fh = null;
 			fh = new FileHandler(filePath, true);
 			fh.setFormatter(new SimpleFormatter());
-			fh.setLevel(Level.INFO);
+			fh.setLevel(getLogLevel());
 			return fh;
 		}
 		throw new IOException("Cannot write file since it cannot be written to");


scottkurz:
@fbricon  - Small enhancement continuing the discussion in:  https://github.com/eclipse/lemminx-maven/issues/300

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/logs/LogHelper.java
```diff
@@ -123,7 +123,7 @@ public static FileHandler getFileHandler(String filePath) throws SecurityExcepti
 			FileHandler fh = null;
 			fh = new FileHandler(filePath, true);
 			fh.setFormatter(new SimpleFormatter());
-			fh.setLevel(Level.INFO);
+			fh.setLevel(getLogLevel());
```
angelozerr, line 126: I didn't know very good FileHandler, but I wonder if we just remove `fh.setLevel(Level.INFO);` it will take the level of root?

scottkurz: Good point.   I can try to confirm a bit later.

angelozerr: @scottkurz had you time to investigate this idea?

scottkurz: No, I haven't yet, sorry.

datho7561: @scottkurz we are planning on doing a 0.25.0 for some bug fixes and API changes. Do you want to get this PR in for the release?

scottkurz: Thanks for asking.  I can't ask you to wait since not sure when I'll get back to this.  Appreciate it though.

datho7561: Okay, sound good!

PR #1486, opened 2023-02-17
Fix bug with selection formatting outside of range

### LINKED ISSUE #1485: Partial formating yields wrong intendation deep
If I format only a part of the document one gets wrong indentation, e.g.:

## Current
image

## Expected
image

JessicaJHee:
Fixes #1485 

Peek 2023-02-17 12-43

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMElementFormatter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/format/DOMElementFormatter.java
@@ -68,7 +68,8 @@ public class DOMElementFormatter {
 			formatChildren(element, constraints, start, end, edits);
 
 			// Format end tag element with proper indentation
-			if (element.hasEndTag()) {
+			if (element.hasEndTag() && (element.getEndTagOpenOffset() > start
+					&& (end == -1 || element.getEndTagCloseOffset() < end))) {
 				width = formatEndTagElement(element, parentConstraints, constraints, edits);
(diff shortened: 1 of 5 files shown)


datho7561 (approved):
Works very well! Thanks, Jessica!

datho7561:
This PR addresses the issue Fred pointed out in his comment https://github.com/eclipse/lemminx/issues/1485#issuecomment-1434591999, but I think the issue 
Christoph pointed out might be related to the limitations of lsp4e with regards to managing spaces vs. tabs (it can't be managed on a per-file type basis from what I understand).

laeubi:
@datho7561 I always wondered if not simply the current file style could be used when doing partial formats? I can't think about a use case where a mixed style is desirable (except the file is already mixed type).

angelozerr:
For formatting it will work, but if you pres tabs (and your configuration is tab), it will generate a tab although you would like to have 2 spaces.

laeubi:
I must confess I never really "configured" this :-)
Just from an users POV, the editor should be smart enough by default to simply translate this, let it be partial (or even full!) format or pressing a key that means "indent it some bit". But if it is only partially supported, even that could help, because at least I mostly use auto format rather than start intended the code/xml myself...

angelozerr:
Before the pom.xml used WTP XML Editor. You can open it with:

image

And I think you should have this settings:

image

Today, the Maven POM Editor uses Generic Editor which uses for XML, lemminx, but badly GenericEditor cannot be configured with custom XML tabs / spaces for a given language.

PR #1491, opened 2023-02-22
RNG attribute completion doesn't generate the proper prefix if the namespace is not declared

### LINKED ISSUE #1489: RNG attribute completion doesn't generate the proper prefix if the namespace is not declared
Given this XML file bound to RelaxNG file:

```xml
<?xml-model href="http://docbook.org/xml/5.0/rng/docbook.rng" schematypens="http://relaxng.org/ns/structure/1.0"?>
<book xmlns="http://docbook.org/ns/docbook">
    <acknowledgements></acknowledgements>
</book>
```

When completion is triggered on `<acknowledgements |>` it show attributes, but it doesn't take care of prefix, if the namespace is not declared correctly. For instance it shows `actuate` although `actuate` should be linked to `http://www.w3.org/1999/xlink` namespace by using `xlink` prefix.

image

Please note that If you declare the xlink namespace

`<book xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" >`

The completion shows correctly completion item with `xlink:actuate`.

image

To fix this issue, it should be nice to get xlink prefix from the grammar, and when completion is applied, it should insert in the book element the proper namespace `xmlns:xlink="http://www.w3.org/1999/xlink"` like the follwoing demo:

AttrWithPrefixCompletion2

angelozerr:
RNG attribute completion doesn't generate the proper prefix if the namespace is not declared

Fixes #1489

### DIFF
--- a/org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGAttributeDeclaration.java
+++ b/org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGAttributeDeclaration.java
@@ -14,6 +14,7 @@ package com.thaiopensource.relaxng.pattern;
 import java.util.Collection;
 import java.util.List;
 
+import org.eclipse.lemminx.dom.DOMElement;
 import org.eclipse.lemminx.extensions.contentmodel.model.CMAttributeDeclaration;
 import org.eclipse.lemminx.extensions.contentmodel.model.CMElementDeclaration;
 import org.eclipse.lemminx.services.extensions.ISharedSettingsRequest;
@@ -38,12 +39,35 @@ public class CMRelaxNGAttributeDeclaration implements CMAttributeDeclaration {
 
 	private final CMRelaxNGElementDeclaration cmElement;
 	private final AttributePattern pattern;
+
+	private String prefix;
+
+	private boolean computedPrefix;
+
 	private boolean required;
 	private List<String> values;
 
 	public CMRelaxNGAttributeDeclaration(CMRelaxNGElementDeclaration element, AttributePattern pattern) {
 		this.cmElement = element;
 		this.pattern = pattern;
+		this.computedPrefix = false;
+	}
+
+	@Override
+	public String getPrefix() {
+		String namespaceURI = getNamespace();
+		if (namespaceURI == null) {
+			return null;
+		}
+		if (computedPrefix) {
+			return prefix;
+		}
+		DOMElement element = ((CMRelaxNGElementDeclaration) getOwnerElementDeclaration()).getDOMElement();
+		if (element != null) {
+			prefix = element.getPrefix(namespaceURI);
+		}
+		computedPrefix = true;
+		return prefix;
 	}
 
 	@Override
@@ -60,7 +84,7 @@ public class CMRelaxNGAttributeDeclaration implements CMAttributeDeclaration {
 	public CMElementDeclaration getOwnerElementDeclaration() {
 		return cmElement;
 	}
-	
+
 	Name getJingName() {
 		NameClass nameClass = pattern.getNameClass();
 		if (nameClass instanceof SimpleNameClass) {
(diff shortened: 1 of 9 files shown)


angelozerr:
It starts working but I need to cover corner usecases, support resolve completion and write tests. Here a demo:

AttrWithPrefixCompletion2

org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGAttributeDeclaration.java
```diff
@@ -38,12 +42,48 @@ public class CMRelaxNGAttributeDeclaration implements CMAttributeDeclaration {
...
+			DOMDocument document = ((CMRelaxNGElementDeclaration) getOwnerElementDeclaration()).getCMDocument()
+					.getDocument(documentURI);
+			try {
+				int offset = document.offsetAt(new Position(locator.getLineNumber(), locator.getColumnNumber()));
+				DOMAttr attr = document.findAttrAt(offset);
+				if (attr != null) {
+					prefix = attr.getOwnerElement().getPrefix(namespaceURI);
+				} else {
+					prefix = document.getDocumentElement().getPrefix(namespaceURI);
+				}
```
fbricon, line 80: can any of those chained method calls (getPattern(), getOwnerElementDeclaration() ...) return something null?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCompletionParticipant.java
```diff
@@ -302,30 +304,53 @@ public void onAttributeName(boolean generateValue, ICompletionRequest request, I
...
+					Range range = XMLPositionUtility.createRange(offset, offset, document);
+					StringBuilder insertText = new StringBuilder();
+					insertText.append("xmlns:");
+					insertText.append(prefix);
+					insertText.append("=\"");
+					insertText.append(namespaceURI);
+					insertText.append("\"");
+					TextEdit edit = new TextEdit(range, insertText.toString());
+					item.setAdditionalTextEdits(new ArrayList<>());
+					item.getAdditionalTextEdits().add(edit);
```
fbricon, line 343: item.setAdditionalTextEdits(Collections.singletonList(edit)); is more efficient

org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGAttributeDeclaration.java
angelozerr: I clean my code. It should be better.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCompletionParticipant.java
angelozerr: fixed

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCompletionParticipant.java
```diff
@@ -302,30 +303,68 @@ public void onAttributeName(boolean generateValue, ICompletionRequest request, I
...
+					insertText.append(' ');
+					insertText.append("xmlns:");
+					insertText.append(prefix);
+					insertText.append("=\"");
+					insertText.append(namespaceURI);
+					insertText.append("\"");
+					char currentChar = document.getText().charAt(offset);
+					char previousChar = document.getText().charAt(offset - 1);
+					if (currentChar == '>' && previousChar == ' ') {
+						insertText.append(' ');
```
JessicaJHee, line 355: I think in this case, we don't want to insert the space inserted on line 346 since it ends up being 2 spaces: 
Peek 2023-02-22 11-20

JessicaJHee (commented):
Overall works great! Just had one small comment about the namespace insert text

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelCompletionParticipant.java
angelozerr: Good catch. It should be fixed now (with tests too).

datho7561 (approved):
Code looks good and appears to be working. I manually tested using the binary version, so there are no bugs related to that. Thanks, Angelo!

PR #1496, opened 2023-03-01
Comments throws off diagnostic (cvc-complex-type.2.3) mark onto the wrong location

### LINKED ISSUE #1495: Comments throws off diagnostic (cvc-complex-type.2.3) mark onto the wrong location
Adding text to an XML element and the diagnostic "Element 'server' cannot contain text content.
The content type is defined as element-only." is expected on the text. In the below screenshot, inserting a comment in between the text and the element, and the diagnostic range switches to the comment even though it is still expected on the text. 

image

See https://github.com/OpenLiberty/liberty-language-server/issues/163 for more details. Observed on `server.xml` and `pom.xml` files.

angelozerr:
Comments throws off diagnostic (cvc-complex-type.2.3) mark onto the wrong location

Fixes #1495


angelozerr:
This PR should fix #1495:

The `test` content from the pom.xml is now reported as error instead of the comment:

image

datho7561 (approved):
Looks good to me. Thanks, Angelo!

PR #1502, opened 2023-03-08
Provide cancel support for hover

### LINKED ISSUE #1474: Provide cancel support for hover
It should be nice to use the cancel checker in our hover.

angelozerr:
Provide cancel support for hover

Fixes #1474


datho7561 (approved):
This appears to be working, I didn't run into any new bugs with hover when running this PR. Just to leave a note about this: hover requests are never cancelled in lemminx itself, but an extension onto lemminx, like lemminx-maven, might benefit from cancellation support. Thanks, Angelo!

PR #1503, opened 2023-03-09
Adjust error range for RelaxNG element-not-allow-yet error

### LINKED ISSUE #1459: Adjust error range for RelaxNG "not allowed yet" when there is `choice`
Given the following `.rng`:

```xml
  <start>
    <ref name="rootElt" />
  </start>
  <define name="rootElt">
    <element name="root">
      <choice>
        <element name="elementA">
          <text />
        </element>
        <element name="elementB">
          <text />
        </element>
      </choice>
      <element name="element2">
        <text />
      </element>
    </element>
  </define>
</grammar>
```

and the following XML:

```xml
<?xml-model href="notAllowedYet.rng"?>
<root>
  <element2></element2>
</root>
```

The error code is listed as `to_implement`, and the error range doesn't select the start tag of `element2`.

JessicaJHee:
Fixes #1459 

Screenshot from 2023-03-09 13-49-37


datho7561 (approved):
Looks good. Thanks Jessica!

PR #1504, opened 2023-03-10
Do not record number of file.identifier, file.extension instances.

rgrunber:
- Add support for RelaxNG grammar detection in telemetry manager

Basic idea here is instead of recording properties of the form `("file.extension.xml", "10")` for each file extension we simply record `("file.extension", [ xml, xsd, rng ])`. Metrics regarding amount will be lost, but the format is more elegant and should be easier to process. Also I think the resolver kind could be used to get a rough idea of the raw number of different formats, which we do still record.

CC @datho7561


datho7561:
I did the usual change where I set the delay and frequency of telemetry events to 1 minute, but testing with vscode-xml I don't see any file event telemetry events being logged.

rgrunber:
Could there be some file that isn't being handled/fails, which causes the telemetry to not be sent ? If you insert a breakpoint in the part that schedules sending on fixed delay, it might indicate what the state of the telemetry cache is before sending. Is it reproducible on any given file ?

datho7561:
My bad, I forgot to also comment out the code that sends only on even hours of the day.

datho7561 (approved):
Looks good! Thanks, Roland.

Issue #1516, opened 2023-05-03
Allow code actions to be created if no Diagnostics is provided


vrubezhny:
Currently `org.eclipse.lemminx.services.XMLCodeActions.doCodeActions(CodeActionContext, Range, DOMDocument, SharedSettings, CancelChecker)` requires a Diagnostics to be explicitly  set in CodeActionContext in order to acquire code actions from `ICodeActionParticipant'` -s defined.

This doesn't allow the creation of `CodeAction` -s that do not require any error/warning/info/etc. diagnostic for a document range, like, for example, for refactoring or code snippets.

In order to allow the creation of code actions for any document position (even if there is no any diagnostic is created for such a position) we need to make diagnostics to be optional in `XMLCodeActions.doCodeActions(...)`

PR #1517, opened 2023-05-08
Allow code actions to be created if no Diagnostics is provided #1516

vrubezhny:


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCodeActions.java
```diff
@@ -39,37 +42,56 @@ public class XMLCodeActions {
...
+					cancelChecker.checkCanceled();
 					try {
-						cancelChecker.checkCanceled();
 						CodeActionRequest request = new CodeActionRequest(diagnostic, range, document,
 								extensionsRegistry, sharedSettings);
 						codeActionParticipant.doCodeAction(request, codeActions, cancelChecker);
 					} catch (CancellationException e) {
-						throw e;
+						LOGGER.log(Level.FINER, e.getMessage(), e);
+						return Collections.emptyList();
```
angelozerr, line 81: Why you dont rethrrow the cancel exception? All another participant like hover completion do that.

I did that to notify to the client that opération has been canceled

angelozerr:
Perhaps Im wrong but my first feeling with noError diagnostic is that the API is strange.

I wonder if we should update the API of ICodeActionPartiipant to introduce a new method like doCodeActionswithoutDiagnostic or à better name.

It would be nice too to have a test which implement a code action without diagnostic by registering a mock participant.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCodeActions.java
vrubezhny: @angelozerr Thanks for asking this. I also wanted to discuss this...

My short answer is: "because:
1) Participants API methods no not declare in their signatures that any exception like `CancellationException` can be thrown. 
2) We don't need to report anything when cancelled (imho - see my thoughts below...)"

In LemMinX-Maven hovers and definitions I don't re-throw any `CanceletionExceptions` -s - any "cancel" makes it to return a 'null" or empty result as soon as possible. I haven't looked into completion or other participants yet.

I don't think the cancellation is to be reported back to the client (especially in case if it was initiated by a client).

IMHO, if a client sends a cancellation request to the server - the server should close its current operation as soon as possible and not initiate sending of any extra statuses back to the client (and client, once it cancelled the current operation, is not expecting any valuable operation result nor any "prove" of that operation is really cancelled). If a server needs to cancel an ongoing operation - it's the sane - return "null" or empty result to a  client as soon as possible and that's enough.

Still not sure if it's really related, but we were already fixing the issue of such "cancellation reporting" - https://github.com/eclipse/lsp4e/pull/603

Also, leaving the `throw (CancellationException)e` here makes us to re-develop the tests so they correctly catch and process these exceptions - it's not the current case right now, so I stopped creating a JUnit test and make changes for  Cancellation Support in LemMinX-Maven diagnostic participant for the same reason: `o.e.lemminx.services.XMLDiagnostics.doDiagnostics(DOMDocument, XMLValidationSettings, Map<String, Object>, CancelChecker)`  throws `CancellationException` whithout any declaration of that this exception can be thrown. 

If I'm wrong and we still really need to throw and re-throw the `CancellationException` from Diagnostic/Definition/Hover/Content assist/Code action/etc participants, then I insist in adding  `throws CancellationException`   to their public API method signatures so everybody knows that this exception need to be handled.

angelozerr:
@vrubezhny if you see the doc of LSP4J you can see that the call of CancellableException is not catched https://github.com/eclipse-lsp4j/lsp4j/blob/79df5e2ac119e0238a95567bb4bc0f1741428343/documentation/jsonrpc.md?plain=1#L67

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCodeActions.java
angelozerr: @vrubezhny if you see the doc of LSP4J you can see that the call of cancelChecker which can throw a CancellableException is not catched https://github.com/eclipse-lsp4j/lsp4j/blob/79df5e2ac119e0238a95567bb4bc0f1741428343/documentation/jsonrpc.md?plain=1#L67

angelozerr: > If I'm wrong and we still really need to throw and re-throw the CancellationException from Diagnostic/Definition/Hover/Content assist/Code action/etc participants, then I insist in adding throws CancellationException to their public API method signatures so everybody knows that this exception need to be handled.

I think it is better to throw a CancellationException instead returning an empty list, but I suggest that you create an issue to LSP4J to clarify that. Once it is clarified, we will need to update APIs (define CancellationException  in the API as suggested or return an empty value for other operations).

vrubezhny:
> Perhaps Im wrong but my first feeling with noError diagnostic is that the API is strange.
> 
> I wonder if we should update the API of ICodeActionPartiipant to introduce a new method like doCodeActionswithoutDiagnostic or à better name.
> 

I also have such feeling...

When I add this  fake "NoError" diagnostic it makes obligatory to CodeAction participants to know and check for that special diagnostics code in order to work correctly and not to produce any CodeAction duplication.

I thought in adding this diagnostics ONLY in case of empty diagnostics list, but this for sure makes CodeAction participants that do not require any dedicated error code to produce duplicated CodeAction-s in some cases without any possibility to control it.

Having an API method like `doCodeActionswithoutDiagnostic` would be a better solution.

mickaelistria:
I agree the NO_ERROR is not so good, but I don't think a new API would be necessary nor desirable as it would break compatibility. The ICodeActionRequest does include the diagnostic, what about always just sending another request with this diagnostic being null, after sending requests that reference particular diagnostics?

angelozerr:
> Code Action participants are obliged to check whether the correct diagnostic is provided for diagnostic-dependent

This kind of think that I would like to avoid. I like the idea with adding à New method because when you implement it you are sûre that diagnostic is null and the  code is clean. When you implement doCodeAction you work with diagnostic, when you implement doCodeActionsIncondicional you work without diagnostic.

We can add this method in participant and implement as default

mickaelistria:
> This kind of think that I would like to avoid.

Why? It's really simple and efficient, and very close to LSP specification; so no-one would be surprised with such pattern.

angelozerr:
> Why? It's really simple and efficient, and very close to LSP specification; so no-one would be surprised with such pattern.

Because it requires for each participant to check diagnostic is not null. I'm surprised that there are not an NPE at

https://github.com/eclipse/lemminx/blob/7f668204d03ea82235efc664d13f462a38bf7fa3/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseTagCodeAction.java#L41 (perhaps it is logged)

More it means that you will call twice doCodeActions. IMHO I think it is better to separate the code actions:

 * code action for a given diagnostic in a doCodeActions method
 * code action for none diagnostic in a doCodeActionsIncondicional method

The dispatch of those code actions are cleaned. If youhave just one doCodeActions, it will require to check if the diagnostic is null.

More if we have a new method `doCodeActionsIncondicional `, we can find with the IDE all code actions which implement this method. It is an easy and fats mean to retrieve all refactoring code actions for instance.

mickaelistria:
> Because it requires for each participant to check diagnostic is not null.

I suspect many already do. Do you have a list of extensions of lemminx? If it requires to just tweak a few ones to add a null-check in order to be more consistent with LSP and avoid new APIs, it can be worth it.

> More it means that you will call twice doCodeActions.

We already call it as many times as there are diagnostics on that position. This can be multiple times.

angelozerr:
>  Do you have a list of extensions of lemminx? 

See https://github.com/eclipse/lemminx/blob/main/docs/LemMinX-Extensions.md#lemminx-extensions

> in order to be more consistent with LSP and avoid new APIs, it can be worth it.

Adding just a new method `doCodeActionsIncondicional ` in ICodeActionParticipant implemented by default will not break the current API.

I like the idea to have `doCodeActionsIncondicional`  because it provides some convention of the lemminx extension:

*  if you wish to manage code action for diagnostic, let's implement doCodeActions
* if you wish to manage code action like refactoring which doesn't need diagnostic, let's implement `doCodeActionsIncondicional` 

It means that when you contribute to a lemminx extsenion of a given project that you don'tknow, you need just to seach implementation of `doCodeActionsIncondicional` to implement new code action refactoring. You use your IDE search to retrieve all existing code action of refactoring.

vrubezhny:
> 
> Because it requires for each participant to check diagnostic is not null. I'm surprised that there are not an NPE at
> 
> https://github.com/eclipse/lemminx/blob/7f668204d03ea82235efc664d13f462a38bf7fa3/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/CloseTagCodeAction.java#L41
> 
> (perhaps it is logged)

Haven't seen any NPEs loged regarding this Code Action.
This code action participant isn't invoked directly from `XMLCodeActions`, it's invoked from `ContentModelCodeActionParticipant.doCodeActions()` which  performs all the required verification of `request.getDiagnostics()` result, so if `request.getDiagnostics() == null` this participant will never be used.

vrubezhny:
@mickaelistria @angelozerr Would you mind if I merge this PR - I need it to continue working on  https://github.com/eclipse/lemminx-maven/issues/383 to provide Inline code actions as well as to continue on  other parts of the issue.

mickaelistria:
I defer the final say to @angelozerr

angelozerr:
Me I would prefer having `doCodeActionsIncondicional ` to provide some rules to implement code actions without diagnostic and find easily in a project those code actions kind.

After that if @mickaelistria and @vrubezhny you are not happy with this API, let's give up my suggestion.

vrubezhny:
The PR is changed so now the `org.eclipse.lemminx.services.extensions.codeaction.ICodeActionParticipant.doCodeActionUnconditional(ICodeActionRequest, List<CodeAction>, CancelChecker)` (default) method is introduced that is to be invoked for the cases where no any Dignostic is required when creating Code Actions . JUnit test is modified accordingly.

vrubezhny:
Actually I would suggest:
- Make `ICodeActionParticipant.doCodeAction` default as well  in order to make it not necessary to add its default implementation in case it's not needed
- Bump version +0.1 and add `@since 0.26` annotation to   ICodeActionParticipant.doCodeActionUnconditional` method,

PR #1523, opened 2023-06-08
xml.foldings.includeClosingTagInFold has no effect in binary mode

angelozerr:
xml.foldings.includeClosingTagInFold has no effect in binary mode

Fixes https://github.com/redhat-developer/vscode-xml/issues/906


fbricon:
@angelozerr works great thanks!

PR #1558, opened 2023-07-22
XInclude cannot be activated with binary mode

angelozerr:
Fixes https://github.com/redhat-developer/vscode-xml/issues/922


datho7561 (approved):
This fixes the use case described in the pretext issue. Thanks, Angelo!

PR #1567, opened 2023-07-31
Implement itemDefaults for completionList

### LINKED ISSUE #1561: Implement itemDefaults for CompletionList
In lemminx maven, there are completions that returns a lot of items in the result. To improve the performance, implement itemDefaults at the end of completion here: 

https://github.com/eclipse/lemminx/blob/2b0fe29f52c111be3d5c1dccdf9d08340c63be31/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java#L290

The idea is to get the most common `editRange` between all completion items and set that as in itemDefaults.

JessicaJHee:
Fixes #1561
* moves `editRange` and `insertTextFormat` to `itemDefaults` when supported

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
@@ -48,6 +48,7 @@ import org.eclipse.lemminx.services.extensions.completion.ICompletionResponse;
 import org.eclipse.lemminx.services.snippets.IXMLSnippetContext;
 import org.eclipse.lemminx.settings.SharedSettings;
 import org.eclipse.lemminx.settings.XMLCompletionSettings;
+import org.eclipse.lemminx.utils.CompletionItemDefaultsUtils;
 import org.eclipse.lemminx.utils.StringUtils;
 import org.eclipse.lemminx.utils.XMLPositionUtility;
 import org.eclipse.lsp4j.CompletionItem;
@@ -288,6 +289,8 @@ public class XMLCompletions {
 			return completionResponse;
 		} finally {
 			collectSnippetSuggestions(completionRequest, completionResponse);
+			// Manage itemDefaults
+			CompletionItemDefaultsUtils.process(completionResponse, settings);
 		}
 	}
 
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/CompletionItemDefaultsUtils.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/CompletionItemDefaultsUtils.java
@@ -1,0 +1,116 @@
+/*******************************************************************************
+* Copyright (c) 2023 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.utils;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.eclipse.lemminx.services.CompletionResponse;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lsp4j.CompletionItem;
+import org.eclipse.lsp4j.CompletionItemDefaults;
+import org.eclipse.lsp4j.InsertTextFormat;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.jsonrpc.messages.Either;
+
+/**
+ * CompletionItemDefaultsUtils for implementing itemDefaults in completion list
+ */
+public class CompletionItemDefaultsUtils {
+
+	private static final String ITEM_DEFAULTS_EDIT_RANGE = "editRange";
+
+	private static final String ITEM_DEFAULTS_INSERT_TEXT_FORMAT = "insertTextFormat";
+
+	/**
+	 * Processes the completion list and manages item defaults.
+	 *
+	 * @param completionResponse the completion response
+	 * @param sharedSettings     the shared settings
+	 */
+	public static void process(CompletionResponse completionResponse, SharedSettings sharedSettings) {
+		CompletionItemDefaults itemDefaults = new CompletionItemDefaults();
+		List<CompletionItem> completionList = completionResponse.getItems();
+		if (sharedSettings.getCompletionSettings().isCompletionListItemDefaultsSupport(ITEM_DEFAULTS_EDIT_RANGE)) {
+			setToMostCommonEditRange(completionList, itemDefaults);
+			completionResponse.setItemDefaults(itemDefaults);
+		}
+		if (sharedSettings.getCompletionSettings()
+				.isCompletionListItemDefaultsSupport(ITEM_DEFAULTS_INSERT_TEXT_FORMAT)) {
+			setToMostCommonInsertTextFormat(completionList, itemDefaults);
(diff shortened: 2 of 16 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
```diff
@@ -288,7 +294,83 @@ public CompletionList doComplete(DOMDocument xmlDocument, Position position, Sha
 			return completionResponse;
 		} finally {
 			collectSnippetSuggestions(completionRequest, completionResponse);
+			// Manage itemDefaults
+			CompletionItemDefaults itemDefaults = new CompletionItemDefaults();
```
angelozerr, line 298: As code from XMLCompletion is very big, could you please create an utility class like CompletionItemDefaultsUtils with only public static method `CompletionItemDefaultsUtils.process(completionRequest, completionResponse);`

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
```diff
@@ -288,7 +294,83 @@ public CompletionList doComplete(DOMDocument xmlDocument, Position position, Sha
 			return completionResponse;
 		} finally {
 			collectSnippetSuggestions(completionRequest, completionResponse);
+			// Manage itemDefaults
+			CompletionItemDefaults itemDefaults = new CompletionItemDefaults();
+			if (settings.getCompletionSettings().isCompletionListItemDefaultsSupport("editRange")) {
```
angelozerr, line 299: Define `editRange` as constant.

datho7561:
Given the schema:
```rng
  <start>
    <ref name="articleContent" />
  </start>
  <define name="articleContent">
    <element name="article">
      <choice>
        <ref name="titleContent" />
        <ref name="titleLineContent" />
      </choice>
      <element name="salami">
        <text/>
      </element>
    </element>
  </define>
  <define name="titleContent">
    <choice>
      <element name="title">
        <empty />
      </element>
    </choice>
  </define>
  <define name="titleLineContent">
    <element name="title2">
      <choice>
        <ref name="lineElt"></ref>
        <ref name="thirdNest"></ref>
      </choice>
    </element>
  </define>
  <define name="thirdNest">
    <choice>
      <element name="nest1">
        <empty/>
      </element>
      <element name="nest2">
        <empty/>
      </element>
      <element name="nest3">
        <empty/>
      </element>
    </choice>
  </define>
  <define name="lineElt">
    <element name="line">
      <text />
    </element>
  </define>
</grammar>
```

And the document:

```xml
<?xml-model href="schema.rng"?>
<article>
    |
</article>
```

If I open completion at the `|`, I get `title` and `title2` suggested as expected, but when I apply completion, I get the following result:

```xml
<?xml-model href="schema.rng"?>
<article>
    title
</article>
```

datho7561:
It seems the `itemDefaults` is almost always `{}`, since the snippets and the element completions have different data.

JessicaJHee:
@datho7561 Hmm it works on my end, do you have any settings set that might affect completion?

datho7561:
I'll give it another try

datho7561:
I know what it is, it's the binary. You will need to add `org.eclipse.lsp4j.CompletionItemDefaults` to the reflection configuration in order for it to work

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/CompletionItemDefaultsUtils.java
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/**
+	 * Returns the most common editRange in the completion list.
+	 *
+	 * @param completionList the completion response
+	 * @return the most common editRange in the completion list
+	 */
+	private static Range findMostCommonEditRange(List<CompletionItem> completionList) {
+		Map<Range, Integer> countMapEditRange = new HashMap<>();
```
angelozerr, line 68: Instead of having a Map with Map<Range, Integer>, I wonder ifit should be beter to have a `Map<Range, List<CompletionItem>>`, it will avoid to loop of for `(CompletionItem item : completionList) {`  to update setInsertTextFormat to null.

JessicaJHee: I'm not sure if I understand, how would it be able to count the number of occurrences of the range?

angelozerr: perhaps I'm wrong, but my first impression that you count the number of item which have the same range. In order to store the count, my suggestion was to store a list of item which have the same range. The count number is the size list, no?

JessicaJHee: Ohhh I see, will do that, thanks!

angelozerr:
@JessicaJHee do you see some improvement with lemminx maven with the dependency completion https://github.com/eclipse/lemminx-maven/pull/482# ?

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/CompletionItemDefaultsUtils.java
```diff
@@ -0,0 +1,105 @@
...
+	 * defaults and processes the completion list.
+	 *
+	 * @param completionList the completion list
+	 * @param itemDefaults   the item defaults
+	 */
+	private static void setToMostCommonEditRange(List<CompletionItem> completionList,
+			CompletionItemDefaults itemDefaults) {
+		Map<Range, List<CompletionItem>> itemsByRange = completionList.stream()
+				.collect(Collectors.groupingBy(item -> item.getTextEdit().getLeft().getRange()));
+		int maxCount = itemsByRange.values().stream().mapToInt(List::size).max().orElse(0);
```
datho7561, line 67: If you use a for-each loop here and iterate over the entry set, you can store both the max list length and the range corresponding to this list. That way, when you set the text edits to null, you can avoid iterating over the map again to find the longest list, and instead access it directly by the key, which in this case is the range.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/CompletionItemDefaultsUtils.java
```diff
@@ -0,0 +1,105 @@
...
+	 *
+	 * @param completionList the completion list
+	 * @param itemDefaults   the item defaults
+	 */
+	private static void setToMostCommonInsertTextFormat(List<CompletionItem> completionList,
+			CompletionItemDefaults itemDefaults) {
+		Map<InsertTextFormat, List<CompletionItem>> itemsByInsertTextFormat = completionList.stream()
+				.filter(item -> item.getInsertTextFormat() != null)
+				.collect(Collectors.groupingBy(item -> item.getInsertTextFormat()));
+		int maxCount = itemsByInsertTextFormat.values().stream().mapToInt(List::size).max().orElse(0);
```
datho7561, line 93: Same here

datho7561 (changes requested):

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/utils/CompletionItemDefaultsUtils.java
JessicaJHee: okay thank you! hoping I got the right idea

datho7561 (approved):
Looks good. Thanks, Jessica!

angelozerr:
It works great on vscode.

I cannot test with Eclipse IDE because it misses the capability https://github.com/eclipse/lsp4e/issues/741 -(

angelozerr:
Once github actions will be finished,I'm OK to merge it.

angelozerr:
Thanks @JessicaJHee !

PR #1599, opened 2023-11-14
OpenQuoteExpected error for ATTLIST breaks DTD validation

angelozerr:
OpenQuoteExpected error for ATTLIST breaks DTD validation

Fixes [#943](https://github.com/redhat-developer/vscode-xml/issues/943)

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/OpenQuoteExpectedCodeAction.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/OpenQuoteExpectedCodeAction.java
@@ -18,6 +18,7 @@ import org.eclipse.lemminx.commons.CodeActionFactory;
 import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMDocument;
 import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.dom.DTDDeclParameter;
 import org.eclipse.lemminx.services.extensions.codeaction.ICodeActionParticipant;
 import org.eclipse.lemminx.services.extensions.codeaction.ICodeActionRequest;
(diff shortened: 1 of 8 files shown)


angelozerr:
@ejgutierrez74  I did a quick PR to fix your issue https://github.com/redhat-developer/vscode-xml/issues/943

See the following demo

DTDQuoteDemo

As you can notice you can see DTD error when you hover the DTD SYSTEM path.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/codeactions/OpenQuoteExpectedCodeAction.java
```diff
@@ -43,10 +44,22 @@ public void doCodeAction(ICodeActionRequest request, List<CodeAction> codeAction
...
+		DOMNode attr = document.findAttrAt(offset);
+		if (attr != null && attr.isAttribute()) {
+			// ex : <foo bar=value > <-- error on value which must be quoted.
+			insertQuotationForAttr(attr, document, request, diagnostic, codeActions);
+		} else {
+			DTDDeclParameter parameter = document.findDTDDeclParameterAt(offset);
+			if (parameter != null) {
+				// ex : <!ATTLIST dadesadministratives idinstitut ID > <-- error on idinstitut
+				// which must be quoted.
+				insertQuotationForParameter(parameter, document, request, diagnostic, codeActions);
```
fbricon, line 56: in all cases of `Open quote is expected for attribute "attributename" associated with an  element type  "elementname".` errors, none of them are about missing a quote, but needing to insert `#REQUIRED` at the end of the declaration. 

So at least you should add that as another quickfix

angelozerr: Indeed code actions is bad, I disable it.

PR #1602, opened 2023-11-17
Separate instructions for testing well formed XML files and validation against DTD

angelozerr:
Separate instructions for testing well formed XML files and validation against DTD

See https://github.com/redhat-developer/vscode-xml/issues/947


angelozerr:
See PR from vscode-xml https://github.com/redhat-developer/vscode-xml/pull/950 which consumes the new command `Revalidate current XML file (XML syntax Only)`

Issue #1625, opened 2024-02-08
Allow completion participants to suggest completion items inside CDATA blocks


pablocabrera85:
In a recent project I worked at, we implemented a generic support for embedded languages inside XML files. All our embedded languages lived inside CDATA blocks. 

In the following example, we could delegate to another extension the calculation of autocomplete options for the java script block:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
  "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
 <head>
 <title>XHTML 1.0 Strict Example</title>
 <script type="application/javascript">
 <![CDATA[
 function loadpdf() {
    document.getElementById("pdf-object").src="http://www.w3.org/TR/xhtml1/xhtml1.pdf";
 }
 ]]>
 </script>
 </head>
 <body onload="loadpdf()">
 </body>
</html>
```
In the future this functionality could be extended to other parts of the API, as we could for example want to find the references of the loadpdf() function and for that we would need to be able to send the corresponding request.

angelozerr:
@pablocabrera85 I close this issue but if you need it again, please reopen it.

PR #1626, opened 2024-02-08
Add completions support inside CDATA content

pablocabrera85:
Addresses [this](https://github.com/eclipse/lemminx/issues/1625) enhancement request 

To facilitate the implementation of embedded language features inside the XML supporting the completion request inside CDATA blocks is required.

The changes where added in a backward compatible way an as such an increment in version was pushed from 0.27.1-SNAPSHOT to 0.28.0-SNAPSHOT

Mockito was added as a dependency as it is easier to mock interfaces and verify invocations rather than implement a new completion participant

### DIFF
--- a/org.eclipse.lemminx/pom.xml
+++ b/org.eclipse.lemminx/pom.xml
@@ -290,7 +290,7 @@
 		<dependency>
 			<groupId>org.mockito</groupId>
 			<artifactId>mockito-core</artifactId>
-			<version>5.10.0</version>
+			<version>4.11.0</version>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
--- a/org.eclipse.lemminx/pom.xml
+++ b/org.eclipse.lemminx/pom.xml
@@ -3,7 +3,7 @@
 	<parent>
 		<groupId>org.eclipse.lemminx</groupId>
 		<artifactId>lemminx-parent</artifactId>
-		<version>0.27.1-SNAPSHOT</version>
+		<version>0.28.0-SNAPSHOT</version>
 	</parent>
 	<artifactId>org.eclipse.lemminx</artifactId>
 	<properties>
@@ -285,6 +285,12 @@
 			<artifactId>jetty-server</artifactId>
 			<!-- Need to stick to 9.x as long as build is running with Java 1.8-->
 			<version>9.4.53.v20231009</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<version>5.10.0</version>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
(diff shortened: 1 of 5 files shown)


org.eclipse.lemminx/pom.xml
```diff
@@ -287,5 +287,11 @@
 			<version>9.4.53.v20231009</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
angelozerr, line 290: We don't use mockito for test.

@fbricon @datho7561 do you think it can be a good idea to use mockito?

angelozerr:
@pablocabrera85 have you a concrete sample with CDATA completion participant?

pablocabrera85:
Hi @angelozerr .

The following are some of the examples I'm aware of:
- Script blocks inside HTML:
```xml
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
  "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
 <head>
 <title>XHTML 1.0 Strict Example</title>
 <script type="application/javascript">
 <![CDATA[
 function loadpdf() {
    document.getElementById("pdf-object").src="http://www.w3.org/TR/xhtml1/xhtml1.pdf";
 }
 ]]>
 </script>
 </head>
 <body onload="loadpdf()">
 </body>
</html>
```
MuleSoft also has some components that allow to use scripting language like Groovy, Python and Ruby
```xml
<scripting:execute  engine="Groovy">
    <scripting:code ><![CDATA[flow = registry.lookupByName("test-flow").get();
        if (flow.isStarted())
	        flow.stop()
        else
	        flow.start()]]>
    </scripting:code>
</scripting:execute>
```
And also they have a transformation language:
```xml
<ee:transform doc:id="747f74d4-cb66-4f8e-8222-24784e6863ae"
  doc:name="Transform" doc:timestamp="1510861248434">
    <ee:message>
        <ee:set-payload>
          <![CDATA[%dw 2.0
            output application/json
            ---
            {
              // YOUR DW SCRIPT
            }
          ]]>
       </ee:set-payload>
    </ee:message>
</ee:transform>
```
I found [this](https://community.sonarsource.com/t/sql-code-in-xml/39961) article talking about SQL inside XML

angelozerr:
> The following are some of the examples I'm aware of:

Sorry I wanted to know if there were an implementation to manage completion with JavaScript inside XML. 

Your PR is great but I would like to avoid providing a participant API if it exists none implementation. 

Could you deal JavaScript completion on your side with your PR?

angelozerr:
> Yes, without this PR, we end up manually process the DOM to check if we are in a CDATA block to cover for this scenario. It would be great if we could avoid manually processing the DOM and just use the lemminx APIs

That's great,could you share your project which does that please?

pablocabrera85:
@angelozerr Sorry for the late reply. The code is in a private repository, and unfortunately I no longer have access to it, as I was affected by the latest round of layoffs. I'll close this PR.

PR #1628, opened 2024-02-15
Add About file to LemMinX OSGi bundle.

rgrunber:
As per https://www.eclipse.org/projects/handbook/#legaldoc-plugins-abouts .

The uber jar already contains one, but I suspect it is the one provided by LSP4J. At least for the non-uber jar, there should be an about file.


laeubi:
@angelozerr this looks like something that can be merged without much risk

PR #1629, opened 2024-02-19
Upgrade to latest lsp4j

laeubi:


mickaelistria:
Such a change needs further coordination as it might break extensions.

angelozerr:
Upgrading lsp4j is too a problem for vscide xml because it will require for user to upgrade their java version becayse today jzva8 is only required.

Hopefully we havr binary support but I think it will require some fix to support binary correctly vecause I suppose lsp4j provides nzw classes that we beed to declare in reflect json file.

laeubi:
@mickaelistria @angelozerr 

I tried to split up at laest the removal ox xtext now:
- https://github.com/eclipse/lemminx/pull/1630

this should make things easier to upgrade later on.

Another problem is that lsp4j changed a constructor in a non backward compatible way, if lsp4 could add a legacy one at laest it will work to easily upgrade if decided to do so.

PR #1699, opened 2024-11-07
ICompletionParticipant#onXMLContent is not called for CDATA

laeubi:
See 
- https://github.com/eclipse-lemminx/lemminx/issues/1694

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLCompletions.java
@@ -109,6 +109,7 @@ public class XMLCompletions {
 			Scanner scanner = XMLScanner.createScanner(text, node.getStart(), isInsideDTDContent(node, xmlDocument));
 			String currentTag = "";
 			TokenType token = scanner.scan();
+			TokenType lastToken = null;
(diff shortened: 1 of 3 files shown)


laeubi:
@angelozerr can you review this?

This fixes the testcase for me, without my change testcase fails, with my change test-case succeeds.

angelozerr:
The code looks good, but CI build fails https://github.com/eclipse-lemminx/lemminx/actions/runs/11721131221/job/32647917208?pr=1699#step:4:681

angelozerr:
I wonder if request should have a flag like isInCDATA, because  when apply completion is done, it must not remove the CDATA.

Perhaps we don't need it, but I think your test class should have some tests with apply completion with CDATA and without CDATA).

laeubi:
Any hint how to best test that case?

laeubi:
I looked at the failing test and I'm not sure if this needs to be adjusted, what seem to happen is that now a completion (to close the `a` tag) is suggested because before CDATA section was a "terminating" element, if I change the test to place the cursor before the cdata section then even three completions are suggested (including the close with `a` tag) so it seems that just the expectation of the test ist wrong.

laeubi:
Now it works for windows but fails for mac/linux...?!?

angelozerr:
> Now it works for windows but fails for mac/linux...?!?

We have some tests which uses Timeout and perhaps they failed, I have restarted the CI build and it is working perfectly.

PR #1701, opened 2024-11-10
Add an uber-bundle artifact

laeubi:
Currently lemminx only provides an uber-jar that has no OSGi meta data at all, because of this for example wild-webdeveloper needs to embed this jar instead of use one installed in the OSGi framework.

This now adds a new uber-bundle artifact that is at the moment very similar to uber-jar but contains some OSGi data useful for writing extensions. This can then be used as a starting point to incrementally make it more OSGified, e.g. by export/import its dependencies that are already available as OSGi bundles (gson, lsp4j, ...)

FYI @mickaelistria 

This currently requires tycho-snapshot repository therefore I marked the PR as a draft but wanted to share current progress already.

### DIFF
--- a/org.eclipse.lemminx/pom.xml
+++ b/org.eclipse.lemminx/pom.xml
@@ -14,6 +14,12 @@
 		<native.maven.plugin.version>0.9.23</native.maven.plugin.version>
 		<graalvm.static />
 	</properties>
+	<pluginRepositories>
+	<pluginRepository>
+		<id>tycho-snpashots</id>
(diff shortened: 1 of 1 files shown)


laeubi:
As an alternative I'm currently try to make the lemminx bundle itself a bit more useful, has anyone actually using that inside OSGi environments and can tell about it more?

laeubi:
There is now
- https://github.com/eclipse-lemminx/lemminx/pull/1703

as a first starting point.

laeubi:
I'm closing this in favor of much simpler
- https://github.com/eclipse-lemminx/lemminx/pull/1710

PR #1702, opened 2024-11-10
Make CacheResourcesManager Constructor used for test package protected

laeubi:
It looks like the CacheResourcesManager(Cache cache) constructor is oly used for test, so this is made package protected now.

This has the advantage that there is no public reference to google Cache API making it an internal implementation detail that can easily be exchanged (or embedded).


angelozerr:
Thanks @laeubi !

PR #1705, opened 2024-11-10
Stabilize two test that seem to often fail on the CI

laeubi:
- Use a timeout instead of static wait times
- retry diagnostics
- use a copy of the current state

This also reduces the test time from 5 sec to less than a second for the 

> TimeUnit.SECONDS.sleep(5); // HACK: to make the timing work on slow machines


org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/XMLAssert.java
```diff
@@ -860,6 +860,24 @@ public static void testPublishDiagnosticsFor(String xml, String fileURI, XMLLang
...
+			XMLValidationRootSettings validationSettings, XMLLanguageService xmlLanguageService,
+			PublishDiagnosticsParams... expected) {
+		long deadline = System.currentTimeMillis() + timeout;
+		while (true) {
+			try {
+				testPublishDiagnosticsFor(xml, fileURI, validationSettings, xmlLanguageService, expected);
+				return;
+			} catch (AssertionError e) {
+				if (System.currentTimeMillis() < deadline) {
+					Thread.onSpinWait();
```
angelozerr, line 873: Not available in java8

laeubi:
Seems one need to sort out what is really "unexpected" here... some of the returned diagnostics do not look that wrong given other seem to be intermediates or even valid... I'll take a look later on this, but currently it makes contributions really hard if there are random failures.

angelozerr:
> Seems one need to sort out what is really "unexpected" here... some of the returned diagnostics do not look that wrong given other seem to be intermediates or even valid... I'll take a look later on this, but currently it makes contributions really hard if there are random failures.

I understand that it is an annoying problem, this problem should be fixed.

PR #1721, opened 2025-01-08
Update to Java 11

datho7561:
- Bump to 0.30.0-SNPASHOT, since updating to Java 11 breaks API
- Use `--release 11` in maven compiler plugin
- Use Java 11 in the Jenkinsfiles
- Update tycho to 2.7.5 (the newest version that supports Java 11)
- Update to the latest lsp4j, adapt to the API changes, and remove the guava exclusion (lsp4j no longer ships with guava, so we don't need to exclude it)
- Delete `ToStringBuilder` and use the one from lsp4j
- Update to the newest jetty version that supports Java 11


rgrunber:
Nice. And this will set us up to finally address https://github.com/eclipse-lemminx/lemminx/issues/1291 (Fred has a draft PR for it) .

datho7561:
I added the necessary reflection config to ensure the server launches. I poked around a bit with a few features and it seems to be working, with no exceptions in the logs.

angelozerr:
LGTM.

I suggest that we merge the PR ASAP in order to see some potential problem when we will develop xml features.

angelozerr (approved):

rgrunber:
Hold on. I see a few references to `JavaSE-8` / `1.8` related stuff. Would be good to verify whether these should remain or change :

https://github.com/eclipse-lemminx/lemminx/blob/e5071835bb39d6fdb131f6a63ae1ec34bdb7c1b2/pom.xml#L121-L122

datho7561:
@rgrunber I deleted that file and used the one from lsp4j; I'll double check the binary still works first though

datho7561:
Another line about 1.8:
https://github.com/eclipse-lemminx/lemminx/blob/main/org.eclipse.lemminx/pom.xml#L294 

I'll update jetty as well I guess

PR #1723, opened 2025-01-09
Update binary documentation in README

### LINKED ISSUE #1722: Update documentation in README for building a binary of lemminx
- The instructions indicate an older version of GraalVM than we use in our build steps
- ~~We now have a maven profile that handles setting everything up correctly for the build~~ apparently the `-Dnative` is still better to use (see [this comment from Roland](https://github.com/eclipse-lemminx/lemminx/pull/1723#discussion_r1909053973))

datho7561:
- Do not specify the version of GraalVM to use, since it's likely that the latest GraalVM works, and the lifecycle of GraalVM versions means any version we list will likely be out of date shortly.
- Remove instructions to use `gu install native-image`, because in newer versions of GraalVM, `gu` is no longer included and `native-image` is included by default

Fixes #1722

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,8 @@ socket.connect(socketPort)
 Generating a native binary:
 ---------------------------------
 To generate a native binary:
-- [Install GraalVM 20.2.0](https://www.graalvm.org/docs/getting-started/#install-graalvm)
-- In a terminal, run `gu install native-image`
+- [Install GraalVM](https://www.graalvm.org/docs/getting-started/#install-graalvm),
+  ideally with a tool that allows setting up multiple versions of Java, such as SDKMAN
 - Execute a Maven build that sets the flag `native`: `./mvnw clean package -Dnative -DskipTests`
   - On Linux, compile with `./mvnw clean package -Dnative -DskipTests -Dgraalvm.static=--static`
     in order to support distributions that don't use `glibc`, such as Alpine Linux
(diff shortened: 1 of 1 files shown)


datho7561:
@rgrunber if you have a chance, could you please look over these changes? You have a good understanding of the GraalVM build process

README.md
```diff
@@ -68,10 +68,10 @@ socket.connect(socketPort)
...
 ---------------------------------
 To generate a native binary:
-- [Install GraalVM 20.2.0](https://www.graalvm.org/docs/getting-started/#install-graalvm)
-- In a terminal, run `gu install native-image`
-- Execute a Maven build that sets the flag `native`: `./mvnw clean package -Dnative -DskipTests`
-  - On Linux, compile with `./mvnw clean package -Dnative -DskipTests -Dgraalvm.static=--static`
+- [Install GraalVM](https://www.graalvm.org/docs/getting-started/#install-graalvm),
+  ideally with a tool that allows setting up multiple versions of Java, such as SDKMAN
+- Execute a Maven build that sets the profile `native`: `./mvnw clean package -Pnative -DskipTests`
+  - On Linux, compile with `./mvnw clean package -Pnative -DskipTests -Dgraalvm.static=--static`
```
rgrunber, line 74: Can we continue to activate it with `-D` as opposed to `-P`. I have always been nervous about activating profiles explicitly like this because IIRC they override (disable) any profiles that would be activated by default due to some condition. As a result I've always made it possible to activate profile by properties.

Also, just something to be aware, just giving `--static` to that property is technically the `alpine` configuration : https://github.com/redhat-developer/vscode-xml/blob/f66692877045eb72460987523d6b03686a4af9fb/.github/workflows/native-image.yaml#L35-L40 . The Linux one has an extra flag to avoid statically linking libc, but I think either should work, and given that the current is a smaller command to type (and this is just for trying it out), maybe we leave it as-is.

rgrunber (commented):
We use the `setup-graalvm` workflow, so I dug into whether it's truly safe (goes without saying downloading a recent graalvm jdk does include `native-image`).

https://github.com/graalvm/setup-graalvm/blob/4a200f28cd70d1940b5e33bd00830b7dc71a7e2b/src/gu.ts#L45

Seems pretty safe based on that (and the PR referenced).

PR #1732, opened 2025-02-20
Remove Service-Component OSGI-INF/*.xml from uber-jar

merks:
https://github.com/eclipse-lemminx/lemminx/issues/1731

### DIFF
--- a/org.eclipse.lemminx/pom.xml
+++ b/org.eclipse.lemminx/pom.xml
@@ -119,7 +119,6 @@
 							<Bundle-Version>${parsedVersion.osgiVersion}</Bundle-Version>
 							<Bundle-Vendor>Eclipse.org</Bundle-Vendor>
 							<Bundle-Name>lemminx uber jar bundle</Bundle-Name>
-							<Service-Component>OSGI-INF/*.xml</Service-Component>
 						</manifestEntries>
 					</archive>
 				</configuration>


merks:
I ran the build locally and there is no such folder:
```
merks@CORE MINGW64 /d/Users/merks/temp/lemminx/org.eclipse.lemminx
$find . -name "*OSGI*"

merks@CORE MINGW64 /d/Users/merks/temp/lemminx/org.eclipse.lemminx
```

mickaelistria:
@laeubi Can you please review?

laeubi:
> Is it an error not to have it? 

Of course not

> What will no longer work without it? 

You can no longer extend the bundle with fragments that contain declarative service components.

> What is looking up what, where, and why? 

https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#d0e30931

> Is anyone actually using that capability or is this something that could exist but doesn't current actually exist?

As it was only added with the last release it could be used now but not before. e.g. m2e is providing an extension to lemminx now, for further details see:
- https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/discussions/1712

> Pretend that I am stupid and know nothing about this. Also imagine a million users coming to this issue who will be equally confused and that being told to please read the spec and/or please go complain somewhere about ambiguous logging specification will perhaps prove unsatisfying. They might shake their heads and ask "What's wrong with these people?" hypothetically speaking...

This is really hypothetical as no one is actually looking at the Eclipse log (or no one really complains) as these are not errors that surface to the usual user, e.g. this is what I see in my daily eclipse (and I regularly clear the error log so it is only from the past days):
grafik

So did we get "million complains"? not really ... are people confused? not really... Does anyone work on these issue? I'm not sure ...

> Can we can do something/anything to that make something be located by the wildcard so that there is no error logged?

Eclipse consumes https://github.com/apache/felix-dev/tree/master/scr what is the component that logs the error.

laeubi:
I have created
- https://github.com/apache/felix-dev/pull/383



> or better yet, can the project itself provide a degenerate/no-op/empty XML file that will make it stop logging?

Having something added as a workaround to suppress a harmless logmessage do not seems valuable to me. For the record an dummy, disabled by default component description will look like this:

```
<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" enabled="false" name="dummy">
   <!-- See https://github.com/apache/felix-dev/pull/383 -->
   <implementation class="java.lang.Object"/>
</scr:component>
```

fbricon:
@merks is it still an issue now that @laeubi provided a fix inFelix? Can we close this PR?

PR #1733, opened 2025-02-27
Use GraalVM CPU architecture compatibility mode

datho7561:
This prevents GraalVM from using CPU architecture extensions (for example, AVX and AVX2 for x86_64). This will increase the portability of the binary (i.e. it can be run on older CPUs that don't support those instructions).

Docs for the setting: https://www.graalvm.org/latest/reference-manual/native-image/overview/Options/#build-options

See https://github.com/redhat-developer/vscode-xml/issues/1055

### DIFF
--- a/org.eclipse.lemminx/pom.xml
+++ b/org.eclipse.lemminx/pom.xml
@@ -176,6 +176,7 @@
 							<imageName>lemminx-${os.detected.classifier}-${project.version}</imageName>
 							<buildArgs>
 								--no-fallback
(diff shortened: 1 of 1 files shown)


datho7561:
I've tested the generated binary using Helix and it works there. I also disassembled the binary and used the awk script here to look for AVX instructions: https://stackoverflow.com/a/50056059 ( I didn't find any ).

datho7561:
@rgrunber if you have some time, do you mind looking over this?

PR #1738, opened 2025-04-23
Show hex colours when `#` is missing

datho7561:
Fixes redhat-developer/vscode-xml#1073

- [x] Fix test failures
- [x] Needs new tests

### DIFF
--- a/org.eclipse.lemminx/pom.xml
+++ b/org.eclipse.lemminx/pom.xml
@@ -3,7 +3,7 @@
 	<parent>
 		<groupId>org.eclipse.lemminx</groupId>
 		<artifactId>lemminx-parent</artifactId>
-		<version>0.30.2-SNAPSHOT</version>
+		<version>0.31.0-SNAPSHOT</version>
 	</parent>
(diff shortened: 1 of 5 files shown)


datho7561:
@Lootwig Do you mind trying to sign the Eclipse Contributor Agreement again? It would be nice to merge your PR instead of mine, since you put a lot of work into it. Sorry about this whole ordeal; the Eclipse Infrastructure outage has been really frustrating for everyone involved.

If you really don't think it's worth it, I cleaned up this PR and added the necessary changes to colour presentation the logic. If you have time, do you mind looking over it?

datho7561:
We're going to do a z release soon, so I'm going to merge this. I'm pretty confident it works, and in the case that I messed up, it's pretty low stakes, since colour presentation is a feature that many editors don't support.

Lootwig:
Sorry for not having gotten back with a reply! (I did check the login though when you asked, but no luck & in case PW reset requires emailing admins or similar at eclipse, I don't think I would've participated at that level of hoop jumping either way 😬)

datho7561:
That's fair, that sounds kind of awful to deal with

PR #1753, opened 2025-09-09
Use `ProcessHandle` instead of shell script to check for parent process

datho7561:
See #1025


datho7561:
I made sure it still runs as normal, but I haven't tried killing the parent process manually yet, I should do that...

datho7561:
I made sure that lemminx starts up properly, then closed all VS Code abnormally (`killall -9 code`), and ensured that eventually the lemminx process stopped. I did this for the java version and the binary on Linux.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ParentProcessWatcher.java
```diff
@@ -46,9 +40,9 @@ public final class ParentProcessWatcher implements Runnable, Function<MessageCon
 	private static final long INACTIVITY_DELAY_SECS = 30 * 1000;
 	private static final int POLL_DELAY_SECS = 10;
 	private volatile long lastActivityTime;
-	private final ProcessLanguageServer server;
 	private ScheduledFuture<?> task;
 	private ScheduledExecutorService service;
+	private ProcessLanguageServer server;
```
fbricon, line 45: I think it should stay final

PR #1755, opened 2025-09-30
Do not format text content of `<style>` and `<script>`

datho7561:
`<style>` is usually used for CSS, and `<script>` is usually used for JavaScript (or some other scripting language). Given the user has probably manually formatted this code, we should avoid formatting the text content by default.


angelozerr (approved):

angelozerr:
Thanks so much to have written tests! Thats super great.

PR #1769, opened 2026-02-20
POC Perf

angelozerr:


angelozerr:
It was just an experimentation, that I have used to create some PR

PR #1771, opened 2026-02-22
Performance with Immutable charsequence

angelozerr:
With large file like https://github.com/eclipse-lemminx/lemminx/blob/main/org.eclipse.lemminx/src/test/resources/xml/content.xml when you type just one space to in the editor, the memory grows up a lot.

image

This grows up comes from with this line:
https://github.com/eclipse-lemminx/lemminx/blob/5ec17679f439fd55b06f3435c72693b2c7654b03/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/TextDocument.java#L172

when  text must be updated.  The StringBuilder copy the text and memory grows up.

This PR fixes this issue by using CharSequence (immutable). It requires to update a of code, but it gains a lot of memory:

image


fbricon:
Since you're breaking the API, you'll need to bump lemminx version to 2.0. Check if lemminx-maven is impacted. You can also browse github to look for anyone else depending on lemminx.

angelozerr:
It was a bad idea, this PR is replaced with https://github.com/eclipse-lemminx/lemminx/pull/1802

PR #1773, opened 2026-03-11
Replace Thread.sleep() with proper synchronization in tests

fbricon:
Tests were using Thread.sleep() (up to 5 seconds each) to wait for
async operations like resource downloads and validation triggers,
making the test suite unnecessarily slow (~71s test time).

Key changes:
- BaseFileTempTest: replace sleep(1050) with Files.setLastModifiedTime()
  to ensure filesystem change detection without waiting
- XMLLanguageService: return download futures from publishDiagnostics()
  so tests can await them directly
- XMLAssert: add awaitDownloads() helper, remove retrigger callback
- MockXMLLanguageClient: add waitForDiagnosticCount() polling helper
- Replace all 5-second HACK sleeps with awaitDownloads()
- Replace async validation sleeps with waitForDiagnosticCount()
- DocumentLifecycleParticipantTest: poll for lifecycle callback completion
- CacheResourcesManagerTest: reduce cache TTL from 1s to 100ms

Build time reduced from ~1m17s to ~15s (5x faster).

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLLanguageService.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLLanguageService.java
@@ -12,7 +12,6 @@
  */
 package org.eclipse.lemminx.services;
 
-import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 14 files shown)


angelozerr:
@fbricon this PR looks very good!

angelozerr:
Let's merge this PR since it seems fix random test failures.

PR #1774, opened 2026-04-18
XML Minify support

angelozerr:
See https://github.com/redhat-developer/vscode-xml/issues/609

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/minify/XMLMinifierDocument.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/minify/XMLMinifierDocument.java
@@ -1,0 +1,426 @@
+/*******************************************************************************
+* Copyright (c) 2026 Red Hat Inc. and others.
+* All rights reserved. This program and the accompanying materials
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lemminx.extensions.minify;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.eclipse.lemminx.commons.BadLocationException;
+import org.eclipse.lemminx.commons.TextDocument;
+import org.eclipse.lemminx.dom.DOMAttr;
+import org.eclipse.lemminx.dom.DOMComment;
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.dom.DOMElement;
+import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.dom.DOMProcessingInstruction;
+import org.eclipse.lemminx.dom.DOMText;
+import org.eclipse.lemminx.settings.SharedSettings;
+import org.eclipse.lemminx.utils.StringUtils;
+import org.eclipse.lsp4j.Range;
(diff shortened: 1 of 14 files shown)


datho7561:
One thing to consider is adding this minifier as a "Source Action" CodeAction, so that it's available in the quickfix menu. This will make the feature usable in IDEs that don't implement the client-side logic to set up the command. If you don't think it's worth the effort though, I understand.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/minify/XMLMinifierDocument.java
```diff
@@ -0,0 +1,415 @@
...
+	private final TextDocument textDocument;
+	private final SharedSettings sharedSettings;
+
+	private int startOffset = -1;
+	private int endOffset = -1;
+
+	/**
+	 * XML minifier document.
+	 */
+	public XMLMinifierDocument(DOMDocument xmlDocument, Range range, SharedSettings sharedSettings) {
```
datho7561, line 57: Is the range used in practice? I don't see an option in vscode-xml to minify by range.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/minify/XMLMinifierDocument.java
```diff
@@ -0,0 +1,415 @@
...
+
+		switch (nodeType) {
+		case Node.ELEMENT_NODE:
+			minifyElement(edits, (DOMElement) node, start, end);
+			break;
+		case Node.TEXT_NODE:
+			minifyText(edits, (DOMText) node);
+			break;
+		case Node.COMMENT_NODE:
+			// Keep comments as-is for now (could add option to remove them)
```
datho7561: It might make sense to remove the comments by default. They don't "do" anything, and removing them will reduce the size of the document.

datho7561 (changes requested):

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/minify/XMLMinifierDocument.java
angelozerr: To be honnest with you, I used our XML formatter code and adapt it to support minify.

I do that because perhaps in the future we could let the user to minify a part of XML content.

But if you prefer I can remove this range support.

datho7561: That makes sense to me, I think we can keep the range as an argument.

org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/minify/XMLMinifierDocument.java
angelozerr: Ok let me try to do that

angelozerr:
> One thing to consider is adding this minifier as a "Source Action" CodeAction, so that it's available in the quickfix menu. This will make the feature usable in IDEs that don't implement the client-side logic to set up the command. If you don't think it's worth the effort though, I understand.

Good point! I will try to do that but I dont want to execute minify everytime. I will provide a codeaction which will be executed on resolve code action. It means that client must support resolve code action otherwise minify will not available. Makes sense?

datho7561:
> I will provide a codeaction which will be executed on resolve code action. It means that client must support resolve code action otherwise minify will not available. Makes sense?

Makes sense to me, it seems like it will be very expensive to compute on big files, so it makes sense to only do it if resolve is available.

datho7561 (approved):
Looks good. Thanks, Angelo!

PR #1782, opened 2026-06-04
Handle referencing schemas through UNC paths

datho7561:
- When downloading external resources is disabled, don't attempt to download non-local schema files referenced using UNCs
  - eg. `file://example.org/schemas/my-schema.xsd`, `//example.org/schemas/my-schema.xsd`, `file:////example.org/schemas/my-schema.xsd`
- Use the schema cache for resources referenced through UNCs
  - Unless the authority is `localhost` eg. `//localhost/C$/Users/Mira/my-schema.xsd`
- Fix a minor bug where the error range is wrong for schemas referenced through a UNC starting with `//` in `<?xml-model href="...`


angelozerr:
LGTM thanks @datho7561 !

PR #1788, opened 2026-07-02
Fix hover documentation for xsi:type-derived child elements across namespaces

### LINKED ISSUE #1787: Hover documentation not resolved for child elements of xsi:type-derived complex types across namespaces
## Description

When an XML document uses `xsi:type` to substitute a complex type from a different namespace, hovering over child elements inside that substitution does not show their `xs:annotation/xs:documentation`.

## Root cause (observed in 0.31.x)

The hover resolution in `ContentModelHoverParticipant` / `CMXSDDocument` appears to walk up from the hovered element only while the parent shares the same namespace. Once it reaches the parent element declared in the base namespace, the path is cut off, so the `xsi:type` attribute on that parent is never evaluated. As a result LemMinX cannot determine that the child belongs to the derived type.

Additionally, when global `xs:element` declarations are used as a workaround, `CMXSDDocument.findElementDeclaration` matches elements by `localName` only and ignores the namespace. If several schemas declare a global element with the same local name (e.g. `Title`), the documentation returned comes from whichever namespace happens to be first internally, not from the namespace actually used in the XML.

## Minimal reproduction

A small reproducible schema set and instance are available here:

https://gist.github.com/hertznsk/603e2ac8d481e59c268bef0be8e0e359

### Files

- `main.xsd` — declares `m:Root` with `m:Data` of base type `base:Data`.
- `base.xsd` — empty base type `base:Data`.
- `a.xsd` — derives `a:ADerived` from `base:Data` and adds `a:Title`.
- `b.xsd` — derives `b:BDerived` from `base:Data` and adds `b:Title`.
- `test.xml` — two `m:Data` elements with `xsi:type="a:ADerived"` and `xsi:type="b:BDerived"`.

### Steps

1. Open `test.xml` in VS Code with the vscode-xml extension (or any LemMinX-based client).
2. Hover over `<a:Title>` and `<b:Title>`.

## Expected behavior

Hovering over `<a:Title>` should show the documentation from `a.xsd`.
Hovering over `<b:Title>` should show the documentation from `b.xsd`.

## Actual behavior

No documentation hover is shown for either `<a:Title>` or `<b:Title>`.

## Workaround and its limitation

Adding global `xs:element` declarations for the child fields in each derived namespace restores hover, but for elements that share the same `localName` across namespaces (e.g. `Title`), the documentation comes from only one namespace because the lookup is localName-only.

## Environment

- LemMinX version: 0.31.x (bundled with redhat.vscode-xml-0.29.2026061109-linux-x64)
- VS Code version: any recent build
- OS: Linux x64

## Additional context

We first hit this while authoring XSD schemas for workflow XML where many step-specific namespaces extend a common `api:Data` type via `xsi:type`. A purely schema-side fix that preserves correct per-step documentation does not seem possible without altering either the XML instances or LemMinX itself.

hertznsk:
Fixes #1787

Hover documentation now resolves for child elements inside "xsi:type"-derived complex types when the type is defined in a different namespace.

Changes:
- Add null guard in XMLCatalogURIResolverExtension.hasDTDorXMLSchema
- Wire ContentModelManager into CMXSDContentModelProvider
- Resolve xsi:type-derived child elements across namespaces in hover

Co-authored-by: Sisyphus <clio-agent@sisyphuslabs.ai>

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/catalog/XMLCatalogURIResolverExtension.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/catalog/XMLCatalogURIResolverExtension.java
@@ -83,6 +83,9 @@ public class XMLCatalogURIResolverExtension implements URIResolverExtension {
 	}
 
 	private boolean hasDTDorXMLSchema(String uri) {
+		if (extensionsRegistry.getDocumentProvider() == null) {
+			return false;
+		}
 		DOMDocument document = extensionsRegistry.getDocumentProvider().getDocument(uri);
 		if (document == null) {
 			return false;
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelHoverParticipant.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/extensions/contentmodel/participants/ContentModelHoverParticipant.java
@@ -16,7 +16,9 @@ import static org.eclipse.lemminx.utils.MarkupContentFactory.createHover;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 
 import org.eclipse.lemminx.dom.DOMAttr;
 import org.eclipse.lemminx.dom.DOMElement;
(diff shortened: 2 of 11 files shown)


hertznsk:
I waive any rights to this code. Please just accept it.

angelozerr:
@hertznsk please sign ECA

angelozerr:
@hertznsk could you please sign your ECA

angelozerr:
Replaced with https://github.com/eclipse-lemminx/lemminx/pull/1850

@hertznsk I did an another PR which fixes your issue because you need also to take care of hover, type defintion and compltetion. Here a demo:

XSDMultiNs

Anyway thanks for your contribution

PR #1798, opened 2026-09-03
Fix content duplication when range formatting large XML files

### LINKED ISSUE #1791: Bug when copying XML parts on large files
There appears to be a bug in https://github.com/redhat-developer/vscode-xml/releases#release-0.29.3 that was introduced with 0.29.0 -> 0.29.1 in that when pasting a XML block in a large XML file, the content of the whole file is duplicated on every paste. For small files, this bug does not appear.

 I believe the bug is related to the PR #1768

To reproduce:

Paste the following block in a new XML file and attempt to copy one single item block within list.items.
Works fine while the file is small.

Now duplicate the list list.items within list until you have a couple thousand lines, usually until around 4000 lines everything works just fine. After that point the bug occurs in that it keeps repeating the full file content on every paste and lines double on each paste (4000->8000->16000->32000). It does not matter how much you copy & paste, the full file is duplicated at the paste location.

```
<list>
	<items>
		<item id="response_status" _delta="define">
			<rank>30</rank>
		</item>
		<item id="location_id" _delta="define">
			<rank>30</rank>
		</item>
		<item id="status" _delta="redefine">
			<rank>50</rank>
		</item>
		<item id="service_id" _delta="define">
			<rank>60</rank>
		</item>
		<item id="caller_id" _delta="define">
			<rank>70</rank>
		</item>
		<item id="team_id" _delta="define">
			<rank>80</rank>
		</item>
		<item id="agent_id" _delta="redefine">
			<rank>90</rank>
		</item>
		<item id="manager_id" _delta="define">
			<rank>100</rank>
		</item>
	</items>
</list>
```

angelozerr:
When merging TextEdits for large files (>100KB), applyEdits() returns the full document text but the TextEdit range incorrectly used the original formatting range instead of the full document range. This caused range formatting (e.g. format-on-paste) to replace a small range with the entire document content, duplicating the file.

Fixes #1791

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/services/XMLFormatter.java
@@ -66,10 +66,9 @@ class XMLFormatter {
 			
 			// For large files, merge all TextEdits into a single one to avoid OutOfMemory
 			// This is more memory efficient as we don't keep thousands of TextEdit objects
-			if (shouldMergeEdits(result, xmlDocument)) {
+			if (range == null && shouldMergeEdits(result, xmlDocument)) {
 				String formatted = applyEdits(xmlDocument.getTextDocument(), result);
-				Range editRange = range != null ? range : getFullDocumentRange(xmlDocument);
-				return List.of(new TextEdit(editRange, formatted));
+				return List.of(new TextEdit(getFullDocumentRange(xmlDocument), formatted));
 			}
 			return result;
 		} catch (BadLocationException e) {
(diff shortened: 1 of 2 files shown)


angelozerr:
@datho7561 could you please review this PR. Thanks!

datho7561 (approved):
This fixes the problem for me. Thanks, Angelo!

datho7561:
There's a new test failure related to code actions. I can reproduce it reliably. However, it isn't caused by this PR, I can also reproduce it reliably on the main branch.

datho7561:
Opened https://github.com/eclipse-lemminx/lemminx/issues/1799; since it's unrelated to this change, I'll merge this PR.

PR #1801, opened 2026-09-04
Implement Roslyn-style red-green tree architecture for DOM parser 

angelozerr:


fbricon:
How about benchmarks?

angelozerr:
@fbricon I will do that, but please try it, the result is impressive (according to my tests with [content.xml](https://github.com/eclipse-lemminx/lemminx/blob/main/org.eclipse.lemminx/src/test/resources/xml/content.xml) file from test which is a large file)

fbricon:
I tested the PR against https://samplelib.com/xml/sample-30mb.xml in vscode, it crashed the server with an OOM, I doubled the Xmx to 512MB, I was removing a bunch of closing tags to test diagnostics, it worked fine when working at the end of the document, but crashed the server again when generating diagnostics in the middle of the file (~L 70000). When it reopened (no Xmx change), I went to modify the file in the middle again, it straight up crashed.

angelozerr:
Thanks for your feedback @fbricon !

My current PR should improve a lot the memory since it manages incremental parser.

Now the memory problem comes from the full string text which is stored in memory. When didChange occurs it update the string content with stringbuilder which is fast, but when toString from this string builder is called memory is growing up a lot 

I am investigating the issue 

Have you tried to open yiur large xml file with released lemminx? It should be super slow when you open completion and oom should occurs faster, right?

datho7561:
It seems there's several merge conflicts

angelozerr:
@datho7561 yes it is normal, I have created several little PR by using tis PR. The last is https://github.com/eclipse-lemminx/lemminx/pull/1805 When it will be merged I will rebase this PR from main

datho7561:
Makes sense, thanks for explaining!

PR #1802, opened 2026-09-07
Add CharSequence API to avoid String allocation on document updates

angelozerr:
### Problem

On every keystroke (`textDocument/didChange`), `TextDocument.update()` was:
1. Creating a **new `StringBuilder`** from the full document text (~60 MB for large files)

https://github.com/eclipse-lemminx/lemminx/blob/110f99525a1f5b78a4bfc798dc1aa94fde294fed/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/TextDocument.java#L172

2. Applying the 1-character change
3. Calling `toString()` to materialize back to `String` (~60 MB)

https://github.com/eclipse-lemminx/lemminx/blob/110f99525a1f5b78a4bfc798dc1aa94fde294fed/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/TextDocument.java#L195

4. Passing that `String` to the parser

For a 60 MB XML document, this meant **~120 MB of throwaway allocation per keystroke** just for text handling.

### Solution

- **`TextDocument`** stores a `CharSequence textSequence` field (String in non-incremental mode, StringBuilder in incremental mode)
- **`getTextSequence()`** returns the CharSequence directly — zero-copy for StringBuilder
- **`getText()`** is `@Deprecated` — kept for external extension compatibility, with lazy String materialization via a `textDirty` flag
- **`DOMParser`** / **`XMLScanner`** / **`MultiLineStream`** accept `CharSequence` instead of `String`
- **`CharSequenceReader`** — new Reader that reads char-by-char from CharSequence without `toString()`
- All internal LemMinX code migrated from `getText()` → `getTextSequence()` (50 files)

### Lazy StringBuilder to prevent OOM

The `String` → `StringBuilder` conversion is done **lazily in `update()`** (first keystroke), not in `setIncremental()`. This is intentional: the ~60 MB humongous allocation triggers G1GC at the right moment — after `cancelModel()` has released the old DOM tree. Without this, the old DOM (~300 MB) and new DOM (~300 MB) would coexist in memory, causing a transient **~660 MB peak that could trigger OOM** on constrained heaps.

### JFR Profiling Results

Tested with [sample-30mb.xml](https://samplelib.com/xml/sample-30mb.xml) (~60 MB unzipped), single space keystroke.

| Metric | `main` | `CharSequence + lazy` |
|--------|--------|-----------------------|
| GC events | 18 | 5 |
| GC pause time | 115 ms | 50 ms |
| Steady-state heap | 473 MB | 372 MB |
| StringBuilder in update | ~120 MB **every** keystroke | 148 MB **once** (1st keystroke only) |
| OOM risk on large files | No (GC triggered by accident) | **No (GC triggered by design)** |

### Test plan

- [x] Existing test suite passes (DOM parsing, completions, diagnostics, formatting)
- [x] New `TextDocumentTest` validates CharSequence behavior for incremental updates
- [x] JFR profiling confirms allocation reduction on large XML files
- [x] No transient memory peak on first keystroke (verified with heap monitoring)

### DIFF
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ILineTracker.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ILineTracker.java
@@ -142,7 +142,7 @@ public interface ILineTracker {
 	 *
 	 * @param text the new tracked text
 	 */
-	void set(String text);
+	void set(CharSequence text);
 
 	Position getPositionAt(int position) throws BadLocationException;
 
--- a/org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGDocument.java
+++ b/org.eclipse.lemminx/src/main/java/com/thaiopensource/relaxng/pattern/CMRelaxNGDocument.java
@@ -317,6 +317,6 @@ public class CMRelaxNGDocument implements CMDocument {
 	private static String getTextContent(DOMElement element) {
 		int start = element.getStartTagCloseOffset() + 1;
 		int end = element.getEndTagOpenOffset();
-		return element.getOwnerDocument().getText().substring(start, end);
+		return element.getOwnerDocument().getText(start, end);
 	}
 }
--- a/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/CharSequenceReader.java
+++ b/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/CharSequenceReader.java
@@ -1,0 +1,95 @@
+/**
+ *  Copyright (c) 2026 Angelo ZERR.
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v2.0
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *  Angelo Zerr <angelo.zerr@gmail.com> - initial API and implementation
+ */
+package org.eclipse.lemminx.commons;
+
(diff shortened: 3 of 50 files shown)


org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/commons/ILineTracker.java
```diff
@@ -142,7 +142,7 @@ public interface ILineTracker {
 	 *
 	 * @param text the new tracked text
 	 */
-	void set(String text);
+	void set(CharSequence text);
```
fbricon, line 145: you're breaking API compatibility.

See 
```
* In order to provide backward compatibility for clients of <code>ILineTracker</code>, extension
 * interfaces are used to provide a means of evolution. The following extension interfaces
 * exist:
 * <ul>
 * <li> {@link org.eclipse.jface.text.ILineTrackerExtension} since version 3.1 introducing the concept
 *      of rewrite sessions.</li>
 * </ul>
 * <p>
 * Clients may implement this interface or use the standard implementation
 * </p>
```

angelozerr: This class is never used by other extensions.

angelozerr: More String extends CharSequence so it will continue to work.

fbricon: **Clients may implement this interface** that would break them.

angelozerr: This Javadoc comes from original JFace code base, I will change that once this PR will be merged.

PR #1855, opened 2026-09-15
Add textDocument/onTypeFormatting support

angelozerr:
Implement LSP onTypeFormatting with '\n' as trigger character to properly
indent new lines when pressing Enter inside XML documents:

- Inside start tags: indent according to splitAttributes settings (alignWithFirstAttr, splitNewLine, preserve)
- Between start/end tags: indent content one level deeper, and when the closing tag is right at the cursor, split it to a new line

Respects all formatter settings: splitAttributes,
splitAttributesIndentSize,
tabSize, insertSpaces, and formatting enabled/disabled toggle.

Infrastructure: register capability in ServerCapabilitiesInitializer, dynamic registration in XMLCapabilityManager, handler in XMLTextDocumentService, logic in XMLFormatter.

To  test this PR in vscode, you will need to define:

`"editor.formatOnType": true`

Here a demo:

 *  **with on type formatting**
OnTypeFormatting

 * **without on type formatting**:

NoOnTypeFormatting


datho7561:
For this case, the result is weird (put the cursor where `|` is then hit enter):

```xml
<aaa>
  <bbb>
    <ccc
        asdf="hjkl">|</ccc>
  </bbb>
</aaa>
```

it results in this (note that the cursor position is wrong):



```xml
<aaa>
  <bbb>
    <ccc
        asdf="hjkl">
      
    |</ccc>
  </bbb>
</aaa>
```

It might be a limitation of the protocol. I know in yaml-language-server there are some bugs about cursor placement after applying format on type.

